PR #223, opened 2018-11-06
Issue 217 - Fix Edge / IE console errors

### LINKED ISSUE #217: Error loading dependencies on Edge and IE，except chrome

Marc-Andre-Rivet:
Fixes #217.

Issue in IE 11 is related to 4 problematic usages:
- `Object.assign` // polyfilled
- `Array.find` // polyfilled
- `new URL` // shimed
- document.queryAllSelector(selector) -> NodeList has no `forEach` method // coerce into Array with `Array.from`

Using the opportunity to upgrade from Babel 6.x to Babel 7.x as it provides a better (much easier) integration of preset-env, core-js and babel-polyfill.


babel.config.js
```diff
@@ -0,0 +1,16 @@
+const presets = [
+    ["@babel/env", {
+        targets: {
+            browsers: [
+                "last 2 Chrome versions",
+                "last 2 Firefox versions",
+                "last 2 Safari versions",
+                "last 2 Edge versions",
+                "Explorer 11"
```
Marc-Andre-Rivet: Adding Explorer 11!

babel.config.js
```diff
@@ -0,0 +1,16 @@
...
+            ]
+        },
+        useBuiltIns: "usage",
```
Marc-Andre-Rivet: This is Babel 7.x, this means babel polyfills will be included on a per-usage basis into the project build based on the target browsers above. This works a lot better than in Babel 6.x and that's why all Babel got a version bump as part of this PR.

dash_table/package.json
```diff
@@ -40,17 +41,19 @@
...
+    "@babel/cli": "^7.1.2",
+    "@babel/core": "^7.1.2",
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-react": "^7.0.0",
     "@cypress/webpack-preprocessor": "^2.0.1",
     "@percy-io/percy-storybook": "^2.0.0",
-    "@storybook/cli": "^4.0.0-alpha.16",
-    "@storybook/react": "^4.0.0-alpha.16",
+    "@storybook/cli": "^4.0.2",
+    "@storybook/react": "^4.0.2",
```
Marc-Andre-Rivet: Bump storybook to 4.0.2 so it expects babel 7.x instead of 6.26

demo/App.js
```diff
@@ -14,7 +14,7 @@ class App extends Component {
     constructor() {
         super();
 
-        const data: any[] = clone(mockData.data);
+        const data = clone(mockData.data);
```
Marc-Andre-Rivet: Not sure how this typing ever worked (js file) but it doesn't anymore!

src/core/environment/index.ts
```diff
@@ -10,7 +10,12 @@ interface ISearchParams {
 
 export default class Environment {
     private static get searchParams(): ISearchParams {
-        return new URL(window.location.href).searchParams || { get: () => null };
+        return (
+            typeof URL !== 'undefined' &&
+            URL.prototype &&
+            URL.prototype.constructor &&
+            new URL(window.location.href
+        ).searchParams) || { get: () => null };
```
Marc-Andre-Rivet: The environment / params are only used for debugging purposes and is not used by a "core" feature -- so the shim here is as lightweight as it can be. URL exists in IE 11 but it doesn't have a ctor or this feature.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -148,15 +148,15 @@ export default class ControlledTable extends PureComponent<ControlledTableProps,
 
         // Adjust [fixed columns/fixed rows combo] to fixed rows height
         let trs = r0c1.querySelectorAll('tr');
-        r0c0.querySelectorAll('tr').forEach((tr, index) => {
+        Array.from(r0c0.querySelectorAll('tr')).forEach((tr, index) => {
```
Marc-Andre-Rivet: querySelectorAll returns an array-like object (NodeList), while Array.forEach existed in ES5 and was supported by IE 11, NodeList does not. Converting for use.

PR #225, opened 2018-11-06
Issue 212 - UMD packaging

### LINKED ISSUE #212: JS facing package does not expose the DataTable component
Make it possible to use the DataTable easily from Javascript.

Currently it needs to be loaded like this:

```
import 'dash-table/dash_table/bundle';
import domReady from './domReady';

const DataTable = window.dash_table.DataTable;

ReactDOM.render(<DataTable columns={['a', 'b', 'c'].map(i => ({ id: i, name: i }))} />, document.getElementById('root'));
```

We want to be able to do this:
```
import DataTable from 'dash-table';

// etc...
```

Marc-Andre-Rivet:
Fixes #212.

The table is a fully functional standalone component and the possibility of using it in non-Dash contexts has been discussed already. This PR updates the way the bundle is packaged so that it is compatible with the import/export syntax that is the ES6 best practice.

No special build is needed as Dash seems to behave just fine after the change.

The JS-facing packaging can be tested with
`npm install --save-dev dash-table@rc`

=============================
[Update] Alright.. this ended being *very* messy for a bunch of reasons.. here's a summary. I think this will be useful for the other projects as we have similar approaches.

npm-shrinkwrap.json *forces* package.json dependency and devDependency downloads when present. Since the dash-table pulls no dependencies and is a fully self-contained build, there's no reason to use the shrinkwrap over the package-lock. Made the switch as part of this PR.

We've been using .npmignore to determine what goes into the publish -- turned this on its head by whitelisting files through package.json "files" property instead.. this gives a much cleaner result as all we're interested in are the md files, package.json, license, and build -- only the build needs to be defined explicitly.

/dash_table/package.json side-effects We use the /dash_table folder to both build the JS bundle and to package the PyPi bundle. This wrecked havoc with npm's expectations: It ignored the folder because a package.json was present, making it impossible to bundle it correctly through whitelisting.

The JS build now builds to /dist and is copied over on a 'post' script into dash_table, enabling python packaging to remain the same and for npm to be sane.

I think we should seriously reconsider committing the build artefacts into the repo at some point.


.config/webpack/base.js
```diff
@@ -19,12 +19,22 @@ module.exports = (preprocessor = {}, mode = 'development') => {
             path: path.resolve(__dirname, './../..', dashLibraryName),
             filename: '[name].js',
             library: dashLibraryName,
-            libraryTarget: 'window',
+            libraryTarget: 'umd'
```
Marc-Andre-Rivet: We don't want the dash table to be accessible through `const DataTable = window.dash_table.DataTable'` but rather through `import { DataTable } from 'dash-table'`

.config/webpack/base.js
```diff
@@ -19,12 +19,22 @@ module.exports = (preprocessor = {}, mode = 'development') => {
...
+                commonjs2: "react",
+                amd: "React",
+                root: "React"
+            },
+            "react-dom": {
+                commonjs: "react-dom",
+                commonjs2: "react-dom",
+                amd: "ReactDOM",
+                root: "ReactDOM"
+            },
```
Marc-Andre-Rivet: Support the various naming conventions if the target is using umd or another package loading method.

CHANGELOG.md
```diff
@@ -2,7 +2,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.1.4rc1] - 2018-11-06
+### Added
+- Package library the UMD way [#212](https://github.com/plotly/dash-table/issues/212)
+
 ## [3.1.3] - 2018-11-05
+### Fixed
```
Marc-Andre-Rivet: Oversight from previous issue

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "dash-table",
-  "version": "3.1.3",
+  "version": "3.1.4rc1",
   "description": "Dash table",
-  "main": "build/index.js",
+  "main": "dash_table/bundle.js",
```
Marc-Andre-Rivet: Changing the default target to be the bundle

.gitignore
```diff
@@ -5,6 +5,7 @@
 
 # production
 /build
+/dist
```
Marc-Andre-Rivet: new build folder for js code, copied over to dash_table for python packaging

.npmignore
```diff
@@ -1,31 +0,0 @@
-# dependencies
```
Marc-Andre-Rivet: .npmignore is still present to prevent npm from using .gitignore (default behavior) -- but using whitelisting now instead so the file is just a placeholder

dash_table/package.json
```diff
@@ -19,6 +19,7 @@
     "private::runtests": "cypress run --browser chrome",
     "build.watch": "webpack-dev-server --content-base dash_table --mode development",
     "build:js": "run-s \"private::build -- --mode production\"",
+    "postbuild:js": "cp -R dist dash_table",
```
Marc-Andre-Rivet: after js build, copy over to the python folder

dash_table/package.json
```diff
@@ -77,6 +78,9 @@
     "webpack-dev-server": "^3.1.5",
     "webpack-preprocessor": "^0.1.12"
   },
+  "files": [
+    "/dist/bundle.js"
```
Marc-Andre-Rivet: this is the only thing beside the md files, license and package.json that we want to publish

.gitignore
valentijnnieman: I'm pretty sure the Python build goes into `/dist`, when doing `python setup.py sdist`, so not sure if that's a good idea!

package.json
```diff
@@ -1,12 +1,13 @@
 {
   "name": "dash-table",
-  "version": "3.1.3",
+  "version": "3.1.4-rc6",
   "description": "Dash table",
-  "main": "build/index.js",
```
valentijnnieman: I'm sure you've explained this before - but what was wrong with using `build/` again?

Marc-Andre-Rivet:
There was nothing wrong per say to /build except that it's never been used afaik. /lib and /dist are normal folders for distribution.. /dist is already in use, so went for /lib

For the other comments about renaming the build output, etc. I'd say this is out of scope for this issue but still something we will want to fix along with the internals of our build procedure and the various other points described above.

PR #237, opened 2018-11-09
Issue 235 - Fix regressions introduced to PyPi and NPM packaging

### LINKED ISSUE #235: bundle.js isn't in the npm distribution anymore?
In the latest version, the remote URL for the dash package isn't found:
image

This URL is specified in the `external_url` key here:
https://github.com/plotly/dash-table/blob/10b80296a8577a3efae20f996e71c2917a4d7e51/dash_table/__init__.py#L35-L37

Indeed, looking at the package that got uploaded, the `dash_table` folder is no longer there:
image

So, either `dash_table/bundle.js` needs to get put back _or_ the `external_url` needs to get update to the new directory: `https://unpkg.com/dash-table@3.1.4/lib/bundle.js`

Now, it would be good if it was the former, so that we're standard across our libraries and across our dash-component-boilerplate project. That way, changes that we make to the boilerplate, like adding dev and prod versions of the bundle (done here: https://github.com/plotly/dash-component-boilerplate/pull/8), can get easily incorporated into this project (related issue is https://github.com/plotly/dash-table/issues/228)

cc @Marc-Andre-Rivet and @valentijnnieman

Marc-Andre-Rivet:
Fixes #235. 

Reverting the `/lib` changes to go back to `/dash_table`.

In the end, changing `dash_table/package.json` to `dash_table/package-info.json` to allow npm to do its job correctly and to minimize impact.

This is the same solution as suggested here
https://github.com/plotly/dash-component-boilerplate/issues/38

### DIFF
--- a/dash_table/__init__.py
+++ b/dash_table/__init__.py
@@ -13,7 +13,7 @@ if not hasattr(_dash, 'development'):
     _sys.exit(1)
 
 _basepath = _os.path.dirname(__file__)
-_filepath = _os.path.abspath(_os.path.join(_basepath, 'bundle.json'))
+_filepath = _os.path.abspath(_os.path.join(_basepath, 'package-info.json'))
 with open(_filepath) as f:
     package = json.load(f)
 
--- a/dash_table/__init__.py
+++ b/dash_table/__init__.py
@@ -13,7 +13,7 @@ if not hasattr(_dash, 'development'):
     _sys.exit(1)
 
 _basepath = _os.path.dirname(__file__)
-_filepath = _os.path.abspath(_os.path.join(_basepath, 'package.json'))
+_filepath = _os.path.abspath(_os.path.join(_basepath, 'bundle.json'))
 with open(_filepath) as f:
     package = json.load(f)
 
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "private::build:js-dev": "run-s \"private::build -- --mode development\"",
     "private::build:js-test": "run-s \"private::build -- --config webpack.test.config.js\"",
     "private::build:js-test-watch": "run-s \"private::build -- --config webpack.test.config.js --watch\"",
-    "private::build:py": "./extract-meta src/dash-table/DataTable.js > dash_table/metadata.json && cp package.json dash_table/bundle.json",
+    "private::build:py": "./extract-meta src/dash-table/DataTable.js > dash_table/metadata.json && cp package.json dash_table/package-info.json",
     "private::host_dash8081": "python tests/cypress/dash/v_be_page.py",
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -1,13 +1,17 @@
...
   "scripts": {
     "preprivate::opentests": "run-s private::wait*",
     "preprivate::runtests": "run-s private::wait*",
     "private::build": "node --max_old_space_size=4096 node_modules/webpack/bin/webpack --display-reasons --bail",
-    "postprivate::build": "cp -R lib/* dash_table",
+    "private::build:js": "run-s \"private::build -- --mode production\"",
+    "private::build:js-dev": "run-s \"private::build -- --mode development\"",
+    "private::build:js-test": "run-s \"private::build -- --config webpack.test.config.js\"",
+    "private::build:js-test-watch": "run-s \"private::build -- --config webpack.test.config.js --watch\"",
+    "private::build:py": "./extract-meta src/dash-table/DataTable.js > dash_table/metadata.json && cp package.json dash_table/bundle.json",
```
Marc-Andre-Rivet: Apart from going back to building in /dash_table, this is the major change, when copying package.json, rename it to something else

chriddyp:
FWIW, if you wanted to keep the previous `lib/` structure, I believe that you could do so by just changing 
https://github.com/plotly/dash-table/blob/10b80296a8577a3efae20f996e71c2917a4d7e51/dash_table/__init__.py#L35-L37

to 
```
'https://unpkg.com/dash-table@{}/lib/bundle.js'.format(version)
```

dash_table/__init__.py
```diff
@@ -13,7 +13,7 @@
     _sys.exit(1)
 
 _basepath = _os.path.dirname(__file__)
-_filepath = _os.path.abspath(_os.path.join(_basepath, 'package.json'))
+_filepath = _os.path.abspath(_os.path.join(_basepath, 'package-info.json'))
```
chriddyp: Whenever we want to include new non-Python files in the distribution, we need to add them to `MANIFEST.in`:
https://github.com/plotly/dash-table/blob/10b80296a8577a3efae20f996e71c2917a4d7e51/MANIFEST.in#L1-L5

Python packaging is a white-list rather than a blacklist like `.npmignore`.

Marc-Andre-Rivet:
@chriddyp I know. I'm ok with this with the other changes I suggested in the boilerplate.
MANIFEST.in, that's the list of files for the python package?

chriddyp:
> MANIFEST.in, that's the list of files for the python package?

Yes. It's the list of non-python files in the python package.

dash_table/__init__.py
Marc-Andre-Rivet: Ok cool, missed this comment when I asked at top level.

PR #243, opened 2018-11-15
Generate python components classes

T4rk1n:
Change the python component loading to generated classes files.

One thing to note, `build:py` now requires dash to be installed on the running python environment.


T4rk1n:
@Marc-Andre-Rivet It's for IDE support, they can read the file instead of being generated at runtime.

.config/webpack/base.js
```diff
@@ -39,7 +39,7 @@ module.exports = (preprocessor = {}, mode = 'development') => {
         module: {
             rules: [
                 {
-                    test: /demo[/]index.html?$/,
+                    test: /demo[/\\]index.html?$/,
```
Marc-Andre-Rivet: Oh.. windows!

Issue #203, opened 2018-11-02
fix rows misalignment with scrollbar


chriddyp:
from the docs:
![image](https://user-images.githubusercontent.com/1280389/47889640-566bea00-de21-11e8-9ee6-515f6559acb9.png)

chriddyp:
See also
![image](https://user-images.githubusercontent.com/1280389/48013949-92d66900-e0f3-11e8-8531-4475e845f904.png)

vedulaabhi:
@Marc-Andre-Rivet @chriddyp 

This issue is coming up again in the latest dash-table 3.1.11 version.

vivekvs1:
I am facing this issue with v.4.5.0 as well

dgibbon-plotly:
I think I have the same issue @Marc-Andre-Rivet 

if I use the following:

                    fixed_rows={ 'headers': True, 'data': 0 },

I get:

![Screen Shot 2020-04-27 at 9 26 40 AM](https://user-images.githubusercontent.com/48504233/80396317-50fe2d00-8869-11ea-81a9-d520aa32ce68.png)

om12nayak:
Issue fixed :+1: : 
https://github.com/plotly/dash-table/issues/205#issuecomment-621921028

PR #247, opened 2018-11-20
Issue 216 - Scrollbar alignment

### LINKED ISSUE #216: Side scroll misalignment
Scrolling sideways in one of the tables in the docs here https://dash.plot.ly/datatable/sizing I see header/body misalignment due to scrollbars (?)

image

Marc-Andre-Rivet:
Fixes #216, #203.

After a pretty big false start, using a "phantom" div we can calculate the width of the vertical scrollbar and apply that width as an extra margin for the fixed rows section.. allowing scrolling to stay aligned throughout.

With disappearing scrollbars:
image

With always visible scrollbars:
image


dash_table/package-info.json
```diff
@@ -53,8 +53,8 @@
     "@babel/preset-react": "^7.0.0",
     "@cypress/webpack-preprocessor": "^2.0.1",
     "@percy-io/percy-storybook": "^2.0.0",
-    "@storybook/cli": "^4.0.2",
-    "@storybook/react": "^4.0.2",
+    "@storybook/cli": "^4.0.7",
+    "@storybook/react": "^4.0.7",
```
Marc-Andre-Rivet: Having issues running storybook locally, updating fixes the issue -- local dep of react needs to be updated

src/core/browser/scrollbarWidth.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    return new Promise<number>(resolve => {
+        setTimeout(() => {
+            const width = child.clientWidth - parent.clientWidth;
+
+            document.body.removeChild(parent);
+            resolve(width);
+        }, 0);
+    });
+};
```
Marc-Andre-Rivet: Create a phantom div with a child to calculate the width of the scrollbar -- if the scrollbar disappears, it will have 0 width

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -144,6 +147,8 @@ export default class ControlledTable extends PureComponent<ControlledTableProps,
 
         this.updateStylesheet();
 
+        getScrollbarWidth().then((scrollbarWidth: number) => this.setState({ scrollbarWidth }));
```
Marc-Andre-Rivet: Re-evaluate scrollbar width on resize (to cover zoom case) and update state

src/core/browser/scrollbarWidth.ts
Marc-Andre-Rivet: (Promise polyfill for IE11 is managed automatically by Babel)

vedulaabhi:
Hi,

This Abhishek from Invesco. Just wanted to know if this issue is fixed and if not is there any workaround from styling perspective included in our code.

Marc-Andre-Rivet:
@vedulaabhi This PR should fix the issue in all known scenarios on all Mac and Windows browsers supported by the table. No additional workaround is necessary to make the fix work.

vedulaabhi:
@Marc-Andre-Rivet will you be including this in the next release and if so can you pls let me know when we can expect the release as this is really important for our ispace app.

Marc-Andre-Rivet:
@vedulaabhi Once the review goes through this will be released as 3.1.7 -- this *is* the next release :)

Issue #250, opened 2018-11-22
Make our own dropdown component (phase out react-select)


Marc-Andre-Rivet:
We are currently locked into a old(ish) version 1.2.1 of https://www.npmjs.com/package/react-select and can't easily migrate to the new 2.x version because of a hardcoded styling dependency (https://github.com/plotly/dash-table/blob/c031ca82ea26368b8b496f48dd56a1a5820fa493/src/dash-table/components/Table/index.tsx#L20)

We've encountered multiple styling and implementation issues with this dropdown (height changing, width forcing, positioning issues within a table with fixed rows/columns -- causing special implementation needs for said "fixed" options, etc.) and we should consider either using another pre-made component or building our own to better fit the needs of the table.

bpostlethwaite:
I support this and not only for dash-table but across all our dash repos. Though we can and probably should do this incrementally and start in dash-table. react-select isn't saving us time given how complex its styling is and how brittle our react-select styling integration has become. It's a freaking dropdown.

Marc-Andre-Rivet:
@bpostlethwaite Yes!

valentijnnieman:
I'm currently looking into some DCC.Dropdown bugs, most of them I can't fix because they're residing in the `react-virtualized-select` package, which is no longer maintained. That package uses `react-select` version 1.0 internally by the way, and even though our deps have `react-select 2.0` (which isn't used anywhere) we're still stuck with version 1.0. I'm currently trying to replace `react-virtualized-select` with `react-select`, but not sure about that solution (large datasets?), but if we're building our own Dropdown component... I'd say that's a much better solution!

PR #251, opened 2018-11-22
3.1 update dependencies

Marc-Andre-Rivet:
Just basic maintenance work, updating deps and clean up


dash_table/package-info.json
```diff
@@ -37,50 +37,49 @@
...
   "license": "MIT",
-  "dependencies": {
-    "@types/react": "^16.4.8",
-    "ramda": "^0.25.0",
-    "react": "16.3.0",
-    "react-dom": "16.3.0",
-    "react-select": "^1.2.1",
-    "sheetclip": "^0.3.0"
-  },
+  "dependencies": {},
```
Marc-Andre-Rivet: There are multiple ways of looking at the meaning of deps and devDeps in the npm world -- but in the end this project only exposes a bundled file containing *all* that's needed and that can be seen as there being only dev dependencies as there is nothing that is required outside that bundle in the production environment / to use the package. This is the approach taken here.

dash_table/package-info.json
```diff
@@ -37,50 +37,49 @@
...
-    "typescript": "^3.0.1",
-    "wait-on": "^2.1.0",
-    "webpack": "^4.8.3",
-    "webpack-cli": "^2.1.3",
-    "webpack-dev-server": "^3.1.5",
+    "typescript": "^3.1.6",
+    "wait-on": "^3.2.0",
+    "webpack": "^4.26.0",
+    "webpack-cli": "^3.1.2",
+    "webpack-dev-server": "^3.1.10",
```
Marc-Andre-Rivet: A bunch of minor and some major version updates.

dash_table/package-info.json
```diff
@@ -37,50 +37,49 @@
...
+    "npm": "^6.4.1",
     "npm-run-all": "^4.1.3",
-    "react-docgen": "^2.20.1",
-    "style-loader": "^0.21.0",
-    "ts-loader": "^4.4.2",
+    "ramda": "^0.26.0",
+    "react": "16.6.3",
+    "react-docgen": "^2.21.0",
+    "react-dom": "16.6.3",
+    "react-select": "^1.2.1",
```
Marc-Andre-Rivet: There is a new version 2.1.2 of this but we can't update because of internal dependencies on this version's structure. See #250

PR #291, opened 2018-12-10
Issue 275 - Remove deprecated nested property 'displayed_pages'

### LINKED ISSUE #275: Remove usage of 'displayed_pages' for pagination
- displayed_pages is deprecated as per documentation but still used by derived/data/viewport.ts
- remove from DataTable prop 'pagination_settings'

Marc-Andre-Rivet:
Fixes #275.

- Removing the nested property 'displayed_pages' from 'paginationation_settings'
- Removing all usages in the tests and code

Note: This is not a breaking change to the contract as removing the nested property from the prop won't cause a Dash error if it's still being passed, as opposed to a top-level prop being removed. e.g. the same can't be said of 'navigation' and 'dropdown_properties'


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-table",
-  "version": "3.1.9",
+  "version": "3.1.10",
```
Marc-Andre-Rivet: bump version

src/dash-table/derived/data/viewport.ts
```diff
@@ -20,7 +20,7 @@ function getFrontEndPagination(settings: IPaginationSettings, data: Data, indice
 
     const firstIndex = settings.page_size * currentPage;
     const lastIndex = Math.min(
-        firstIndex + settings.displayed_pages * settings.page_size,
+        firstIndex + settings.page_size,
```
Marc-Andre-Rivet: removed and implicitly using the value 1

tsconfig.json
```diff
@@ -10,6 +10,11 @@
...
     },
     "exclude": [
-        "tests/*"
+        "dash_table/*",
+        "dist/*",
+        "node_modules/*",
+        "storybook-static/*",
+        "tests/*",
+        "vv/*"
     ]
```
Marc-Andre-Rivet: Unrelated, just to help VSCode decide what code it needs to analyze for JS/TS typing, etc.

PR #309, opened 2018-12-19
Test Dash table against head of master and release-v1

Marc-Andre-Rivet:
Fixes https://github.com/plotly/dash-core/issues/20.

The table modification is a bit different from what will happen for the other repos as the master branch will support both the v0 and v1 of Dash.

- server specific tests are run against both v0 and v1
- python visual tests are run against both v0 and v1
- execute against head version of release-v1 and master (this will need to be updated to release-v0 and master upon release of v1)
- prevent caching on requirements (as it's a moving target)


Marc-Andre-Rivet:
These checks do not exist anymore and have been supplanted.
ci/circleci: python-3.6
ci/circleci: test
percy/dash-table-python

.circleci/config.yml
```diff
@@ -34,11 +34,59 @@ jobs:
...
+                name: Run build:py
+                command: |
+                    . venv/bin/activate
+                    npm run private::build:py
+
+            - run:
+                name: Run tests
+                command: |
+                    . venv/bin/activate
+                    npm run test-v0
```
Marc-Andre-Rivet: test v0 stuff -- we only run server tests

.circleci/config.yml
```diff
@@ -54,7 +102,7 @@ jobs:
                 name: Run tests
                 command: |
                     . venv/bin/activate
-                    npm run test
+                    npm run test-v1
```
Marc-Andre-Rivet: for v1 we run everything (server, standalone w/ React 16, unit tests)

.circleci/config.yml
```diff
@@ -110,24 +158,21 @@ jobs:
...
             - checkout
 
             - run:
                 name: Inject Percy Environment variables
                 command: |
-                  echo 'export PERCY_TOKEN="$PERCY_PYTHON_TOKEN"' >> $BASH_ENV
-
-            - restore_cache:
-                key: deps1-{{ .Branch }}-{{ checksum "requirements.txt" }}
+                  echo 'export PERCY_TOKEN="$PERCY_PYTHON_TOKEN_V0"' >> $BASH_ENV
```
Marc-Andre-Rivet: v0 and v1 visual tests end up in https://percy.io/plotly/dash-table-python-v0 and https://percy.io/plotly/dash-table-python-v1

tests/dash/app_dropdown.py
```diff
@@ -13,7 +13,8 @@
 ID_PREFIX = "app_dropdown"
 IDS = {
     "dropdown": ID_PREFIX,
-    "dropdown-by-cell": '{}-row-by-cell'.format(ID_PREFIX)
+    "dropdown-by-cell": '{}-row-by-cell'.format(ID_PREFIX),
+    "dropdown-by-cell-deprecated": '{}-deprecated-row-by-cell'.format(ID_PREFIX)
```
Marc-Andre-Rivet: Not certain what happen but I think the recent change to renderer (keys/ids) is responsible for the regression -- both table had the same id -- apparently causing the second one to disappear from the DOM entirely.

Marc-Andre-Rivet: https://github.com/plotly/dash-renderer/pull/101

Marc-Andre-Rivet: release-v1 did not have this problem because the latest master changes haven't been merged in yet -- process yet to be put in place :)

alexcjohnson: Oh interesting - that's probably going to bite us and our users a lot of places. Should we make it an exception on the Python side if you ever provide multiple components with matching id?

T4rk1n: > make it an exception on the Python side if you ever provide multiple components with matching id?

There is a duplicate ids check: https://github.com/plotly/dash/pull/320

alexcjohnson: > There is a duplicate ids check: plotly/dash#320

why wasn't that triggered here?

T4rk1n: Layout is dynamic, it doesn't work when that is the case.

requirements.txt
```diff
@@ -1,60 +1,2 @@
...
-six==1.11.0
-toml==0.9.6
-traitlets==4.3.2
-typed-ast==1.1.0
-urllib3==1.23
-wcwidth==0.1.7
-Werkzeug==0.14.1
-wrapt==1.10.11
+-r requirements-base.txt
+-r requirements-v1.txt
```
Marc-Andre-Rivet: @alexcjohnson So interestingly Heroku does not seem super flexible about its setup procedure. The project has been tagged as Python and as far as I understand, that means it will look for setup.py, runtime.txt and requirements.txt -- so whatever weird multi version setup we have, we must play around that limitation.

alexcjohnson: Annoying, but nice solution!

PR #327, opened 2019-01-10
Issue 297 - Deeper support for 'number' and 'text' types

Marc-Andre-Rivet:
- remove 'dropdown' type and make it a presentation detail instead
- rename 'numeric' to 'number'
- add 'any' type
- add per type coercion / reconciliation / validation on edit to props contract
- add presentation logic to props contract

- [ ] explain contract logic & props in PR
- [ ] document dash props
- [ ] write unit tests


demo/data.ts
```diff
@@ -11,51 +12,54 @@ export interface IDataMock {
 export const generateMockData = (rows: number) => unpackIntoColumnsAndData([
     {
         id: 'rows',
-        type: 'numeric',
+        type: ColumnType.Number,
```
Marc-Andre-Rivet: Using the type enum instead of a string..

demo/data.ts
```diff
@@ -11,51 +12,54 @@ export interface IDataMock {
...
     },
 
     {
         id: 'bbb',
         name: ['', 'Weather', 'Climate'],
-        type: 'dropdown',
+        type: ColumnType.Text,
+        text: {
+            presentation: 'dropdown'
+        },
```
Marc-Andre-Rivet: The dropdown type does not exist anymore, instead it is managed through the type's configuration nested prop -- for text the presentation options are 'input' and 'dropdown' at the moment

src/dash-table/DataTable.js
```diff
@@ -178,6 +178,31 @@ export const propTypes = {
...
+            presentation: PropTypes.oneOf(['input', 'dropdown']),
+            validation: PropTypes.shape({
+                allow_nan: PropTypes.bool,
+                on_change: PropTypes.oneOf([
+                    'nan',
+                    'passthrough',
+                    'prevent',
+                    'skip'
+                ])
+            })
```
Marc-Andre-Rivet: coercion / validation / reconciliation are affected by the validation settings

on_change defines how the table will behave if the value cannot be coerced into the destination type (number)
- nan will set the value NaN
- passthrough will set the value as-is without coercion (this provides a default behavior consistent with the current table behavior)
- skip will not set the value but will not prevent larger scope modifications (e.g. during a copy/paste)
- prevent will not set the value and will prevent larger scope modifications (the entire copy/paste operation will be cancelled if one cell's data is incompatible)

src/dash-table/DataTable.js
```diff
@@ -178,6 +178,31 @@ export const propTypes = {
...
+
+        text: PropTypes.shape({
+            presentation: PropTypes.oneOf(['input', 'dropdown']),
+            validation: PropTypes.shape({
+                allow_nully: PropTypes.bool,
+                on_change: PropTypes.oneOf([
+                    'passthrough',
+                    'prevent',
+                    'skip'
+                ])
```
Marc-Andre-Rivet: maybe something like "convert" and "convert simple types" could be interesting

src/dash-table/DataTable.js
```diff
@@ -639,7 +668,8 @@ export const propTypes = {
     style_cell_conditional: PropTypes.arrayOf(PropTypes.shape({
         // .exact
         if: PropTypes.shape({
-            column_id: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+            column_id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+            type: PropTypes.oneOf(['any', 'number', 'text'])
```
Marc-Andre-Rivet: add the ability to filter cells & headers by data type

src/dash-table/components/CellFactory.tsx
```diff
@@ -92,9 +85,21 @@ export default class CellFactory {
             dropdown_properties
         );
 
+        const presentations = derivedCellPresentations(columns);
```
Marc-Andre-Rivet: A bit more complex than just picking up the type -- evaluating once for reuse in later steps

src/dash-table/utils/applyClipboardToData.ts
```diff
@@ -22,8 +23,9 @@ export default (
         Logger.debug(`Clipboard -- Do not create new columns`);
     }
 
-    let newData = data;
-    const newColumns = columns;
+    // don't modify the data and columns directly -- we may abort the paste
+    let newData = R.clone(data);
+    const newColumns = R.clone(columns);
```
Marc-Andre-Rivet: start by cloning the data, we may abort the copy/paste if a "prevent" happens

Issue #166, opened 2018-10-24
data types architecture (sponsored: due feb 1)


chriddyp:
placeholder issue for 1st class data type support in dash-table

Phase I - Type support
- [x] #280 (prep work)
- [x] #297 (first pass at data type support -- for existing types)
- [x] [Datetime type](https://github.com/plotly/dash-table/pull/364)
- [x] #342 (support conditionals based on column type)
- [ ] type inference with filtering
- [ ] type inference / precedence with sorting
- [ ] #332 (make numeric columns return number when possible instead of string)
- [ ] update dash docs
- [ ] clearly document the upgrade path (3.1 --> 3.2)

Phase II+
Related issues:
- [x] #189 type=number formatting
- #232 
- #248
- #305 (number styling)
- #331 (date picker)

chriddyp:
There is a lot to consider here. I've started considering 'what data types means in the table' in this gist, https://gist.github.com/chriddyp/afcaf01d124b610e69de066753d36fc5, and I'll update this thread as I make more progress spec'ing this out.

chriddyp:
Great first pass on data types 🚀 . We can create separate issues for more features as they come up.

PR #329, opened 2019-01-11
Issue 297 - Improve type support

### LINKED ISSUE #297: Deeper support for 'number' and 'text' types
Relates to the #166 epic.

1 - drop support for 'dropdown' column type
A dropdown is not a type, hence it makes sense to move it out of there. It is also not quite a representation of the data because it is limiting the available values, not defining the presentation type.

As a first step, moving this out of type into its own flag seems to make sense. It will probably need to be revisited in further steps as the API gets more defined.

2 - add new column nested prop 'enumeration' that takes values true | false | 'maybe'
A cell can either be always be a dropdown, never be a dropdown or sometimes be a dropdown. true and false behaves like the old type: 'dropdown' while 'maybe' will display the default representation for its type (e.g. an input) if there is no associated dropdown values -- and a dropdown if values are found

4 - add data coercion on user-initiated value change and on copy/paste
Validate numbers and strings and coerce into the destination type on those actions.

5 - define behaviour on failed coercion for user-initiated value change and copy/paste
Coercion and validation means that sometimes the validation will fail. We need to define the behaviour for those cases. Does copy/paste with some valid values and some invalid values partially work or fail? What is the user-interaction / error / warning for those cases / should a number default to NaN?

One possibility would be to leave it to the user (+ default behaviour), for example:

```
number: {
    validation: {
        on_change:  'nan' | 'skip' | 'prevent' | 'passthrough'
    }
}
```

Where NaN would assign NaN as a value, skip would simply skip the invalid data / cell and prevent  would prevent a larger copy/paste from working at all if a column with 'prevent' was to fail validation, 'passthrough' just let's the value go through as-is if coercion fails. Would default to 'passthrough' initially and full implementation would require ux elements from the out of scope points below.

6 - what do we do with initial data received from server -- do we assume the values are sane? do we coerce them -- if so, what is the end-result, user-interaction, etc. like in point 5.
Up to now we have been considering that this data is sane and have had the same assumptions when filtering and sorting -- I think that for this issue it should stay that way.

7 - 'any' type essentially behaves like the current implementation -- the table doesn't enforce anything

API contract would look like this:

```
columns: [{
    type: 'number' | 'text' | 'any',
    enumeration: boolean | 'maybe',
    number /* TBD */: {
        representation: 'input' | 'slider' | 'bar',
        validation: /* TBD */,
        formatter: /* TBD */
    },
    text /* TBD */: {
        representation: 'input' | 'textarea',
        validation: /* TBD */
    },
    /* etc for other types */
}, ...]
```

Follow up tasks include
- additional types
- fleshing out the `type` configurations (representation, validation, etc.)
- adding conditional styling on column type
- filtering, sorting improvements
- data sanitation?
- ux for failed validation

Marc-Andre-Rivet:
Fixes #297 

- add `column.presentation` to decide how to display the data
- add `column.type` = 'any'
- add on_change action `column.on_change.action` defining the table's coercion/validation behaviour on user-provided data
- add on_change failure action `column.on_change.failure` defining the table's reconciliation behaviour on failed user-provided data
- remove `column.type` 'dropdown', replace it with `column.presentation` = 'dropdown'

- add by-type unit tests for action and failure action
- add strongly typed table integration tests (standalone)


src/dash-table/DataTable.js
```diff
@@ -178,6 +178,45 @@ export const propTypes = {
...
+                action: PropTypes.oneOf([
+                    'passthrough',
+                    'coerce',
+                    'validate'
+                ]),
+                failure: PropTypes.oneOf([
+                    'default',
+                    'prevent',
+                    'skip'
+                ])
```
Marc-Andre-Rivet: action/failure processing is identical for text -- hoping that the division in two steps + the validation options should be enough to cover all the flavoring without having to make custom lists by type

src/dash-table/DataTable.js
```diff
@@ -198,16 +237,20 @@ export const propTypes = {
 
         /**
          * The data-type of the column's data.
-         * `numeric` and `text` are currently the same.
-         * `dropdown` is used to render cells as dropdowns.
+         * 'number': represents both floats and ints
+         * 'text': represents a string
+         * 'any': represents any type of data
+         *
+         * Defaults to 'any' if undefined.
```
Marc-Andre-Rivet: The columns default to 'any' type if not defined and defaults to a bare boned type configuration (on_change, validation, presentation) if it's not defined, conserving a behaviour that should be consistent with the behaviour before the change

src/dash-table/components/Table/props.ts
```diff
@@ -76,6 +121,8 @@ export interface IVisibleColumn {
     editable_name?: boolean | number;
     id: ColumnId;
     name: string | string[];
+    number?: INumberConfiguration;
+    text?: ITextConfiguration;
```
Marc-Andre-Rivet: all type configuration is optional (and tagged with '?' for that purpose)

src/dash-table/handlers/cellEvents.ts
```diff
@@ -98,9 +99,15 @@ export const handleChange = (propsFn: () => ICellFactoryProps, idx: number, i: n
         return;
     }
 
+    const result = reconcile(value, c);
+
+    if (!result.success) {
+        return;
+    }
```
Marc-Andre-Rivet: Make sure the input data can be reconciled with the destination column, otherwise skip

src/dash-table/reconcile/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            break;
+        case ChangeAction.Passthrough:
+            res = { success: true, value, action };
+            break;
+        case ChangeAction.Validate:
+            res = { action, ...(getValidator(c) as any)(value, options) };
+            break;
+    }
+
+    if (res.success) {
```
Marc-Andre-Rivet: if coerce/validate was successful, no need to process the failure

src/dash-table/reconcile/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // If c/v unsuccessful, process failure
+    const failure = (config && config.on_change && config.on_change.failure) || ChangeFailure.Skip;
+    res.failure = failure;
+
+    // If Skip/Prevent
+    if (failure !== ChangeFailure.Default) {
+        return res;
+    }
```
Marc-Andre-Rivet: if there's no default value, the failure can't be salvaged

src/dash-table/utils/applyClipboardToData.ts
```diff
@@ -56,8 +58,8 @@ export default (
     const lastEntry = derived_viewport_indices.slice(-1)[0] || 0;
     const viewportSize = derived_viewport_indices.length;
 
-    values.forEach((row: string[], i: number) =>
-        row.forEach((cell: string, j: number) => {
+    for (let [i, row] of values.entries()) {
+        for (let [j, value] of row.entries()) {
```
Marc-Andre-Rivet: using for-of instead of forEach makes it possible to terminate the loop execution (return in fn scope instead of lambda)

tests/visual/percy-storybook/fixtures.ts
```diff
@@ -7,7 +9,10 @@ export default [
                 {
                     name: 'Column 1',
                     id: 'column-1',
-                    type: 'dropdown',
+                    type: ColumnType.Text,
```
Marc-Andre-Rivet: update the percy fixtures tests like AppMode and data above

demo/data.ts
```diff
@@ -11,51 +12,54 @@ export interface IDataMock {
...
         data: gendata(i => i * 10, rows)
     },
 
     {
         id: 'bbb',
         name: ['', 'Weather', 'Climate'],
-        type: 'dropdown',
+        type: ColumnType.Text,
+        text: {
+            presentation: 'dropdown'
```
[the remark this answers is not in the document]
alexcjohnson: Why nest this inside `text` (or `number`)? Can't an `any` also be presented as a dropdown?

Marc-Andre-Rivet:
@alexcjohnson Reduced the nesting and reused the same column-level properties for the `text` and `number` types. The Dash typing is a bit looser than it was but not by much.

src/dash-table/DataTable.js
```diff
@@ -178,6 +178,45 @@ export const propTypes = {
...
         ]).isRequired,
 
+        number: PropTypes.shape({
+            presentation: PropTypes.oneOf(['input', 'dropdown']),
+            on_change: PropTypes.shape({
+                action: PropTypes.oneOf([
+                    'passthrough',
+                    'coerce',
+                    'validate'
+                ]),
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: Nitpicking about nomenclature... maybe:

action: `none`, `coerce`, `validate`
failure: `accept`, `reject`, `default` (clean up the global vs. local scope messiness that comes with skip/reject)

Make the copy/paste all-or-nothing live elsewhere -- e.g. as a table level configuration -- or simply pick one behaviour and out-of-scope the configurable behaviour

demo/data.ts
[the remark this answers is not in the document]
Marc-Andre-Rivet: I guess it can. Adding it.

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Added
+- Data type support & coercion / reconciliation / validation [#297](https://github.com/plotly/dash-core/issues/297)
+### Regression
```
alexcjohnson: why **Regression** instead of **Changed**?

src/dash-table/DataTable.js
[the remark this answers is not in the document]
alexcjohnson: I like `accept`, `reject`, and `default`. And I agree, all-or-nothing vs cell-by-cell would be cleaner as table-level config. Seems like it could be very confusing if garbage has different effects on the rest of the table depending on which column it's in.

`none` is nicer than `passthrough`, but I still don't see its purpose - how would that be different from `type: 'any'`? On slack you mentioned this would be used if:
> you don’t care that much but maybe you want the initial data to be sane

Is initial data treated differently from changes in some way? Anyway there's also `action: 'validate', failure: 'accept'` which would also never change the value provided, right?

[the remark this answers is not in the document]
Marc-Andre-Rivet: So `none` might be interesting and not redundant with `type: any` when taking into consideration wider data type functionality. For example, we want to make the filter type-aware so that entering `> 50` in a number column does not have the same meaning as in a text column. Sorting with type awareness could similarly be impacted. I'm not certain how much of this really requires a `none` or how much of it is a bad idea. We should probably discuss this further.

As currently implemented, initial data isn't handled by the `on_change` prop, neither are programmatically driven changes -- only user-initiated changes are.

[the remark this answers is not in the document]
alexcjohnson: Filtering and sorting FTW - `action: 'none'` is certainly more intuitive than `action: 'validate', failure: 'accept'` even if the result is equivalent. So OK, let's keep `'none'`.

Marc-Andre-Rivet:
@alexcjohnson After reading up on the NaN support in JSON in general it turns out that `null` is OK and the expected dumped value. Modified the validation interface from `allow_nan`+`allow_nully` to simply `allow_null`. If JS encounters `NaN` it will force it to `null` internally to be consistent with the exposed interface.

alexcjohnson:
OK, thanks for checking into it. A single `allow_null` sounds great. We'll probably want to mention this behavior in the docs somewhere (I guess where we describe `allow_null`?) as it's guaranteed to trip people up at some point.

Marc-Andre-Rivet:
This is where we document the prop
https://github.com/plotly/dash-table/blob/d4098aabbcd044d871b447372901fd3c64b6c136/src/dash-table/DataTable.js#L215

So far we've been expecting people that want the detailed contract to go dig in that file. We should be able to take the generated metadata.json and expose that contract in a more user friendly location in the future (e.g. in dash-docs/datatable) -- created an issue to follow that up here https://github.com/plotly/dash-core/issues/33

Marc-Andre-Rivet:
Also added this as out of scope for this issue but in scope for phase 1 of #166 
#332

src/dash-table/reconcile/number.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function coerce(value: any, options?: INumberColumn) {
+    const validation = validate(value, options);
+    if (validation.success) {
+        return validation;
+    }
+
+    switch (typeof value) {
+        case 'string':
+            return { success: Boolean(/(-)?\d+([.]\d+)?/.test(value)), value: parseFloat(value) };
```
alexcjohnson: Depending on how complete you want to be here, there can be a lot of other numbers... https://github.com/plotly/fast-isnumeric

src/dash-table/reconcile/text.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (validation.success) {
+        return validation;
+    }
+
+    const allowNull = Boolean(options && options.validation && options.validation.allow_null);
+    const nully = isNully(value);
+
+    return {
+        success: Boolean(!nully || allowNull),
+        value: nully ? value : JSON.stringify(value)
```
alexcjohnson: I was going to say that the nully case was already handled by `validate`, but I guess you're not stringifying nully values? Is that what we want for text/coerce? I would have thought that even they should coerce to strings.

src/dash-table/reconcile/number.ts
Marc-Andre-Rivet: Yeah, I did this to get basic support but switching to `fast-isnumeric` is trivial. Will update.

alexcjohnson: Note that, though it's not really described there, `fast-isnumeric` is based on using type casting (plotly.js for example does [`+value`](https://github.com/plotly/plotly.js/blob/39915526327c657055feebc5e13b762616d52ebd/src/lib/coerce.js#L90) rather than `parseFloat`). It looks like the only case where `parseFloat` behaves differently (among values that pass `fast-isnumeric`) is hex strings (`'0xff'` - `parseFloat` just drops everything after the first number it finds), and it would be worthwhile deciding whether we want to accept hex (I would argue yes, and let it be converted to decimal unless and until we support hex formatting).

The other weird thing `parseFloat` does is drop any trailing junk (hence its handling of hex) which I don't think we want. Your test here, as written, ignores leading AND trailing junk... we definitely don't want that!

Marc-Andre-Rivet: I was young and optimistic?

Marc-Andre-Rivet: Supporting hex input sounds like a win to me. If we need hex presentation later (or type!), cool. I don't really see a scenario where someone should care that the value came in as `120` or as `0x78`...

src/dash-table/reconcile/text.ts
alexcjohnson: > If we need/want to support the conversion of nully values, it should be done through another prop.

That all makes sense, sold. Thanks for expanding on your thought process.

src/dash-table/reconcile/number.ts
```diff
@@ -0,0 +1,17 @@
+import isNumeric from 'fast-isnumeric';
+
+import { INumberColumn } from 'dash-table/components/Table/props';
+import { reconcileNull } from './null';
+import { IReconciliation } from '.';
+
+export function coerce(value: any, options: INumberColumn | undefined): IReconciliation {
+    return isNumeric(value) ?
```
Marc-Andre-Rivet: Updated number reconciliation using fast-isnumeric + reworked nully handling

src/dash-table/reconcile/text.ts
```diff
@@ -0,0 +1,17 @@
...
+        typeof value === 'string' ?
+            { success: true, value } :
+            { success: true, value: JSON.stringify(value) };
+}
+
+export function validate(value: any, options: ITextColumn | undefined): IReconciliation {
+    return typeof value === 'string' ?
+        { success: true, value } :
+        reconcileNull(value, options);
+}
```
Marc-Andre-Rivet: Reworked logic for text handling

@Types/modules.d.ts
```diff
@@ -6,5 +6,11 @@ declare module 'sheetclip' {
         }
     };
 
+    export default value;
+}
+
+declare module 'fast-isnumeric' {
+    const value: (value: any) => boolean;
+
```
Marc-Andre-Rivet: fast-isnumeric has no typing info -- declaring the module's signature as part of this project instead so TS allows its use

src/dash-table/reconcile/number.ts
```diff
@@ -1,33 +1,17 @@
...
-
-    value = nully ? null : value;
-
-    if (typeof value === 'number') {
-        return { success: !nully, value };
-    } else {
-        return { success: false, value };
-    }
+export function validate(value: any, options: INumberColumn | undefined): IReconciliation {
+    return typeof value === 'number' && !isNaN(value) ?
```
alexcjohnson: `isFinite` will filter out `+/-Infinity` in addition to `NaN`

src/dash-table/reconcile/null.ts
```diff
@@ -0,0 +1,20 @@
...
+export const reconcileNull = (
+    value: any,
+    options: ITypeColumn | undefined
+): IReconciliation => {
+    const allowNull = Boolean(options && options.validation && options.validation.allow_null);
+    const nully = isNully(value);
+
+    return {
+        success: nully && allowNull,
+        value: nully ? null : value
```
alexcjohnson: `reconcileNull` is very nice! Especially collapsing all nully values to actual `null` here 🏆

src/dash-table/reconcile/number.ts
Marc-Andre-Rivet: Checked that `+/-Infinity` also resolves to `null` -- will add them to the nully cases

tests/cypress/tests/unit/numberCoercion_test.ts
```diff
@@ -1,13 +1,17 @@
...
     { input: 42, output: 42, name: 'from number' },
     { input: '42', output: 42, name: 'from number string' },
     { input: '-42', output: -42, name: 'from negative number string' },
     { input: '4.242', output: 4.242, name: 'from float string' },
-    { input: '-4.242', output: -4.242, name: 'from negative float string' }
+    { input: '-4.242', output: -4.242, name: 'from negative float string' },
+    { input: '0x2A', output: 42, name: 'from negative hex string' },
+    { input: '0o52', output: 42, name: 'from negative octal string' },
+    { input: '42e0', output: 42, name: 'from negative exponent string' },
+    { input: '0b101010', output: 42, name: 'from negative exponent string' }
```
alexcjohnson: some copypasta in these `name`s

src/dash-table/reconcile/number.ts
alexcjohnson: Ah right - that'll keep us pretty safe. If anyone ever finds a way around that we can address it then 😎

PR #338, opened 2019-01-21
Issue 307 - Tooltip support

### LINKED ISSUE #307: Tooltip Support [Sponsored, Feb 1 Target]
Some requirements:
- Ability to display tooltips when hovering over cells.
- Tooltip data will be provided as an additional property in the table
- Each tooltip string will be matched with a cell via a row ID and a column ID so that the tooltips remain associated with the cells when filtering and sorting.
- Tooltip strings will be rendered as plain text or as Markdown. For security reasons, raw HTML will not be supported. For architectural reasons, users will not be able to pass arbitrary Dash components as tooltips. Markdown strings will enable bolded text, italics, line breaks, headers, and tables.
- Tooltips will be styleable via external CSS.
- The position of the tooltip will be automatically determined so that it:
  - Doesn’t block the cell
  - Isn’t hidden
- Users will not necessarily be able to mouse over the tooltip itself to select text or click on links. Doing so would prevent the tooltip from disappearing, a potentially confusing experience.

Marc-Andre-Rivet:
Fixes #307.

- Adds markdown library
- Adds dynamic imports for markdown library

- Add props for defining tooltips (`tooltips`, `column_static_tooltip`, `column_conditional_tooltips`)
- Add CellTooltip component to augment the table with cursor-associated-tooltip functionality


src/dash-table/DataTable.js
```diff
@@ -537,6 +540,48 @@ export const propTypes = {
         })).isRequired
     })),
 
+    column_static_tooltip: PropTypes.objectOf(PropTypes.shape({
```
Marc-Andre-Rivet: `objectOf` defines the shape of each property but not the name of those properties, this is exactly equivalent to the TS typing

```
interface IMyInterface {
    [key: string]: IMyInterfaceProp;
}
```

Marc-Andre-Rivet: I was unaware before that (a) PropTypes supported this, (b) Dash supported this (we support a pretty comprehensive subset of PropTypes but not *quite* all of it afaik) -- there's some debt in other types (especially conditional dropdowns) that would benefit from type improvements / better React props validation

src/dash-table/components/Cell/index.tsx
```diff
@@ -20,15 +20,15 @@ export default class Cell extends Component<ICellProps> {
...
+        const { attributes, classes, style } = this.propsWithDefaults;
 
         return (<td
             ref='td'
             children={(this as any).props.children}
             tabIndex={-1}
             className={classes}
             style={style}
-            data-dash-column={property}
+            {...attributes}
```
Marc-Andre-Rivet: Pass all data-attributes structure for passing column id and row index information to the cell -- used by the tooltip decorator to figure out what it's hovering over

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
```diff
@@ -0,0 +1,201 @@
...
+                            );
+                        })()
+                    )
+                );
+        }, column_conditional_tooltips);
+
+        return conditionalTooltips.length ?
+            conditionalTooltips[0] :
+            legacyTooltip || staticTooltip;
+    }
```
Marc-Andre-Rivet: This `if` block of logic is very closely related to another one for the filters -- need to refactor / cleanup usage

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
```diff
@@ -0,0 +1,201 @@
...
+        const xds = xts - xvs;
+        const xde = xve - xte;
+
+        if (xde >= 0) {
+            return;
+        }
+
+        // the tooltip can only exceed the view in one direction (right)
+        // bound the adjustment to whatever is smallest (the right excess, or the left space)
+        tooltip.style.marginLeft = `${Math.max(xde, -xds)}px`;
```
Marc-Andre-Rivet: Making sure the tooltip is visible based on the x-axis positioning and viewport

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
```diff
@@ -0,0 +1,201 @@
...
+    render() {
+        let children = (this.props as any).children;
+
+        const { displayTooltip } = this.state;
+
+        let tooltip = (displayTooltip || undefined) && this.getTooltip();
+
+        let tooltipComponent;
+        if (tooltip && !this.state.md) {
+            import(/* webpackChunkName: "markdown" */ 'remarkable')
```
Marc-Andre-Rivet: Dynamically load the markdown library. The component's state will be update when it becomes available to re-render itself correctly.

src/dash-table/components/Table/index.tsx
```diff
@@ -57,7 +58,15 @@ export default class Table extends Component<PropsWithDefaultsAndDerived, Standa
...
-        return (<ControlledTable {...this.controlled} />);
+        return (<TooltipCursorDecorator
+            column_static_tooltip={this.controlled.column_static_tooltip}
+            column_conditional_tooltips={this.controlled.column_conditional_tooltips}
+            tooltips={this.controlled.tooltips}
+            delay={500}
+            virtualized={this.controlled.virtualized}
+        >
+            <ControlledTable {...this.controlled} />
+        </TooltipCursorDecorator>);
```
Marc-Andre-Rivet: The tooltip decorator (a) shows the tooltip, (b) wraps up whatever is under its sphere of influence (e.g the ControlledTable) -- the table is unaware that it is decorated with tooltips.

src/dash-table/conditional/index.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export interface INamedElement {
+    column_id?: ColumnId;
+}
+
+export type ConditionalBasicFilter = INamedElement;
+export type ConditionalDataCell = IConditionalElement & IIndexedRowElement & INamedElement;
+export type ConditionalCell = INamedElement;
+export type ConditionalHeader = IIndexedHeaderElement & INamedElement;
```
Marc-Andre-Rivet: Regrouping conditional interfaces outside of the styling interfaces as they are now used in two places

src/dash-table/DataTable.js
```diff
@@ -537,6 +540,48 @@ export const propTypes = {
...
+        type: PropTypes.oneOf([
+            'text',
+            'markdown'
+        ]),
+        value: PropTypes.string
+    })),
+
+    column_conditional_tooltips: PropTypes.arrayOf(PropTypes.shape({
+        if: PropTypes.shape({
```
alexcjohnson: Even if it works, makes me a bit uneasy to use a keyword `if` as a key. Perhaps `condition`?

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
alexcjohnson: neat - perhaps we should use this in `dcc` as well!

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
```diff
@@ -0,0 +1,184 @@
...
+                style={{
+                    left: this.state.pageX,
+                    top: this.state.pageY
+                }}
+                dangerouslySetInnerHTML={{
+                    __html: typeof tooltip === 'string' ?
+                        tooltip :
+                        tooltip.type === TooltipSyntax.Markdown ?
+                            this.state.md.render(tooltip.value) :
+                            tooltip.value
```
alexcjohnson: For markdown yes, but do we need non-markdown tooltips to be raw HTML?

src/dash-table/DataTable.js
Marc-Andre-Rivet: So this follows the same pattern that's been used elsewhere (e.g. dropdown and style conditions) -- there was a pretty long discussion involving @cldougl and @chriddyp over the use of 'if' vs 'condition' vs other things I don't quite remember at the top of my head. General feeling was that the possible confusion was offset by the clarity and terseness.

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
Marc-Andre-Rivet: This is not raw html, this is raw text. We do not plan on supporting raw html input / parsing

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
Marc-Andre-Rivet: Yes! There will be a few things to iron out like dependency packaging and reuse.. wouldn't do to do all this and reload your library multiple times b/c they come from different packages!

Marc-Andre-Rivet:
Updated with proper documentation for the tooltips and a standalone version of the demo app that uses tooltips. When running `npm run build.watch`, use `http://localhost:8080/index.html?mode=tooltips`.

There are issues with tooltip alignment in certain scroll situations in this version. Looking to fix those.

src/dash-table/DataTable.js
cldougl: > General feeling was that the possible confusion was offset by the clarity and terseness.

yeah that was it as I remember:
didn't want to use `if` but the overall syntax didn't look user friendly/readable when using `conditional` or more specific descriptors

alexcjohnson: OK, if this is matching other usage in the ecosystem I can live with it :)

src/dash-table/components/ControlledTable/TooltipCursorDecorator.tsx
alexcjohnson: But you're providing it to the `dangerouslySetInnerHTML` prop - so won't it be rendered as HTML?

Marc-Andre-Rivet: Right.. changing that.

alexcjohnson:
Finally got a chance to play with the demo - works great @Marc-Andre-Rivet! My main questions now are about when and where exactly we want the tooltips to appear, and how configurable this should be.

**Where:** this one may be controversial, but to me a tooltip is labeling/explaining an object (in this case a table cell), not the mouse pointer; therefore it should be *attached* to the cell and not to the mouse pointer - specifically to an edge of the cell, so you can still see the whole cell content while the tooltip is displayed.

**When:** there's what feels to me a very long delay before the tooltip first appears. For some purposes this might be desirable. For example, editable cells where the primary purpose of the tooltip is to help you know how to edit it, most of the time you'll be quickly mousing over a cell and editing it, and you only want the tooltip if you're hesitating. But for other purposes, particularly read-only cells where the primary purpose of mousing over it is to get more information, it would be nice if the tooltip appeared quickly. I don't know which case will be more common or if it's necessary to allow this configurability right now, but I suspect at some point we'll want to support a variable delay. Would it be too confusing to have multiple delays within one table? I think so, even if both use cases are present in the same table.

Currently the tooltip disappears on any `mousemove` event, and the delay begins again before it reappears. If we anchored it to the cell rather than the pointer, I think it would make sense for the delay to begin on `mouseenter` and the tooltip to disappear only on `mouseleave`.

Right now there's some inconsistent behavior between moving the pointer and scrolling. If you move the pointer at all, the tooltip disappears, and it reappears after the pointer has been stationary for the delay time. But if you scroll without moving the pointer at all, the tooltip will stay, and if you scroll into a different cell, it immediately updates to the tooltip for the new cell. I suspect this would be fixed by using only `mouseenter` and `mouseleave`... I see you have an `onScroll` handler that looks like it's supposed to sort this out but it's not - perhaps it's only some parent element that's getting the scroll event?

Marc-Andre-Rivet:
Commit [above](https://github.com/plotly/dash-table/pull/338/commits/67222544d4a36d8e3ce1f188491e56063734671c) addresses:

- the long delay for the tooltip to become visible (500ms --> 350ms)
- links the tooltip visibility to `mouseenter` and `mouseleave` of cells
- positions the tooltip relative to the cell instead of the mouse

alexcjohnson:
> the long delay for the tooltip to become visible (500ms --> 350ms)

Definite improvement. I could still see people wanting an option to have it instantaneous in some cases, but that can be left out for now.

> links the tooltip visibility to `mouseenter` and `mouseleave` of cells -- improves behavior on scrolling as a side-effect

Beautiful 🎉 

> positions the tooltip relative to the cell instead of the mouse

Love it. Does need to shift left if the normal positioning rolls off the right side of the page, and I guess flip to above the cell if it's too far down?
screen shot 2019-01-23 at 5 00 45 pm
screen shot 2019-01-23 at 5 02 07 pm

> Will now concentrate on fixing positioning issues (first tooltip often shows up in top/left corner)

Nice, I noticed that issue too...

> and styling. Will look into not displaying the tooltip when the cell is in edit mode.

Also: currently if you either move the mouse down or scroll up such that your mouse is over the tooltip, it will disappear (good, I think, though I can see your point about wanting to be able to select things in the tooltip... not sure how to handle this) but you don't get a tooltip for the new cell you're over.

Marc-Andre-Rivet:
> Does need to shift left if the normal positioning rolls off the right side of the page, and I guess flip to above the cell if it's too far down?

Right. Side-effect of the implementation change is that the positioning logic has been wiped out :)

> currently if you either move the mouse down or scroll up such that your mouse is over the tooltip, it will disappear

Noticed that too. But good to have it in the thread!

With a small delay for the tooltip to disappear, the user should have time to move onto the tooltip before it disappears (to bridge the gap between exiting the cell and entering the tooltip). Once there the mouseenter-mouseleave of the tooltip would take precedence over the cell's mouse events.

Marc-Andre-Rivet:
With the latest commits the positioning is improved to try and respect the viewport's limits both horizontally and vertically, taking into consideration the body's scrolling. Scrolling behavior is improved and scrolling through an existing tooltip doesn't prevent the next cell's tooltip from showing up -- using onMouseOver for this is there is neither a mouse move or enter when scrolling without the bounds of a cell.. mouseOver does not suffer from such limitations.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -831,7 +831,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         }, column_conditional_tooltips);
 
         return conditionalTooltips.length ?
-            conditionalTooltips[0] :
+            conditionalTooltips.slice(-1)[0] :
```
Marc-Andre-Rivet: Inverting the conditional tooltip precedence, the last one that fits the criteria wins.

Marc-Andre-Rivet:
Inverted precedence of tooltips. Dropdowns were already behaving as desired
https://github.com/plotly/dash-table/blob/79e8d7ca7c6ce047c6a700c7d44dbaa571994295/src/dash-table/derived/cell/dropdowns.ts#L42

.config/webpack/base.js
```diff
@@ -19,6 +19,7 @@ module.exports = (preprocessor = {}, mode = 'development') => {
             path: path.resolve(__dirname, `./../../${dashLibraryName}`),
             filename: '[name].js',
             library: dashLibraryName,
+            publicPath: '/_dash-component-suites/dash_table/',
```
Marc-Andre-Rivet: Dynamically loaded files expect to be in the same path as the requester -- this is not true in Dash. I hate breaking this assumption -- creates a deep (and mostly unknown) dependency between Dash BE logic and the component build.

Marc-Andre-Rivet:
> behavior while editing
I'm also wondering as to the tooltip behavior to adopt when the cell is being edited -- do we want to show the tooltip when the user has already drilled down into a single cell for editing (vs. just active) -- I would think not.

Going with something simple as a starting point. The tooltip will display itself for 1.5s before disappearing. This value is rather arbitrary and taken from the `material design` tooltip behavior. As for the delay, this is not currently customizable.

Marc-Andre-Rivet:
@alexcjohnson I think this is ready for another look feature wise -- still missing tests.

alexcjohnson:
@Marc-Andre-Rivet looking great! Positioning works perfectly AFAICT, logic looks great.

> Going with something simple as a starting point. The tooltip will display itself for 1.5s before disappearing.

I think it's going to be important to have this configurable - in some cases there is likely to be a ton of info in the tooltip, and users will want it to display indefinitely.

You haven't said much about styling - and I think that's fine, as long as users know how to override the defaults with their own CSS. But one thing we may want to help with is adding an arrow - personally I generally prefer tooltips with arrows, but I know that's not universal. Anyway if we make arrows they're probably going to need some JS to position & flip from top to bottom. I don't know whether they can be sized (including turning them off) with just CSS or if that would need to be a prop.

.config/webpack/base.js
alexcjohnson: (for completeness: lazy loading was removed in https://github.com/plotly/dash-table/pull/338/commits/cd277ed66dfd96ac1b774aaa825222b2126cf6ee pending discussion in https://github.com/plotly/dash-core/issues/39)

Marc-Andre-Rivet:
Items identified as missing:

- [ ] Unit tests for tooltip logic
- [ ] Dash test with tooltip display
- [ ] Configurable tooltip display time
- [ ] Tooltip styling (directional arrow)

Marc-Andre-Rivet:
> I think it's going to be important to have this configurable - in some cases there is likely to be a ton of info in the tooltip, and users will want it to display indefinitely.

Wondering what your thoughts are on this one. Having a globally configurable tooltip delay/duration does not cater well to the `ton of info` tooltip that would coexist with smaller, more standard ones.

On the other end of the spectrum, could make each static and conditional tooltip setting have an extra & optional delay/duration nested prop -- and default back to defaults otherwise.

I don't really like either solution. Maybe use both.. and check if two negatives make a positive in this case!

alexcjohnson:
Having both a global default and per-item overrides is a pattern we use all over plotly.js - so I'm certainly comfortable with that. As long as the per-item override docs specify that the global one will be used as the default, users should be sufficiently alerted to the (easier) global, avoiding unintentionally surprising experiences, and only giving surprising experiences that the user explicitly intended :)

src/dash-table/DataTable.js
```diff
@@ -537,6 +542,151 @@ export const propTypes = {
...
+    /**
+     * `column_static_tooltip` represents the tooltip shown
+     * for different columns.
+     * The `property` name refers to the column ID.
+     * The `type` refers to the type of tooltip syntax used
+     * for the tooltip generation. Can either be `markdown`
+     * or `text`. Defaults to `text`.
+     * The `value` refers to the syntax-based content of
+     * the tooltip. This value is required.
+     * The `delay` represents the delay in milliseconds before
```
cldougl: it might be helpful to have the delay and duration defaults in this doc string

src/dash-table/DataTable.js
```diff
@@ -537,6 +542,151 @@ export const propTypes = {
...
+     * the tooltip is shown when hovering a cell.
+     *
+     * Default to 350.
+     */
+    tooltip_delay: PropTypes.number,
+
+    /**
+     * `tooltip_duration` represents the table-wide duration in milliseconds
+     * during which the tooltip will be displayed when hovering a cell.
+     */
```
cldougl: also would be helpful to see the duration default as the delay default is included above.

src/dash-table/DataTable.js
```diff
@@ -537,6 +542,151 @@ export const propTypes = {
...
+    /**
+     * `tooltip_delay` represents the table-wide delay in milliseconds before
+     * the tooltip is shown when hovering a cell.
+     *
+     * Default to 350.
+     */
+    tooltip_delay: PropTypes.number,
+
+    /**
+     * `tooltip_duration` represents the table-wide duration in milliseconds
```
cldougl: After trying out the tooltip demo the duration seems a bit to short imo. *I might be mentally comparing this too strongly to plotly.js hover behaviour but I think increasing the default would make the tooltips a bit more user friendly.

Marc-Andre-Rivet:
@cldougl I've added support for `null` on both delay and duration on the table's default and on the tooltips nested props.

`null` delay is equivalent to passing `0`
`null` duration is equivalent to passing `Infinity` (not really.. but 50 days in milliseconds is close enough...)

The table's `null` will not be overriden by tooltip-specific configurations.

@alexcjohnson Added the anchors / arrow to the tooltip -- this was tricky as regenerating the whole table just to get the arrow alignment right would have been silly. The arrow can be up or down and will react to scrolling -- if the arrow would be pointing to the next/previous cell by scrolling enough... the arrow disappears and displays as before. Styling pseudo element can't be done in a purely JS way and would have required a dynamic stylesheet -- again, seems a bit silly for an edge case. Maybe later :) Will leave more comments in the PR to explain the "hacky" implementation and the reasoning.

src/dash-table/DataTable.js
```diff
@@ -537,6 +542,161 @@ export const propTypes = {
...
+     * for different columns.
+     * The `property` name refers to the column ID.
+     * The `type` refers to the type of tooltip syntax used
+     * for the tooltip generation. Can either be `markdown`
+     * or `text`. Defaults to `text`.
+     * The `value` refers to the syntax-based content of
+     * the tooltip. This value is required.
+     * The `delay` represents the delay in milliseconds before
+     * the tooltip is shown when hovering a cell. This overrides
+     * the table's `tooltip_delay` property. If set to `null`,
```
Marc-Andre-Rivet: Added reference to the table's default tooltip delay + `null` explanation

Marc-Andre-Rivet: Same for duration below

src/dash-table/DataTable.js
Marc-Andre-Rivet: Bumped up the default value to 2000 from 1500

src/dash-table/DataTable.js
Marc-Andre-Rivet: Done :)

src/dash-table/components/ControlledTable/fragments/TableTooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+import tooltipHelper from 'dash-table/components/tooltipHelper';
+import ReactDOM from 'react-dom';
+import { isEqual } from 'core/comparer';
+
+interface IState {
+    arrow?: Arrow;
+    cell?: any;
+}
+
+export default class TableTooltip extends PureComponent<ITooltipProps, IState> {
```
Marc-Andre-Rivet: This component is *not* standalone and is meant to be used in the context of the controlled table (hence the /ControlledTable/fragments/... location) -- in order for the table not to have to keep track of the scroll's exact position for positioning the tooltip & re-rendering uselessly, exposing a `updateBounds` method below that the CTable will call when needed (scroll, mouse over, etc.) -- this is a little against React's philosophy but keeps all the info / state in the child, making it possible to re-render the tooltip only on non state affecting user operations at the table's level.

src/dash-table/components/ControlledTable/fragments/TableTooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    render() {
+        const { arrow } = this.state;
+
+        return (<Tooltip
+            key='tooltip'
+            ref='tooltip'
+            arrow={arrow}
+            {...this.props}
+        />);
```
Marc-Andre-Rivet: Displays a regular tooltip

src/dash-table/components/ControlledTable/fragments/TableTooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        el.style.top = `${top}px`;
+        el.style.left = `${left}px`;
+        el.style.position = 'absolute';
+
+        if (this.state.arrow !== arrow) {
+            this.setState({ arrow });
+        }
+    }
```
Marc-Andre-Rivet: On component update, readjust the tooltip's position and arrow type as required.

src/dash-table/components/ControlledTable/fragments/TableTooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+            return;
+        }
+
+        const positionalBounds = positionalParent.getBoundingClientRect();
+        const parentBounds = parent.getBoundingClientRect();
+
+        const { clientWidth: elWidth, clientHeight: elHeight } = el;
+        const elAnchorHeight = Math.max(
+            parseFloat(getComputedStyle(el, ':before').borderWidth || '0'),
+            parseFloat(getComputedStyle(el, ':after').borderWidth || '0')
```
Marc-Andre-Rivet: The height of the element is not enough.. need to know how much gets added by the arrow -- luckily while we can't style the pseudo-elements from JS, we can still access their existing style.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -760,4 +789,23 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+
+        const id = tooltip.id;
+        const row = tooltip.row - virtualized.offset.rows;
+
+        const { r1c1, tooltip: t } = this.refs as { [key: string]: any };
+
+        if (t) {
+            const cell = r1c1.querySelector(`td[data-dash-column="${id}"][data-dash-row="${row}"]`);
+
+            (this.refs.tooltip as TableTooltip).updateBounds(cell);
```
Marc-Andre-Rivet: Get the element being hovered and ask the tooltip fragment to do its thing.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -727,12 +738,30 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+    /* Tooltip */
+        let tableTooltip = derivedTooltips(
+            tooltip,
+            tooltips,
+            column_conditional_tooltips,
+            column_static_tooltip,
+            virtualized,
+            tooltip_delay,
+            tooltip_duration
+        );
```
Marc-Andre-Rivet: Memoized tooltip object with resolved values -- makes it easier to manage the tooltip's lifecycle than with plain value props.

src/dash-table/components/Table/Table.less
```diff
@@ -211,6 +211,76 @@
...
+
+    .dash-table-tooltip {
+        position: relative;
+        background-color: #f6f6f6;
+        max-width: 300px;
+        min-width: 300px;
+        padding: 2px 10px;
+    }
+
+    &[data-attr-anchor='top'] {
```
Marc-Andre-Rivet: Styling specific to each arrow type

src/dash-table/components/Table/props.ts
```diff
@@ -199,6 +219,12 @@ interface IProps {
 
     id: string;
 
+    tooltips?: ITableTooltips;
+    tooltip_delay: number | null;
+    tooltip_duration: number | null;
```
Marc-Andre-Rivet: Can now be null!

src/dash-table/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+interface ITooltipState {
+    display?: boolean;
+    displayTooltipId?: any;
+    hideTooltipId?: any;
+    md: Remarkable;
+}
+
+export default class Tooltip extends PureComponent<ITooltipProps, ITooltipState> {
+    constructor(props: ITooltipProps) {
```
Marc-Andre-Rivet: This is the actual, non-decorated, non-fragment tooltip implementation -- should work everywhere / be reusable.

src/dash-table/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    componentWillReceiveProps(nextProps: ITooltipProps) {
+        const { delay, duration } = nextProps.tooltip;
+
+        if (isEqual(
+            R.omit(['arrow'], this.props),
+            R.omit(['arrow'], nextProps)
+        )) {
+            return;
+        }
```
Marc-Andre-Rivet: If we're scrolling the only that can change is the arrow -- don't listen to it for resetting the timeouts

src/dash-table/components/tooltipHelper.ts
```diff
@@ -0,0 +1,39 @@
...
+    let relativeParent = el;
+    while (getComputedStyle(relativeParent).position !== 'relative') {
+        if (!relativeParent.parentElement) {
+            break;
+        }
+
+        relativeParent = relativeParent.parentElement;
+    }
+
+    return { positionalParent, parent };
```
Marc-Andre-Rivet: Finds out who is the relative parent and the real parent of the tooltip -- necessary to position it absolutely correctly.

src/dash-table/derived/cell/eventHandler.ts
```diff
@@ -33,6 +36,9 @@ const getter = (propsFn: () => ICellFactoryProps): CacheFn => {
         [Handler.Change, handleChange.bind(undefined, propsFn)],
         [Handler.Click, handleClick.bind(undefined, propsFn)],
         [Handler.DoubleClick, handleDoubleClick.bind(undefined, propsFn)],
+        [Handler.Enter, handleEnter.bind(undefined, propsFn)],
+        [Handler.Leave, handleLeave.bind(undefined, propsFn)],
+        [Handler.Move, handleMove.bind(undefined, propsFn)],
```
Marc-Andre-Rivet: New events necessary for handling the various tooltip scenarios.

Marc-Andre-Rivet:
> But it doesn't seem to work when the tooltip flips to the top 

Weird. Checking it out!

Marc-Andre-Rivet:
Hahaha! Ordering the logic correctly is key!

```
        if (delta > parent.clientWidth / 2) {
            arrow = undefined;
        }

        // too low
        if (top + elHeight > vwTop + vwHeight) {
            top = (parentBounds.top - positionalBounds.top) + positionalParent.scrollTop - (elHeight + elAnchorHeight);
            arrow = Arrow.Bottom; <----------- COME ON!!!
        }
```

PR #343, opened 2019-01-23
Issue 342- Add column type support for conditionals

Marc-Andre-Rivet:
As part of epic #166 and follow up to type support #297.

Adds type support for conditionals.


Marc-Andre-Rivet:
As mentioned by @alexcjohnson, style_data_conditional is missing `filter` in the prop types

Marc-Andre-Rivet:
I'll also use the occasion to add a few more visual tests for the existing conditionals.

One thing that that might be of interest is the style resolution logic that's going on behind the scene. For example, a data cell will apply *all* matching static and conditional styles, overriding previously set values in this order

static cell styles --> conditional cell styles --> static data cell styles --> conditional data cell styles

Styles defined at a higher index in the conditional arrays have precedence over the styles defined before.
So if for some given data cell the follow apply:

static_cell ```{ backgroundColor: 'red' }```
conditional cell ```{ backgroundColor: 'blue', color: 'white' }```
static data ```{ color: 'black' }```
conditional data ```[{ backgroundColor: 'yellow', color: 'red', fontFamily: 'cursive' }, { backgroundColor: 'orange', fontFamily: 'serif' }]```

The final result will be ```{ backgroundColor: 'orange', color: 'red', fontFamily: 'serif' }```

The same logic applies to header and filter cells and is located in:

- https://github.com/plotly/dash-table/blob/6bb5215e7d4402a0eb7e811db3ff91f6aa93559d/src/dash-table/derived/style/index.ts#L65 (notice the nomenclature is incorrect -- might as well fix this now)
- https://github.com/plotly/dash-table/blob/master/src/dash-table/derived/cell/wrapperStyles.ts (notice again the wishy-washy cell instead of data nomenclature)

The style merging priority is the opposite of the priority when selecting dropdowns and tooltips -- maybe we can discuss if it would make sense to invert the behavior of those props so as to behave like the styles (e.g. last defined as priority instead of first defined). Both behavior are fine in isolation but the combination of two priority schemes that essentially contradict each other isn't exactly great at the API level.

Marc-Andre-Rivet:
@alexcjohnson I think this is ready to be looked at again. Since last update:
- confirmed or modified dropdown and tooltip selection to choose the last one defined in the list, matching the priority for styles
- added visual tests
- added `filter` nested prop to the prop types of `style_data_conditional` -- no additional change required as the TS code already supported it

andreaallegra:
hello, 
which are the current possibilities of filter string in dropdowns header of the column_conditional_dropdowns?
Thank you

Issue #347, opened 2019-01-28
[Tech Debt] Clean up the typing for memoized / derived values


Marc-Andre-Rivet:
Better leverage TS type inference -- and clean up the interfaces. They are harder to understand than they should be and makes modifying these calcs harder / more error prone.

PR #349, opened 2019-01-29
Issue 347- Improve TS type inference & simplify caching logic

Marc-Andre-Rivet:
The table uses various caches for limit expensive calculations during render.

These calculations and caches are typed in a very explicit way, making them hard to understand and very unwieldy especially if there are multiple layers of caching involved like in the dropdowns generation. This is in part due to my limited knowledge of deeper TS typing, which is getting better. 

This is a first attempt at #347 -- there are many more cached/memoized calculations using another implementation of the memoization that need improvement but this should be representative of the overall direction -- for inference, less explicit stuff, less duplicated params.


src/core/cache/memoizer.ts
```diff
@@ -0,0 +1,21 @@
+import { memoizeOne } from 'core/memoizer';
+import { CacheKeyFragment, getCache } from '.';
+
+export default <TKey extends CacheKeyFragment[]>() => {
```
Marc-Andre-Rivet: Using two main types of caches in the table so far:

- Memoizer caches take in a key to get access to a memoized function that takes a different set of parameters. The key may remain the same but the params change. (e.g. evaluating the dropdown for each cell)
- Value caches take in a key and executes a function with the key parameters if the value is no present in the cache. The key and the params are always equal. (e.g. evaluating/binding the event handlers for each cell)

src/dash-table/components/CellFactory.tsx
```diff
@@ -23,7 +23,7 @@ export default class CellFactory {
         private readonly cellDropdowns = derivedDropdowns(),
         private readonly cellOperations = derivedCellOperations(),
         private readonly cellStyles = derivedCellStyles(),
-        private readonly cellWrappers = derivedCellWrappers(propsFn().id),
+        private readonly cellWrappers = derivedCellWrappers(),
```
Marc-Andre-Rivet: Cleaning up usage -- needed to call a fn into a fn for normal usage.

src/dash-table/derived/cell/dropdowns.ts
```diff
@@ -1,94 +1,128 @@
...
-                    (cd, i) => [cd, i],
-                    conditionalDropdowns
-                ))
-        )
-    ];
+export default () => new Dropdowns().get;
 
-    return dropdowns.length ? dropdowns.slice(-1)[0] : undefined;
-};
+class Dropdowns {
```
Marc-Andre-Rivet: Generate the dropdown options values for each cell in the table.

src/dash-table/derived/cell/dropdowns.ts
```diff
@@ -1,94 +1,128 @@
...
-    columnStaticDropdown: IColumnDropdown[],
-    dropdown_properties: IDropdownProperties
-): (DropdownValues | undefined)[][] => mapData((datum, rowIndex) => R.map(column => {
-    const realIndex = indices[rowIndex];
+        return this.dropdown.get(column.id, rowIndex)(
+            applicable,
+            column,
+            datum
+        );
+    }, columns), data));
```
Marc-Andre-Rivet: First level -- in many case the table get regenerated because the active cell has changed -- this does not affect the dropdowns -- only regenerate if the table structure, data or dropdown options have changed.

src/dash-table/derived/cell/dropdowns.ts
```diff
@@ -1,94 +1,128 @@
...
+        const conditional = columnConditionalDropdown.find((cs: any) => cs.id === column.id);
+        const base = columnStaticDropdown.find((ss: any) => ss.id === column.id);
 
-    const conditionalDropdowns = (conditional && conditional.dropdowns) || [];
-    const staticDropdown = legacyDropdown || (base && base.dropdown);
+        return [
+            legacyDropdown || (base && base.dropdown),
+            (conditional && conditional.dropdowns) || []
+        ];
+    });
```
Marc-Andre-Rivet: Per cell caching -- this could probably be dropped but would require the use of a constant `[]` to not trigger re-renderers on React's side.

Marc-Andre-Rivet: Eval / cache the applicable dropdowns for each cell

src/dash-table/derived/cell/dropdowns.ts
```diff
@@ -1,94 +1,128 @@
...
-    indices: Indices,
-    columnConditionalDropdown: any,
-    columnStaticDropdown: any,
-    dropdown_properties: any
-) => (DropdownValues | undefined)[][]) => {
-    const astCache = memoizerCache<[ColumnId, number], [string], SyntaxTree>(
-        (query: string) => new SyntaxTree(query)
-    );
+        return matches.length ? matches.slice(-1)[0] : undefined;
+    });
```
Marc-Andre-Rivet: Identify the dropdown to use for a given cell -- this evaluates the filter condition through the SyntaxTree and is fairly expensive, hence the caching.

src/dash-table/derived/cell/dropdowns.ts
```diff
@@ -1,94 +1,128 @@
...
+    ) => new SyntaxTree(query));
 
-export default memoizeOneFactory(decoratedGetter);
+    /**
+     * Evaluate if the query matches the cell's data.
+     */
+    private readonly evaluation = memoizerCache<[ColumnId, number]>()((
+        ast: SyntaxTree,
+        datum: Datum
+    ) => ast.evaluate(datum));
```
Marc-Andre-Rivet: Caching both the SyntaxTree generated from the filter condition and the result of a given tree evaluation. As per above, this is for each cell.

src/dash-table/derived/cell/eventHandler.ts
```diff
@@ -18,34 +16,36 @@ export enum Handler {
...
+        columnIndex: number,
+        rowIndex: number
+    ) => {
+        let handlerFn = this.handlers.get(handler);
 
-        return cache(args, handlers.get(handler) as HandlerFn, rowIndex, columnIndex);
-    };
-};
+        return handlerFn && handlerFn.bind(undefined, rowIndex, columnIndex);
+    });
```
Marc-Andre-Rivet: Here the value and the key are generated with the same params and the value is good for the duration of the life of the table. No memoization, just a regular value cache.

src/dash-table/derived/cell/eventHandlerProps.ts
```diff
@@ -30,11 +30,11 @@ const getter = (propsFn: () => ICellFactoryProps): CacheFn => {
...
-            onChange: derivedHandlers(Handler.Change, rowIndex, columnIndex),
-            onClick: derivedHandlers(Handler.Click, rowIndex, columnIndex),
-            onDoubleClick: derivedHandlers(Handler.DoubleClick, rowIndex, columnIndex),
-            onMouseUp: derivedHandlers(Handler.MouseUp, rowIndex, columnIndex),
-            onPaste: derivedHandlers(Handler.Paste, rowIndex, columnIndex)
+            onChange: derivedHandlers(Handler.Change, columnIndex, rowIndex),
+            onClick: derivedHandlers(Handler.Click, columnIndex, rowIndex),
+            onDoubleClick: derivedHandlers(Handler.DoubleClick, columnIndex, rowIndex),
+            onMouseUp: derivedHandlers(Handler.MouseUp, columnIndex, rowIndex),
+            onPaste: derivedHandlers(Handler.Paste, columnIndex, rowIndex)
```
Marc-Andre-Rivet: Inverted params order to better match what was done elsewhere.
This and the above eventHandler will probably be merged in the future -- it just impacts a little too many things for now.

src/dash-table/derived/cell/wrappers.tsx
```diff
@@ -1,31 +1,26 @@
...
+    /**
+     * Returns the wrapper for each cell in the table.
+     */
+    get = memoizeOne((
+        activeCell: ActiveCell,
+        columns: VisibleColumns,
+        data: Data,
+        offset: IViewportOffset,
+        selectedCells: SelectedCells
+    ) => R.addIndex<Datum, JSX.Element[]>(R.map)(
```
Marc-Andre-Rivet: All the wrappers for the table, impacted by the data, structure/pagination/virtualization and user interactions.

src/dash-table/derived/cell/wrappers.tsx
```diff
@@ -38,33 +33,29 @@ function getter(
...
+                    columnIndex,
+                    column.id
+                );
+            }, columns
+        ), data));
+
+    /**
+     * Returns the wrapper for a cell.
+     */
+    private wrapper = memoizerCache<[number, number]>()((
```
Marc-Andre-Rivet: a cell by cell memoizer driven cache of the wrapper

src/core/cache/value.ts
```diff
@@ -0,0 +1,19 @@
+import { CacheKeyFragment, getCache } from '.';
+
+export default <TKey extends CacheKeyFragment[]>() =>
+    <TEntry>(fn: (...a: TKey) => TEntry) => {
```
Marc-Andre-Rivet: Turns out TS is great at inferring types when there's only one generic, and not so good when there are 2 or more.. in this case it makes sense to have a first call to access the cache entry and a second to execute the memoized function so it works out well enough.

src/core/cache/value.ts
```diff
@@ -0,0 +1,19 @@
...
+        const cache = new Map<CacheKeyFragment, any>();
+
+        function get(...key: TKey) {
+            const lastKey = key.slice(-1)[0];
+
+            const nestedCache = getCache(cache, ...key);
+
+            return (
+                nestedCache.get(lastKey) ||
```
alexcjohnson: What if `fn(...key)` is falsy? That can't happen in `memoizer` (returns a function) or `getCache` (returns a `Map`) but here it could, right?

alexcjohnson: Ah, I guess the only way you're currently using the value cache is to store functions so not a problem in practice - still, in case it ever finds a use with primitives it might be nice to protect against that case.

Marc-Andre-Rivet: Actually this is a good point -- I use this to store the result of the ast's evaluation for the dropdowns -- a `false` eval will reevaluate -- always.

Marc-Andre-Rivet: > I guess the only way you're currently using the value cache is to store functions so not a problem in practice

No. In the value cache I store the actual value returned by the function, not the decorated function as in the memoizer cache.

Marc-Andre-Rivet: So the memoizer cache is fine but the value cache is incorrect. 🥇

PR #359, opened 2019-02-04
Issue 224 - Improve column id support for filtering

### LINKED ISSUE #224: Filtering doesn't work if column name has space [Sponsored: Due March 1]
Filtering on columns that contain a space in their name results in the filter string being considered as invalid.

After looking at the docs, it most likely is that the filter parser is being confused when turning the filter string into `<column name> <operator> <value>`

Temporary fix is, of course, removing spaces from column names.

Marc-Andre-Rivet:
Fixes #224.

Many normal usage scenarios are calling for column ids with non-word [^a-zA-Z0-9_] characters.
This PR adds the needed support for a subset of special characters [-+.:] without quotes, and for the full set of characters when single or double quoted.

Adds a few unit tests for the syntactic tree and a standalone/integration test for real-world usage validation.

Tests are passing against dash/dash-renderer [latest release](https://github.com/plotly/dash-table/pull/359/commits/1782dc41332a1dfbf3b52176a9495d8d10783571)


demo/AppMode.ts
```diff
@@ -202,6 +212,8 @@ function getState() {
             return getFixedVirtualizedState();
         case AppMode.ReadOnly:
             return getReadonlyState();
+        case AppMode.ColumnsInSpace:
+            return getSpaceInColumn();
```
Marc-Andre-Rivet: A new mock/mode with specially named columns

src/core/syntax-tree/lexicon.ts
```diff
@@ -31,10 +31,21 @@ const isPrime = (c: number) => {
...
+        if (/^('.*')|(".*")$/.test(tree.value)) {
+            return target[
+                tree.value.slice(1, tree.value.length - 1)
+            ];
+        } else if (/^(\w|[:.\-+])+$/.test(tree.value)) {
+            return target[tree.value];
+        }
+    },
+    regexp: /^('([^()']|\\')+'|"([^()"]|\\")+"|(\w|[:.\-+])+)/
+};
```
Marc-Andre-Rivet: Separate operand and expression logic -- support arbitrary quoted string or limited character set without quotes.

src/dash-table/components/FilterFactory.tsx
```diff
@@ -55,7 +55,7 @@ export default class FilterFactory {
         }
 
         setFilter(R.map(
-            ([cId, filter]) => `${cId} ${filter}`,
+            ([cId, filter]) => `"${cId}" ${filter}`,
```
Marc-Andre-Rivet: Auto-quote the column id when generating the filter fragments from column filters

tests/cypress/src/DashTable.ts
```diff
@@ -7,6 +7,14 @@ export default class DashTable {
         return cy.get(`#table tbody tr td[data-dash-column="${column}"]`).eq(row);
     }
 
+    static getFilter(column: number) {
+        return cy.get(`#table tbody tr th.dash-filter.column-${column}`);
+    }
+
+    static getFilterById(column: number | string) {
+        return cy.get(`#table tbody tr th.dash-filter[data-dash-column="${column}"]`);
+    }
```
Marc-Andre-Rivet: New selectors for getting the filter cells

tests/cypress/tests/standalone/filtering_test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+        DashTable.getFilterById('g.gg').click();
+        DOM.focused.type(`gt num(1000)${Key.Enter}`);
+
+        DashTable.getFilterById('b+bb').click();
+        DOM.focused.type(`eq "Wet"${Key.Enter}`);
+
+        DashTable.getCellById(0, 'rows').within(() => cy.get('.dash-cell-value').should('have.html', '101'));
+        DashTable.getCellById(1, 'rows').should('not.exist');
+    });
```
Marc-Andre-Rivet: test that the filters are working fine for various special cases

tests/cypress/tests/unit/syntactic_tree_test.ts
```diff
@@ -1,10 +1,92 @@
...
+        it('support single quoted column name with " " (space)', () => {
+            const tree = new SyntaxTree('\'a dot\' eq "1 dot" || \'a dot\' eq "2 dot"');
+
+            expect(tree.isValid).to.equal(true);
+            expect(tree.evaluate(data0)).to.equal(false);
+            expect(tree.evaluate(data1)).to.equal(true);
+            expect(tree.evaluate(data2)).to.equal(true);
+            expect(tree.evaluate(data3)).to.equal(false);
+        });
+    });
```
Marc-Andre-Rivet: Unit tests for the various special cases

src/core/syntax-tree/lexicon.ts
alexcjohnson: `[^()']` - why are `()` excluded?

tests/cypress/tests/unit/syntactic_tree_test.ts
alexcjohnson: Can you include some nested quotes?

src/core/syntax-tree/lexicon.ts
Marc-Andre-Rivet: No good reason. Removing the exclusion.

src/dash-table/DataTable.js
```diff
@@ -71,7 +71,7 @@ export const defaultProps = {
     column_conditional_dropdowns: [],
     column_static_dropdown: [],
 
-    column_static_tooltip: [],
+    column_static_tooltip: {},
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: As discussed, documented this in https://github.com/plotly/plotly.js/issues/420

tests/cypress/tests/unit/syntactic_tree_test.ts
Marc-Andre-Rivet: Added

src/core/syntax-tree/lexicon.ts
```diff
@@ -33,22 +33,22 @@ const isPrime = (c: number) => {
 
 const operand = {
     resolve: (target: any, tree: ISyntaxTree) => {
-        if (/^('.*')|(".*")$/.test(tree.value)) {
+        if (/^(('([^'\\]|\\'|\\)+')|("([^"\\]|\\"|\\)+")|(`([^`\\]|\\`|\\)+`))$/.test(tree.value)) {
```
alexcjohnson: Is `'([^'\\]|\\'|\\)+'` different from `'([^']|\\')+'`?

alexcjohnson: oh actually, it's important to NOT match `'blah\'` (missing the real closing quote) which seems like both of these would match?
So perhaps `'([^'\\]|\\.)+'`?

Marc-Andre-Rivet: You're right. Will add some tests for it too.

PR #361, opened 2019-02-05
Fix table tests

Marc-Andre-Rivet:
`dash` pulling other `dash-*` installs prevents the table tests from running correctly as the latest production build is used instead of the branch test build.

Correct test run here: https://circleci.com/gh/plotly/dash-table/tree/pip-local-packages

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,9 +33,9 @@ jobs:
                 name: Install requirements
                 command: |
                     . venv/bin/activate
+                    pip install -e .
                     pip install -r requirements-base.txt --quiet
                     pip install -r requirements-v0.txt --quiet
-                    pip install -e .
 
             - run:
                 name: Run build:js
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -35,6 +35,7 @@ jobs:
                     . venv/bin/activate
                     pip install -r requirements-base.txt --quiet
                     pip install -r requirements-v0.txt --quiet
+                    pip install -e .
```
alexcjohnson: Might be better to put this before the `install -r` items? As it is seems like `install -r` will grab `dash-table@3.3.0` each time, then `install -e` will put back the local version... same result but should be faster if we tell it up front we're using local `dash-table`.

Marc-Andre-Rivet: Sure. If it doesn't get overriden, makes sense!

PR #371, opened 2019-02-12
Caching clean up

Marc-Andre-Rivet:
A continuation of the previous cleanup tasks.. in this case, simplifying the cell events mapping and caching by merging together two very tightly coupled calcs / functions in charge of caching and maintaining a pool of event handlers for the cells.


src/dash-table/derived/cell/contents.tsx
```diff
@@ -48,69 +45,76 @@ function getCellType(
...
-        },
-        columns
-    ),
-    data
-);
+                const className = [
+                    ...(active ? ['input-active'] : []),
+                    ...(isFocused ? ['focused'] : ['unfocused']),
+                    ...['dash-cell-value']
+                ].join(' ');
```
alexcjohnson: Non-blocking: that'll work, and you're just moving this code around... but seems a bit convoluted. Couldn't it be simplified to:
```js
const className = [
    ...(active ? ['input-active'] : []),
    isFocused ? 'focused' : 'unfocused',
    'dash-cell-value'
].join(' ');
```

Marc-Andre-Rivet: sure

PR #372, opened 2019-02-12
Issue 370 - Clearing `filtering_settings` clears the by-column filter UI

### LINKED ISSUE #370: Clearing `filtering_settings` does not clear the by-column filter fields
From community [thread](https://community.plot.ly/t/clearing-filters-in-a-datatable/19381/2).

Clearing the `filtering_settings` does not cause the by-column filter fields to be cleared. The table displays the right data but the fields themselves are showing the previous value.

Expected resolution:
- fixes the behavior
- adds at least one test demonstrating that resetting the `filtering_settings` prop causes the filter fields to be cleared in the UI

community-19381

```
import pandas as pd
import numpy as np

from datetime import date

import dash
from dash.dependencies import Input, Output, State

app = dash.Dash(__name__)
app.css.config.serve_locally = True
app.scripts.config.serve_locally = True

sColNames = ['a' + str(x) for x in range(date.today().year-6, date.today().year+1)]
df =  pd.DataFrame(data = np.array([[1.5,0.0,-1.5]]*7).T, index=range(3), columns=sColNames) # Example dataframe

app.layout = html.Div(children=[
    html.Button(id='clear', children='Clear filter'),
    DataTable(
        id='table',
        columns=[{'name': i, 'id': i} for i in df.columns],
        data=df.to_dict('rows'),
        filtering=True
    )
])

@app.callback(
    Output('table', 'filtering_settings'),
    [Input('clear', 'n_clicks')],
    [State('table', 'filtering_settings')],
)
def clearFilter(n_clicks, state):
    if n_clicks is None:
        return '' if state is None else state

    return ''


if __name__ == "__main__":
    app.run_server(debug=True)
```

Marc-Andre-Rivet:
Fixes #370.
Relates to https://community.plot.ly/t/clearing-filters-in-a-datatable/19381/2

Two problems here:
- the UI was not picking up the updated value for the column filters
- the syntax tree was evaluating empty queries as invalid -- preventing them from being processed further (this is done to prevent the UI from updating with invalid values when typing in a new / potentially invalid query fragment for a column) -- small change of logic here


demo/App.js
```diff
@@ -28,11 +29,30 @@ class App extends Component {
...
+                className='clear-filters'
+                onClick={() => {
+                    const tableProps = R.clone(this.state.tableProps);
+                    tableProps.filtering_settings = '';
+
+                    this.setState({ tableProps });
+                }}
+            >Clear Filter</button>);
+        }
+    }
```
Marc-Andre-Rivet: For this new case, need an additional button to trigger the filter reset -- either that or make this a full on e2e test, which I think is pointless.

Marc-Andre-Rivet: Mode based conditional additional comps

demo/App.js
```diff
@@ -28,11 +29,30 @@ class App extends Component {
...
+
     render() {
-        return (<DataTable
-            setProps={this.setProps}
-            {...this.state.tableProps}
-        />);
+        return (<div>
+            {this.renderMode()}
```
Marc-Andre-Rivet: Conditionals added with the table

demo/AppMode.ts
```diff
@@ -31,6 +32,11 @@ export const ReadWriteModes = [
     AppMode.Virtualized
 ];
 
+export const BasicModes = [
+    ...ReadWriteModes,
+    AppMode.ReadOnly
+];
```
Marc-Andre-Rivet: Each time we added a mode we were running the standalone tests against an additional mode. In most cases, we want a new mode to test specific aspects of the table not all cases.

src/core/components/IsolatedInput/index.tsx
```diff
@@ -38,6 +38,16 @@ export default class IsolatedInput extends PureComponent<IProps, IState> {
...
+    componentWillReceiveProps(nextProps: IProps) {
+        const { value: nextValue } = nextProps;
+
+        if (this.state.value !== nextValue) {
+            this.setState({
+                value: nextValue
+            });
+        }
+    }
+
```
Marc-Andre-Rivet: The input needs to update its internal state when the value/prop updates. That prevented the UI from updating to match the filter.

src/core/syntax-tree/index.ts
```diff
@@ -18,18 +18,16 @@ export default class SyntaxTree {
...
         }
 
-        const evaluate = this.tree.lexeme.evaluate;
-
-        return evaluate ?
-            evaluate(target, this.tree) :
-            false;
+        return this.tree && this.tree.lexeme && this.tree.lexeme.evaluate ?
+            this.tree.lexeme.evaluate(target, this.tree) :
+            true;
```
Marc-Andre-Rivet: A valid null tree (no query string) now evaluates to true (the target matches the criteria)

src/core/syntax-tree/syntaxer.ts
```diff
@@ -64,6 +64,10 @@ export default (lexerResult: ILexerResult): ISyntaxerResult => {
         return { valid: false, error: `lexer -- ${lexerResult.error}` };
     }
 
+    if (lexerResult.lexemes.length === 0) {
+        return { valid: true };
+    }
```
Marc-Andre-Rivet: Having no query string is now a valid / null tree case instead of an invalid case

src/dash-table/components/Filter/Column.tsx
```diff
@@ -27,16 +23,6 @@ export default class ColumnFilter extends PureComponent<IColumnFilterProps, ICol
...
-    componentWillReceiveProps(nextProps: IColumnFilterProps) {
-        const { value: nextValue } = nextProps;
-
-        if (this.state.value !== nextValue) {
-            this.setState({
-                value: nextValue
-            });
-        }
-    }
-
```
Marc-Andre-Rivet: This was dead code -- dating back to before the IsolatedInput was added - nothing depends on the state here, removing.

src/dash-table/components/FilterFactory.tsx
```diff
@@ -130,8 +130,12 @@ export default class FilterFactory {
         }
 
         const { tree } = syntaxerResult;
-        const toCheck: (ISyntaxTree | undefined)[] = [tree];
+        if (!tree) {
+            this.ops.clear();
+            return;
+        }
```
Marc-Andre-Rivet: When rebuilding the fragments, clear all fragments if there's no query

tests/cypress/tests/standalone/filtering_test.ts
```diff
@@ -5,31 +5,71 @@ import Key from 'cypress/Key';
...
+
+            DashTable.getFilterById('g.gg').click();
+            DOM.focused.type(`gt num(1000)${Key.Enter}`);
+
+            DashTable.getFilterById('b+bb').click();
+            DOM.focused.type(`eq "Wet"${Key.Enter}`);
+
+            DashTable.getCellById(0, 'rows').within(() => cy.get('.dash-cell-value').should('have.html', '101'));
+            DashTable.getCellById(1, 'rows').should('not.exist');
+        });
```
Marc-Andre-Rivet: Weird diff here but nothing changed

tests/cypress/tests/standalone/filtering_test.ts
```diff
@@ -5,31 +5,71 @@ import Key from 'cypress/Key';
...
 
-        DashTable.getFilterById('b+bb').click();
-        DOM.focused.type(`eq "Wet"${Key.Enter}`);
+            cy.get('.clear-filters').click();
 
-        DashTable.getCellById(0, 'rows').within(() => cy.get('.dash-cell-value').should('have.html', '101'));
-        DashTable.getCellById(1, 'rows').should('not.exist');
+            DashTable.getCellById(0, 'ccc').within(() => cy.get('.dash-cell-value').should('have.html', cell_0));
+            DashTable.getCellById(1, 'ccc').within(() => cy.get('.dash-cell-value').should('have.html', cell_1));
+        });
```
Marc-Andre-Rivet: Apply a filter column by column, see that it's applied, clear it, see that it's not applied anymore!

Marc-Andre-Rivet: And make sure the fields are empty -- as per described issue

Marc-Andre-Rivet:
Need to do changelog!

PR #377, opened 2019-02-15
Issue 189 - Number formatting

### LINKED ISSUE #189: data types - decimal formatting options for numerical columns

Marc-Andre-Rivet:
This is a work in progress and not quite ready for review 😄 

Fixes #189 

- [x] general formatting of numbers through [d3-format](https://github.com/d3/d3-format)
- [x] [locale format](https://github.com/d3/d3-format#locale_format)
- [x] [si prefix](https://github.com/d3/d3-format#locale_formatPrefix)
- [x] [four digits](https://github.com/plotly/plotly.js/blob/b03bb3b01a851e201ebf98341144ca11a414378d/src/plots/cartesian/layout_attributes.js#L520-L528) grouping case
**separate_4digits** nested prop under `locale` will prevent grouping on values ]-10000,+10000[
- [x] python helper methods
- [x] proper documentation of the new props / nested props
- [ ] proper documentation / IDE integration (?) or discoverability of Python helpers
- [x] push BI logic (column format defaults, etc.) higher in the stack (ideally in/around datatable.js)
**BI logic** here means (1) data validation and (2) data sanitizing -- more data sanitizing/validation can be added in the future, for now concerning itself with columns (new) & pagination/filtering/sorting (refactored)
- [x] general / sanity unit testing of the formatter functionality
- [x] integration testing of formatting / table content
- [x] testing for Python wrappers
- [x] changelog entry
- [x] basic performance verifications
- [x] toPlotlyJson for Format objects
- [x] rename Python-facing `currency` to `symbol`
- [x] sibling PR for `dash-docs` (https://github.com/plotly/dash-docs/pull/416)

Edge cases identified while writing tests or documentation
- [x] handling of coerce/validation/passthrough/isNully cases during formatting -- a number column may very well not contain a number, how to configure and handle these cases from a formatting perspective?
**Approach**: overridable `nully` behavior, non-number types are passthroughs
- [x] Clipboard copy/paste behavior is ok when data comes from outside but does not deal well with null/undefined cases as everything gets transformed to a string (`''` in those cases) and information is lost
**Approach** Keep a duplicate of the df slice copied -- if during paste the df slice's stringification is the same as the clipboard content, assume that the data came from a datatable.
- [ ] determine desired `nully` case behavior (currently `''` (empty string)

Note: `Diff settings` -> `Hide whitespace` helps with a few of the diffs that involve indentation changes..


dash_table/Format.py
```diff
@@ -0,0 +1,200 @@
...
+    PERCENTAGE = '%'
+    PERCENTAGE_ROUNDED = 'p'
+    BINARY = 'b'
+    OCTAL = 'o'
+    DECIMAL_INTEGER = 'd'
+    LOWER_CASE_HEX = 'x'
+    UPPER_CASE_HEX = 'X'
+    UNICODE = 'c'
+
+class Format():
```
Marc-Andre-Rivet: A fuent interface to create the desired format/locale/prefix configuration

dash_table/Format.py
```diff
@@ -0,0 +1,200 @@
...
+            self.specifier['align'].value,
+            self.specifier['sign'].value,
+            self.specifier['symbol'].value,
+            self.specifier['padding'].value,
+            self.specifier['width'],
+            self.specifier['group'].value,
+            self.specifier['precision'],
+            self.specifier['trim'].value,
+            self.specifier['type'].value
+        )
```
Marc-Andre-Rivet: Returns the actual format configuration object corresponding to all previous operations

dash_table/FormatTemplate.py
```diff
@@ -0,0 +1,21 @@
...
+    .create()
+
+
+def percentage(decimals: int = 0):
+    decimals = decimals if isinstance(decimals, int) else 0
+
+    return Format() \
+    .precision(decimals) \
+    .type(Type.PERCENTAGE) \
+    .create()
```
Marc-Andre-Rivet: Two basic helper functions for money and percentages, with very limited configurability

src/dash-table/DataTable.js
```diff
@@ -24,23 +25,64 @@ export default class DataTable extends Component {
...
+            Logger.error(`Invalid combination of filtering / sorting / pagination`);
+            return (<div>Invalid props combination</div>);
+        }
+
+        if (!this.allValidColumns()) {
+            Logger.error(`Invalid column format`);
+            return (<div>Invalid props combination</div>);
+        }
+
+        return this.props.id ? (<RealTable {...this.props} />) : (<RealTable {...this.props} id={this.getId()} />);
```
Marc-Andre-Rivet: Due to the way we use prop-types, function-based validations are not available to us -- making sure the column format props involving arrays contain the expected number of items -- if not, the table is deemed invalid

src/dash-table/DataTable.js
```diff
@@ -157,6 +199,44 @@ export const propTypes = {
...
+                currency: PropTypes.arrayOf(PropTypes.string),
+                decimal: PropTypes.string,
+                grouping: PropTypes.arrayOf(PropTypes.number),
+                numerals: PropTypes.arrayOf(PropTypes.string),
+                percent: PropTypes.string,
+                thousands: PropTypes.string
+            }),
+            prefix: PropTypes.number,
+            specifier: PropTypes.string
+        }),
```
Marc-Andre-Rivet: Per-column format configuration -- combines together d3-format's locale, prefix and specifier.

src/dash-table/DataTable.js
```diff
@@ -275,9 +355,32 @@ export const propTypes = {
...
+     *   'thousands': the string used for the groups separator
+     */
+    locale_format: PropTypes.shape({
+        currency: PropTypes.arrayOf(PropTypes.string),
+        decimal: PropTypes.string,
+        grouping: PropTypes.arrayOf(PropTypes.number),
+        numerals: PropTypes.arrayOf(PropTypes.string),
+        percent: PropTypes.string,
+        thousands: PropTypes.string
+    }),
```
Marc-Andre-Rivet: table-wide locale configuration, overriden/merged by the columns

src/dash-table/derived/cell/contents.tsx
```diff
@@ -58,63 +60,69 @@ class Contents {
...
         editable: boolean,
         isFocused: boolean,
         dropdowns: (DropdownValues | undefined)[][]
-    ): JSX.Element[][] => mapData(
-        (datum, rowIndex) => mapRow(
-            (column, columnIndex) => {
-                const active = isActiveCell(activeCell, rowIndex + offset.rows, columnIndex + offset.columns);
+    ): JSX.Element[][] => {
+        const localizedFormatter = getFormatter(defaultLocale);
+        const formatters = R.map(localizedFormatter, columns);
```
Marc-Andre-Rivet: Get the formatters for all the columns from the start -- d3's formatting functions generation is not expensive but over thousands of cells it starts to have an impact.

src/dash-table/derived/cell/contents.tsx
```diff
@@ -58,63 +60,69 @@ class Contents {
...
+                                value={datum[column.id]}
+                            />);
+                        case CellType.Label:
+                        default:
+                            return (<CellLabel
+                                className={className}
+                                key={`column-${columnIndex}`}
+                                onClick={this.handlers(Handler.Click, rowIndex, columnIndex)}
+                                onDoubleClick={this.handlers(Handler.DoubleClick, rowIndex, columnIndex)}
+                                value={formatters[columnIndex](datum[column.id])}
```
Marc-Andre-Rivet: When the cell is not selected, format the content as required.

src/dash-table/handlers/cellEvents.ts
```diff
@@ -1,7 +1,7 @@
 import * as R from 'ramda';
 import { SelectedCells, ICellFactoryProps } from 'dash-table/components/Table/props';
 import isActive from 'dash-table/derived/cell/isActive';
-import reconcile from 'dash-table/reconcile';
+import reconcile from 'dash-table/type/reconcile';
```
Marc-Andre-Rivet: Restructured the /reconcile (coerce, validate) folder to a more general /type folder that contains a variety of operations on the types (now coerce, validate, format)

src/dash-table/type/formatter.ts
```diff
@@ -0,0 +1,20 @@
...
+    (c: IColumnType) => {
+        let formatter;
+        switch (c.type) {
+            case ColumnType.Numeric:
+                formatter = getNumberFormatter(locale, c);
+                break;
+        }
+
+        return formatter || DEFAULT_FORMATTER;
+    };
```
Marc-Andre-Rivet: formatter resolver for the columns, if no formatter is found, use one that is a value-passthrough

src/dash-table/type/number.ts
```diff
@@ -0,0 +1,39 @@
...
+        R.merge(
+            defaultLocale,
+            R.omit(['specifier', 'prefix'], options.format)
+        )
+    );
+
+    return options.format.prefix ?
+        locale.formatPrefix(options.format.specifier, options.format.prefix) :
+        locale.format(options.format.specifier);
+}
```
Marc-Andre-Rivet: Only getFormatter is new -- the diff is weird because of the file rename.

Marc-Andre-Rivet: This returns a formatter function that accepts a value (number)

Marc-Andre-Rivet:
@alexcjohnson Not quite ready yet (no tests, bad python integration, etc.) but the core functionality is there and having a first look before fine tuning would make most sense.

Will add a test mode for formatting.

dash_table/Format.py
```diff
@@ -0,0 +1,200 @@
+from enum import Enum
```
Marc-Andre-Rivet: Def looking for input on the Python portions!
Here, mostly trying to hide away complexity through an interface that exposes the different aspects independently and through enums instead of the actual d3-format values.

The hope is to create an abstraction that would survive a switch from d3-format to some other library without having to rewrite the user facing contract.

dash_table/Format.py
```diff
@@ -0,0 +1,200 @@
...
+            'padding': Padding.NO,
+            'precision': '',
+            'sign': Sign.UNDEFINED,
+            'symbol': Symbol.UNDEFINED,
+            'trim': Trim.NO,
+            'type': Type.UNDEFINED
+        }
+
+    # Specifier
+    def align(self, align: Align = Align.UNDEFINED):
```
alexcjohnson: I'm afraid we can't use type annotations and support py2.7 If you really want to, there's a comment-based convention https://mypy.readthedocs.io/en/latest/python2.html

Also, if you're already filling in the default below it seems redundant here. Unless you wanted this to function as a setter/getter, in which case you could use `align=None` and `if(align is None): return self.specifier['align']`

dash_table/FormatTemplate.py
```diff
@@ -0,0 +1,21 @@
...
+
+def money(decimals: int = 2, sign: Sign = Sign.UNDEFINED):
+    decimals = decimals if isinstance(decimals, int) else 2
+
+    return Format() \
+    .precision(decimals) \
+    .sign(sign) \
+    .symbol(Symbol.CURRENCY) \
+    .type(Type.FIXED_POINT) \
+    .create()
```
alexcjohnson: `\` is generally to be avoided - I suppose if we add constructor kwargs that'll do it here, but using this same syntax you can just wrap the whole expression in parentheses and the `\` will be unneeded.

src/dash-table/DataTable.js
Marc-Andre-Rivet: Added a nully case that can take in an arbitrary value -- will use this value instead and will format it like a normal value if it's of the right type.

src/dash-table/type/number.ts
```diff
@@ -0,0 +1,65 @@
...
+    return (value: any) => {
+        if (isNully(value)) {
+            return typeof nully === 'number' ?
+                numberFormatter(nully) :
+                nully;
+        } else if (typeof value === 'number') {
+            return numberFormatter(value);
+        } else {
+            return value;
+        }
```
Marc-Andre-Rivet: Wraps d3-format logic to handle incorrect types and null cases

tests/cypress/tests/unit/formatting_test.ts
```diff
@@ -0,0 +1,129 @@
+import { getFormatter } from 'dash-table/type/number';
+
+describe('formatting', () => {
+    describe('number', () => {
+        it('returns undefined formatter with undefined format', () => {
+            const formatter = getFormatter(undefined, undefined);
+            assert.isNotOk(formatter);
+        });
+
+        describe('without nully handling / default locale', () => {
```
Marc-Andre-Rivet: A few unit tests testing mostly the wrapper behavior -- `locale + specifier + prefix + nully` with a few configurations -- as most of the functionality actually comes from d3-format

tests/cypress/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
     "extends": "./../../tsconfig.base.json",
     "compilerOptions": {
-        "baseUrl": "./../..",
+        "baseUrl": ".",
```
Marc-Andre-Rivet: This was causing warnings in the IDE -- now pointing to the correct base folder to resolve the imports location

dash_table/Format.py
```diff
@@ -0,0 +1,270 @@
...
+    return collections.namedtuple(name, dict.keys())(*dict.values())
+
+
+Align = get_named_tuple('align', {
+    'default': '',
+    'left': '<',
+    'right': '>',
+    'center': '^',
+    "right_sign": '='
+})
```
Marc-Andre-Rivet: Not sure if this is better than the Enum approach but it does seem more flexible / lightweight from the user's perspective

dash_table/Format.py
```diff
@@ -0,0 +1,270 @@
...
+            'type': Scheme.default
+        }
+
+        valid_methods = [m for m in dir(self) if m != '__init__' and m != 'create' and inspect.ismethod(getattr(self, m))]
+
+        for arg in kwargs:
+            if arg not in valid_methods:
+                raise Exception('{0} is not a format method. Expected one of'.format(arg), str(list(valid_methods)))
+
+            getattr(self, arg)(kwargs[arg])
```
Marc-Andre-Rivet: Stumbled my way to this solution for applying the values to the underlying data structure in the instance

package.json
```diff
@@ -19,6 +19,8 @@
     "private::host_dash8082": "python tests/cypress/dash/v_copy_paste.py",
     "private::host_dash8083": "python tests/cypress/dash/v_fe_page.py",
     "private::host_js": "http-server ./dash_table -c-1 --silent",
+    "private::lint:ts": "tslint '{src,demo,tests}/**/*.{js,ts,tsx}' --exclude '**/@Types/*.*'",
+    "private::lint:py": "flake8 --exclude=DataTable.py,__init__.py,_imports_.py --ignore=E501,F401,F841,F811 dash_table",
```
Marc-Andre-Rivet: Added linting for py -- ignores copied over from the dcc configuration. Not running it against the generated files.

src/dash-table/DataTable.js
```diff
@@ -6,14 +7,8 @@ import RealTable from 'dash-table/components/Table';
...
-
-function isFrontEnd(value) {
-    return ['fe', true, false].indexOf(value) !== -1;
-}
-
-function isBackEnd(value) {
-    return ['be', false].indexOf(value) !== -1;
-}
+import isValidProps from './DataTable.validate';
+import sanitizeProps from './DataTable.sanitize';
```
Marc-Andre-Rivet: Moved the top-level table data validation & editing to helpers..

src/dash-table/DataTable.js
```diff
@@ -157,6 +145,49 @@ export const propTypes = {
...
+         */
+        format: PropTypes.shape({
+            locale: PropTypes.shape({
+                currency: PropTypes.arrayOf(PropTypes.string),
+                decimal: PropTypes.string,
+                group: PropTypes.string,
+                grouping: PropTypes.arrayOf(PropTypes.number),
+                numerals: PropTypes.arrayOf(PropTypes.string),
+                percent: PropTypes.string,
+                separate_4digits: PropTypes.bool
```
Marc-Andre-Rivet: The 4 digits case is toggled here

src/dash-table/components/CellInput/index.tsx
```diff
@@ -47,6 +47,11 @@ export default class CellInput extends PureComponent<ICellProps, ICellState> {
             value
         } = this.props;
 
+        // input does not handle `null` correct (causes console error)
+        const sanitizedValue = this.state.value === null ?
```
Marc-Andre-Rivet: null is a valid value in the table but the input does not handle it well -- kind of annoying that JSON supports what the other does not

src/dash-table/type/number.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const locale = formatLocale(convertToD3(format.locale));
+
+    const numberFormatter = format.prefix ?
+        locale.formatPrefix(format.specifier, format.prefix) :
+        locale.format(format.specifier);
+
+    const thousandsSpecifier = format.locale.separate_4digits ?
+        format.specifier :
+        format.specifier.replace(/,/, '');
```
Marc-Andre-Rivet: If the number is ]-10000,+10000[ remove the `,` from the formatting string

src/dash-table/type/reconcile.ts
```diff
@@ -63,16 +63,14 @@ function doFailureRecovery(result: IReconciliation, c: IColumnType) {
...
 
-    // If Skip/Prevent
-    if (failure !== ChangeFailure.Default) {
-        return result;
+    if (failure === ChangeFailure.Default) {
+        const defaultValue = (c && c.validation && c.validation.default) || null;
+        result.success = true;
+        result.value = defaultValue;
+    } else if (failure === ChangeFailure.Accept) {
+        result.success = true;
```
Marc-Andre-Rivet: The logic here was incorrect and based off the previous Skip/Prevent implementation -- the new tests uncovered the case as it was not properly tested.

src/dash-table/utils/TableClipboardHelper.ts
```diff
@@ -21,6 +23,7 @@ export default class TableClipboardHelper {
         );
 
         const value = SheetClip.prototype.stringify(df);
+        TableClipboardHelper.lastLocalCopy = df;
```
Marc-Andre-Rivet: Turns out that SheetClip saves undefined|null as a empty string. Which is probably fine when copying to/from Excel but causing unexpected behavior when copying from a table to another (in the same context).

Here, keeping a copy of the table's df slice used for the copy and comparing its stringify result to the value in the clipboard on paste. If the values are identical, making the assumption that the copy/paste was initiated with table data.. reusing the df slice instead of the parse result, keeping the undefined|null values..

It's a bit hacky..

src/dash-table/utils/applyClipboardToData.ts
```diff
@@ -3,11 +3,11 @@ import * as R from 'ramda';
 import Logger from 'core/Logger';
 
 import { ActiveCell, Columns, Data, ColumnType } from 'dash-table/components/Table/props';
-import reconcile from 'dash-table/reconcile';
+import reconcile from 'dash-table/type/reconcile';
 import isEditable from 'dash-table/derived/cell/isEditable';
 
 export default (
-    values: string[][],
+    values: any[][],
```
Marc-Andre-Rivet: More general type to handle the table's copy/paste scenario above

.circleci/config.yml
```diff
@@ -154,9 +158,17 @@ jobs:
...
+                command: |
+                    . venv/bin/activate
+                    pip install -r requirements-base.txt --quiet
+
             - run:
                 name: Run eslint
-                command: npm run lint
+                command: |
+                    . venv/bin/activate
+                    npm run lint
```
Marc-Andre-Rivet: Add some flake linting

Marc-Andre-Rivet:
@alexcjohnson This is ready for another round I think. JS should be fairly complete. Still missing tooling for Python probably & implementation may or may not make total sense, and dash-docs to be started / linked!

dash_table/Format.py
alexcjohnson: Looks good to me! It has all the right behaviors AFAICT and seems easier for us to work with (`value not in Align`) than a straight `class Align(object)`.

dash_table/Format.py
alexcjohnson: Sure, that'll work. Style point, since `args` is typically used for positional arguments, I'd avoid `arg` as the name. Perhaps:
```py
for kw, val in kwargs.items():
    ...
    getattr(self, kw)(val)
```

dash_table/Format.py
```diff
@@ -0,0 +1,270 @@
...
+            'width': '',
+            'padding': Padding.no,
+            'precision': '',
+            'sign': Sign.default,
+            'symbol': Symbol.none,
+            'trim': Trim.no,
+            'type': Scheme.default
+        }
+
+        valid_methods = [m for m in dir(self) if m != '__init__' and m != 'create' and inspect.ismethod(getattr(self, m))]
```
alexcjohnson: potentially simpler to use the class instead, so you don't have to filter out the instance attrs added above?
```py
[m for m in dir(self.__class__) if m[0] != '_' and m != 'create']
```

dash_table/Format.py
```diff
@@ -0,0 +1,270 @@
...
+    # Specifier
+    def align(self, value):
+        if value not in Align:
+            raise Exception('expected value to be one of', str(list(Align)))
+
+        self.specifier['align'] = value
+        return self
+
+    def fill(self, value):
+        self.specifier['fill'] = value
```
alexcjohnson: do we need to check that this is a length-1 string?

dash_table/Format.py
Marc-Andre-Rivet: Ah! Didn't find that one when I looked at the docs, updating.

dash_table/Format.py
Marc-Andre-Rivet: Agreed. Also, forgot about Python supporting iterating on two lists at a time!.. although this is destructuring a tuple, not two lists..

dash_table/Format.py
Marc-Andre-Rivet: Yes, updating

dash_table/Format.py
```diff
@@ -0,0 +1,286 @@
...
+def is_string_py2(value):
+    return isinstance(value, str) or isinstance(value, unicode)
+
+
+def is_string_py3(value):
+    return isinstance(value, str)
+
+
+def is_string(value):
+    return is_string_py3(value) if sys.version_info >= (3, 0) else is_string_py2(value)
```
Marc-Andre-Rivet: dash-docs was having issues in py27 because unicode is a thing.. choosing the function based on the situation and disabling F821 in package.json (py3 doesn't like seeing unicode lying around!)

Marc-Andre-Rivet:
@alexcjohnson Ready for yet another go at this. The dash-docs PR is opened and linked in the summary. My main question remaining for you is if we need to make the Python code more discoverable and how.

dash_table/Format.py
alexcjohnson: I generally think it's cleaner to do feature detection than switch based on the version number. This one has some nice options https://stackoverflow.com/questions/11301138

This one is particularly terse: `isinstance(x, ("".__class__, u"".__class__))`

Marc-Andre-Rivet: Sure. Will actually go with isinstance(x, (str, u''.__class__)) as there's no edge case for string

dash_table/Format.py
```diff
@@ -0,0 +1,286 @@
...
+
+        self._prefix = value
+        return self
+
+    def to_plotly_json(self):
+        f = {}
+        f['locale'] = self._locale.copy()
+        f['nully'] = self._nully
+        f['prefix'] = self._prefix
+        f['specifier'] = '{0}{1}{2}{3}{4}{5}{6}{7}{8}{9}'.format(
```
alexcjohnson: minor - you can omit the numbers here... `'{}{}{}{}{}{}{}{}{}{}'`

It's annoying that `fill` has extra logic, without that you could do
`'{fill}{align}{sign}...'.format(**self._specifier)`

I suppose it would work to do something like:
```py
'{fill_if}{align}{sign}...'.format(
    fill_if=self._specifier['fill'] if self._specifier['align'] != Align.default else '',
    **self._specifier
)
```

dash_table/Format.py
```diff
@@ -0,0 +1,286 @@
...
+            raise Exception('expected value to be a string')
+
+        if 'symbol' not in self._locale:
+            self._locale['symbol'] = ['', value]
+        else:
+            self._locale['symbol'][1] = value
+
+        return self
+
+    def decimal_delimitor(self, value):
```
alexcjohnson: `decimal_delimiter` (and `group_delimiter` below)

dash_table/Format.py
```diff
@@ -0,0 +1,286 @@
...
+            'symbol': Symbol.no,
+            'trim': Trim.no,
+            'type': Scheme.default
+        }
+
+        valid_methods = [m for m in dir(self.__class__) if m[0] != '_' and m != 'to_plotly_json']
+
+        for kw, val in kwargs.items():
+            if kw not in valid_methods:
+                raise Exception('{0} is not a format method. Expected one of'.format(kw), str(list(valid_methods)))
```
alexcjohnson: We should try to be more specific than `Exception`. Seems like the right one here may be `TypeError`, as that's what you get from a function (or constructor) that you call with the wrong args:
```py
>>> def f(a=5):
...     return a+5
... 
>>> f(b=6)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: f() got an unexpected keyword argument 'b'
```

dash_table/Format.py
```diff
@@ -0,0 +1,286 @@
...
+
+            getattr(self, kw)(val)
+
+    # Specifier
+    def align(self, value):
+        if value not in Align:
+            raise Exception('expected value to be one of', str(list(Align)))
```
alexcjohnson: 🌴 can these be pulled out, and called with something like `self._validate(value, Align)`?
Also the exception type - possibly `TypeError` again, unless we want `TypeError` for non-strings and `ValueError` for invalid strings - but that distinction is probably not worth adding extra code for https://docs.python.org/3/library/exceptions.html#TypeError
> Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

dash_table/FormatTemplate.py
```diff
@@ -0,0 +1,23 @@
...
+        precision=decimals,
+        scheme=Scheme.fixed,
+        sign=sign,
+        symbol=Symbol.yes
+    )
+
+
+def percentage(decimals, rounded=False):
+    if not isinstance(rounded, bool):
+        raise Exception('expected rounded to be a boolean')
```
alexcjohnson: TypeError

src/dash-table/DataTable.js
```diff
@@ -157,6 +145,49 @@ export const propTypes = {
...
+         *
+         * This prop is derived from the [d3-format](https://github.com/d3/d3-format) library specification. Apart from
+         * being structured slightly differently (under a single prop), the usage
+         * is the same.
+         *
+         * 'locale': represents localization specific formatting information
+         *   When left unspecified, will use the default value provided by d3-format.
+         *
+         *   'symbol': (default: ['', $']) a list of two strings representing the
+         *   prefix and suffix symbols
```
alexcjohnson: default `['$', '']`

OK, I'm comfortable generalizing currency to symbol here - but let's describe what's happening, both to elucidate the common use case and to not confuse d3 aficionados. Perhaps:

a list of two strings representing the prefix and suffix symbols. Typically used for currency, and implemented using d3's currency format, but you can use this for other symbols such as measurement units.

tests/cypress/tests/unit/formatting_test.ts
```diff
@@ -0,0 +1,153 @@
...
+                expect(formatter(0)).to.equal('$0.00');
+                expect(formatter(1)).to.equal('$1.00');
+                expect(formatter(-1)).to.equal('-$1.00');
+                expect(formatter(1.23)).to.equal('$1.23');
+                expect(formatter(1.232)).to.equal('$1.23');
+                expect(formatter(1.239)).to.equal('$1.24');
+                expect(formatter(1766)).to.equal('$1766.00');
+                expect(formatter(''), 'Empty string case').to.equal('');
+                expect(formatter('foo'), 'Foo string case').to.equal('foo');
+                expect(formatter(true), 'Empty string case').to.equal(true);
```
alexcjohnson: not an empty string 🍝

tests/cypress/tests/unit/formatting_test.ts
```diff
@@ -0,0 +1,153 @@
...
+                expect(formatter(0.13)).to.equal('0.13');
+                expect(formatter(0.131)).to.equal('0.13');
+                expect(formatter(1.23)).to.equal('1.2');
+                expect(formatter(1.299)).to.equal('1.3');
+                expect(formatter(1299)).to.equal('1,300');
+                expect(formatter(1299431)).to.equal('1,300,000');
+                expect(formatter(''), 'Empty string case').to.equal('');
+                expect(formatter('foo'), 'Foo string case').to.equal('foo');
+                expect(formatter(true), 'Empty string case').to.equal(true);
+                expect(formatter(NaN), 'NaN case').to.equal('42');
```
alexcjohnson: Ah interesting - if `nully` is a number it still has to pass through the formatter, that makes sense. Good call!

tests/cypress/tests/unit/formatting_test.ts
```diff
@@ -0,0 +1,153 @@
...
+                        separate_4digits: false
+                    }),
+                    nully: getNully('42.4242'),
+                    specifier: getSpecifier(',.2f')
+                });
+
+                assert.isOk(formatter);
+
+                expect(formatter(-1299)).to.equal('-1299.00');
+                expect(formatter(-1299431)).to.equal('-1,2,99,4,31.00');
```
alexcjohnson: huh, weird to use `separate_4digits` along with `grouping`, thanks for showing this case! Might extend more naturally as `separate_one_digit_more_than_last_group` 🤔 or even a number `min_digits_to_separate` but I guess absent a real user need we should just leave it as is.

tests/unit/format_test.py
```diff
@@ -0,0 +1,244 @@
...
+        self.assertEqual(res['specifier'], '.1%')
+
+    def test_valid_align_named(self):
+        Format().align(f.Align.center)
+
+    def test_valid_align_string(self):
+        Format().align('=')
+
+    def test_invalid_align_string(self):
+        self.assertRaises(Exception, Format().align, 'i')
```
alexcjohnson: Make sure to update these exception types to match the changed types in source. They'll pass with the base `Exception` but in principle the error could have come from some other problem.

src/dash-table/DataTable.sanitize.ts
```diff
@@ -0,0 +1,58 @@
+import * as R from 'ramda';
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: I would not normally do this but I've run into issues moving DataTable.js in the past -- normally I would do this:

DataTable
  |- index.js
  |- validate.ts
  |- sanitize.ts

It's been a long time and I certainly understand more about what's going on -- could give it another go.

tests/cypress/tests/unit/formatting_test.ts
Marc-Andre-Rivet: Yes, I think it's also weird, also considered something along the lines of `one_more_than` and slowly backed away.. without any real thoughts on how it was to behave, I wanted at least a test that would make sure it didn't change behavior without us knowing.

dash_table/Format.py
```diff
@@ -118,21 +106,43 @@ def __init__(self, **kwargs):
...
-                raise Exception('{0} is not a format method. Expected one of'.format(kw), str(list(valid_methods)))
+                raise TypeError('{0} is not a format method. Expected one of'.format(kw), str(list(valid_methods)))
 
             getattr(self, kw)(val)
 
+    def _validate_char(self, value):
+        self._validate_string(value)
+
+        if len(value) != 1:
+            raise ValueError('excepted value to a string of length one')
```
alexcjohnson: excepted->expected

dash_table/Format.py
```diff
@@ -118,21 +106,43 @@ def __init__(self, **kwargs):
...
+
+        if not isinstance(value, int):
+            raise TypeError('expected value to be an integer')
+
+        if value < 0:
+            raise ValueError('expected value to be non-negative', str(value))
+
+    def _validate_named(self, value, named_values):
+        if value not in named_values:
+            raise TypeError('expceted value to be one of', str(list(named_values)))
```
alexcjohnson: expceted 🙈

src/dash-table/DataTable.sanitize.ts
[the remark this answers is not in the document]
alexcjohnson: That would be fine... or leaving `DataTable.js` where it is and just giving these other ones other names without the extra dots - even `dataTableSanitize.ts` would make me more comfortable (I guess the leading capital also strikes me as off, since these don't define a class)

alexcjohnson:
> determine desired `nully` case behavior (currently `''` (empty string)

@chriddyp do you have an opinion about the default `nully` behavior? I like the empty string (as currently in this PR) but could be convinced to default to `'NA'` or something instead.

PR #394, opened 2019-03-04
Improve build - run against real latest

Marc-Andre-Rivet:


.circleci/config.yml
```diff
@@ -82,28 +83,22 @@ jobs:
...
-
             - run:
                 name: Install requirements
                 command: |
+                    sudo pip install --upgrade virtualenv
+                    python -m venv venv || virtualenv venv
                     . venv/bin/activate
-                    pip install -e .
                     pip install -r requirements-base.txt --quiet
                     pip install -r requirements-v1.txt --quiet
```
Marc-Andre-Rivet: release-v1 stays as is for now as the `build` scripts are not reworked

.circleci/config.yml
```diff
@@ -228,18 +239,22 @@ jobs:
...
-                    python3 -m venv venv
-
             - run:
                 name: Install requirements
                 command: |
+                    sudo pip install --upgrade virtualenv
+                    python -m venv venv || virtualenv venv
                     . venv/bin/activate
                     pip install -r requirements-base.txt --quiet
                     pip install -r requirements-v1.txt --quiet
```
Marc-Andre-Rivet: release-v1 stays as is for now as the `build` scripts are not reworked

PR #397, opened 2019-03-13
Improve Filtering

Marc-Andre-Rivet:
- isolate lexemes from lexicons and grammar/syntax
- support multiple lexicons (single column, multiple columns, full query)
- allow lexicons termination and validation logic for lexemes based on previous lexeme(s) (essentially a deterministic finite state-machine with these [state transitions](https://docs.google.com/spreadsheets/d/1tAK5jqJaRYxQC-gIi2L2-tUN-0TPrzclv12By2edWvg/edit?usp=sharing) -- copy of table below)

**Basics**
- [x] enable short filter criteria on column filters (e.g. `<binary op> <expression>`, `<unary op>`, `<expression>`)
- [x] update `operand` syntax to expect ```{field}``` instead of ```field``` or ```"field"```, [\{}] can be escaped with ```{\\{\\\\}field}```, see unit tests for details
- [x] additional unit tests on all lexicons to test syntax / edge cases
- [x] refactor multi column AST's transformation into and from single column AST
- [x] rework expression lexeme to be consistent with changes made to operand lexeme
- [x] increase lexicon's granularity (one lexeme per binary operation, etc.) -- will make handling the structured cases simpler
- [x] expose query as structure (lhs, rhs, pivot, block, etc.) on derived prop -- consumable programmatically

**After this PR**
- [x] `contains` operator
- [ ] `regex` operator
- [ ] type-aware/conditional lexemes (e.g. can't `contains` on `number`)
- [ ] type-aware default operator (single column query)
- [ ] support column types implicitly
- [ ] support for contextual conditionals (e.g. `if` in `style_**_conditional` -- allow expressing column_type, row_id, etc. as a filter instead of a structure)
- [ ] fix python unit tests -- the remaining py visual tests could (all?) probably be rewritten as Cypress / functionality tests
- [ ] suport `eq(...)` and `and(...)` syntax

Items with `x` are valid, `nest` refers to the level of nesting of the query up to that point -- each `block open` (open parentheses) increases the nesting by 1, each `block close` reduces the nesting by 1.

image


src/core/syntax-tree/index.ts
```diff
@@ -1,20 +1,29 @@
...
 import Logger from 'core/Logger';
-import lexer from 'core/syntax-tree/lexer';
+import lexer, { ILexerResult } from 'core/syntax-tree/lexer';
 import syntaxer, { ISyntaxerResult } from 'core/syntax-tree/syntaxer';
+import { Lexicon } from './lexicon';
 
 export default class SyntaxTree {
-    private result: ISyntaxerResult;
+    protected lexerResult: ILexerResult;
+    protected syntaxerResult: ISyntaxerResult;
```
Marc-Andre-Rivet: Keeping the intermediary results allows for finer manipulations in the derived syntax tree classes

tests/cypress/tests/unit/lexeme_test.ts
```diff
@@ -0,0 +1,111 @@
...
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: '{abc' } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: 'abc}' } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: '}abc{' } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: '{{abc}}' } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: '"' } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: '`' } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: `'` } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: '{' } as ISyntaxTree)).to.throw(Error);
+            expect(fieldExpression.resolve.bind(undefined, undefined, { value: '}' } as ISyntaxTree)).to.throw(Error);
```
alexcjohnson: Perhaps include a `target` here, so we know these errors aren't just from failed lookup. Also would be nice to demonstrate what happens with a real failed lookup, ie:
`fieldExpression.resolve({ abc: 3 }, { value: '{def}' } as ISyntaxTree)`

tests/cypress/tests/unit/lexeme_test.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+    });
+
+    it('resolves value expressions', () => {
+        expect(!!valueExpression.resolve).to.equal(true);
+        expect(typeof valueExpression.resolve).to.equal('function');
+
+        if (valueExpression.resolve) {
+            expect(valueExpression.resolve(undefined, { value: 'abc' } as ISyntaxTree)).to.equal('abc');
+            expect(valueExpression.resolve(undefined, { value: 'abc   ' } as ISyntaxTree)).to.equal('abc');
```
alexcjohnson: do we drop leading spaces as well? what happens if the target data contains leading/trailing spaces? Is this behavior different for `stringExpression`? Also a test with spaces mid-string in `stringExpression` would be nice, to show that you (presumably) don't need to escape them.

tests/cypress/tests/unit/lexeme_test.ts
```diff
@@ -0,0 +1,111 @@
...
+            expect(operand.resolve.bind(undefined, undefined, { value: 'abc' } as ISyntaxTree)).to.throw(Error);
+            expect(operand.resolve.bind(undefined, undefined, { value: '123' } as ISyntaxTree)).to.throw(Error);
+            expect(operand.resolve.bind(undefined, undefined, { value: '{abc' } as ISyntaxTree)).to.throw(Error);
+            expect(operand.resolve.bind(undefined, undefined, { value: 'abc}' } as ISyntaxTree)).to.throw(Error);
+            expect(operand.resolve.bind(undefined, undefined, { value: '{{abc}}' } as ISyntaxTree)).to.throw(Error);
+
+            expect(operand.resolve({ abc: 3 }, { value: '{abc}' } as ISyntaxTree)).to.equal(3);
+            expect(operand.resolve({ ['a bc']: 3 }, { value: '{a bc}' } as ISyntaxTree)).to.equal(3);
+            expect(operand.resolve({ ['{abc}']: 3 }, { value: '{\\{abc\\}}' } as ISyntaxTree)).to.equal(3);
+            expect(operand.resolve({ ['"abc"']: 3 }, { value: '{"abc"}' } as ISyntaxTree)).to.equal(3);
```
alexcjohnson: how is `operand` different from `fieldExpression`? (and why are the keys here in `[...]`?)

tests/cypress/tests/unit/query_syntactic_tree_test.ts
```diff
@@ -1,38 +1,38 @@
...
-            expect(new SyntaxTree('myField eq \\`value`').isValid).to.equal(false);
+            expect(new QuerySyntaxTree(`{myField} eq 'value'`).isValid).to.equal(true);
+            expect(new QuerySyntaxTree(`{myField} eq "value"`).isValid).to.equal(true);
+            expect(new QuerySyntaxTree('{myField} eq `value`').isValid).to.equal(true);
+            expect(new QuerySyntaxTree(`{myField} eq 'value\\'`).isValid).to.equal(true);
+            expect(new QuerySyntaxTree(`{myField} eq 'value\\''`).isValid).to.equal(true);
+            expect(new QuerySyntaxTree(`{myField} eq "value\\"`).isValid).to.equal(true);
+            expect(new QuerySyntaxTree('{myField} eq `value\\`').isValid).to.equal(true);
+            expect(new QuerySyntaxTree(`{myField} eq \\'value'`).isValid).to.equal(false);
+            expect(new QuerySyntaxTree(`{myField} eq \\"value"`).isValid).to.equal(false);
```
alexcjohnson: non-blocking style point - I find this kind of test easier to read (and more likely that more thorough cases will be added) if they're written as a loop (one for all the passes, another for all the fails):
```js
[
  `{myField} eq 'value'`,
  `{myField} eq "value"`,
  '{myField} eq `value`',
  `{myField} eq 'value\\'`,
  `{myField} eq 'value\\''`,
  `{myField} eq "value\\"`,
  '{myField} eq `value\\`'
].forEach(v => {
  expect(new QuerySyntaxTree(v).isValid).to.equal(true, v);
});
```
(assuming the extra msg arg `.equal(true, v)` works so you know which one failed)

tests/dash/app_styling.py
```diff
@@ -443,10 +443,10 @@ def layout():
                     "width": "100%"
                 },
                 style_data_conditional=[{
-                    "if": { "column_id": "Region", "filter": "Region eq str(Montreal)" },
+                    "if": { "column_id": "Region", "filter": "Region eq Montreal" },
```
alexcjohnson: I thought we needed to bracket `{Region}` now?

alexcjohnson:
This is looking really good! Aside from the comments above, the only test I think would be nice is putting extra whitespace (including characters like `\n`, `\r`, `\t`) in various places in the query (leading, trailing, multiple characters before & after operators and parens...) and showing that it still works.

tests/cypress/tests/unit/lexeme_test.ts
Marc-Andre-Rivet: Adding escaped leading/trailing spaces cases. The underlying assumption for the lexemes is that the lexer will take care of trimming the string prior to trying to find a match.

tests/cypress/tests/unit/lexeme_test.ts
Marc-Andre-Rivet: Adding

tests/cypress/tests/unit/lexeme_test.ts
Marc-Andre-Rivet: They differ only in terms of when they can be used in respect to their predecessors. Will clean up the implementation of the field expression a bit.

Marc-Andre-Rivet: `[...]` odd syntax for odd cases like property name generated through a function call

```
{ [myFunc()]: 'myValue' }
```

This is not really necessary here. Removing.

tests/dash/app_styling.py
Marc-Andre-Rivet: Updating.

Marc-Andre-Rivet:
Python/Selenium tests are failing. Not sure why. Investigating.

Marc-Andre-Rivet:
@alexcjohnson: Seems like the python visual tests got broken with the move to react 16 -- since the table was only built / never in a real PR, I never realized there were visual differences. The latest commits rewrite all the Python tests that can be written as storybook tests. The storybook tests have shown themselves very resilient to change / to be more stable. Some Python tests cannot be rewritten this way as they involve callbacks between the table and the graph. Technically these are not really visual tests, instead the visual result is an artefact indicating the dash app behaves correctly.. I hate those but there's no easy replacement for now. These seem to render just fine but the style is not being picked up when snapshot for whatever reason -- somehow seems related to the React 15 -> 16 update but that's no more than an hypothesis.

Still have to have look into `\r`, `\n` `\t` and the likes in the queries, fix issues to be found and add tests.

src/dash-table/syntax-tree/lexeme/expression.ts
```diff
@@ -0,0 +1,59 @@
+import isNumeric from 'fast-isnumeric';
+import * as R from 'ramda';
+
+import { LexemeType, IUnboundedLexeme } from 'core/syntax-tree/lexicon';
+import { ISyntaxTree } from 'core/syntax-tree/syntaxer';
+import operand from './operand';
+
+const STRING_REGEX = /^(('([^'\\]|\\'|\\)+')|("([^"\\]|\\"|\\)+")|(`([^`\\]|\\`|\\)+`))/;
```
alexcjohnson: hmm looking again at this regex - `('([^'\\]|\\'|\\)+')` will match `'stuff\'` ie anything starting with a simple quote and ending with an escaped quote. Perhaps the group should be `('([^'\\]|\\.)+')`? Similarly for the other quote types.

src/dash-table/syntax-tree/lexeme/expression.ts
```diff
@@ -0,0 +1,59 @@
...
+const VALUE_REGEX = /^(([^\s'"`{}()\\]|\\[\s'"`{}()]?)+)(?:[\s)]|$)/;
```
alexcjohnson: Similarly here, I'm having a harder time figuring out the consequences, but do we really want ```\\[\s'"`{}()]?```, which means the match could end with a `\`, or do we *always* include *any* character after `\` by just using `\\.`?

src/dash-table/syntax-tree/lexeme/logical.ts
```diff
@@ -0,0 +1,49 @@
...
+        Logger.trace('evaluate -> &&', target, tree);
+
+        const t = tree as any;
+        const lv = t.left.lexeme.evaluate(target, t.left);
+        const rv = t.right.lexeme.evaluate(target, t.right);
+        return lv && rv;
+    },
+    type: LexemeType.LogicalOperator,
+    priority: 2,
+    regexp: /^(and\s|&&)/i,
```
alexcjohnson: probably want to allow `(` immediately after, like `(1<2)and(3<4)` - that works in Python anyway, will confuse ppl if it fails. Same for `or`.

Marc-Andre-Rivet: This is allowed in full query form -- e.g. `{a} eq "1" and ({a} eq "0" or ({b} eq "1" or {b} eq "0"))`

Marc-Andre-Rivet: Ah! You mean the space?

src/dash-table/syntax-tree/lexeme/expression.ts
Marc-Andre-Rivet: As discussed, updating to force escaping `\` everywhere and allow escaping any character in all contexts.

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const contains: IUnboundedLexeme = R.merge({
+    evaluate: relationalEvaluator(([op, exp]) =>
+        typeof op === 'string' &&
+        typeof exp === 'string' &&
+        op.indexOf(exp) !== -1
+    ),
+    subType: RelationalOperator.Contains,
+    regexp: /^(contains)/i
+}, LEXEME_BASE);
```
Marc-Andre-Rivet: Added `contains` relational operator -- works only for strings -- to eventually be allowed on string columns/data

Marc-Andre-Rivet:
@alexcjohnson Adding a changelog entry. This might also be the last good time to rename `filtering_settings`. Leaning towards `filter` to stay consistent with other existing props and renaming `derived_query_structure` to `derived_filter_structure`

alexcjohnson:
👍 for `filter`. We're already using that name in conditional styles & tooltips, and even though the usage is a little different (can be simpler) it's similar enough that keeping the same name makes sense.

PR #410, opened 2019-04-17
Improve filtering type awareness

Marc-Andre-Rivet:
Follow up to #397

**Feature**
- `equal` operator: if both lhs and rhs are numeric, converts and compares; otherwise compare lhs and rhs as-is
- `contains` operator: expects a stringify-able lhs and a string rhs, returns true if the stringified lhs contains the rhs
- `like` operator: expects a date string lhs and rhs, normalizes lhs and rhs, returns true if the normalized lhs starts with the normalized rhs

- column-type based default operator
    Text, Any: contains
    Numeric: equal
    Date: like

**Maintenance**
- rework the `expression` vs `operand` lexeme differentiation so that only expressions remain; the old operand is now one of the expression cases

- refactor the terminal & validity logic of the various lexicons so they share the code instead of duplicating it


.circleci/config.yml
```diff
@@ -170,5 +236,7 @@ workflows:
         jobs:
             - "python-3.6"
             - "node"
-            - "test"
+            - "server-test"
+            - "standalone-test"
+            - "unit-test"
```
Marc-Andre-Rivet: These were taking way too long. Breaking up the server, standalone and unit tests.

src/dash-table/components/FilterFactory.tsx
```diff
@@ -48,14 +48,14 @@ export default class FilterFactory {
...
+    private onChange = (column: IVisibleColumn, setFilter: SetFilter, ev: any) => {
+        Logger.debug('Filter -- onChange', column.id, ev.target.value && ev.target.value.trim());
 
         const value = ev.target.value.trim();
-        const safeColumnId = columnId.toString();
+        const safeColumnId = column.id.toString();
 
         if (value && value.length) {
-            this.ops.set(safeColumnId, new SingleColumnSyntaxTree(safeColumnId, value));
+            this.ops.set(safeColumnId, new SingleColumnSyntaxTree(value, column));
```
Marc-Andre-Rivet: Now need the id and type information in the syntax tree in order to (1) apply the default field, (2) apply the right relational operator

src/dash-table/syntax-tree/SingleColumnSyntaxTree.ts
```diff
@@ -23,29 +36,38 @@ function isUnary(lexemes: ILexemeResult[]) {
...
+                value: getDefaultRelationalOperator(config.type)
+            },
             ...res.lexemes
         ];
     }
 
     return res;
 }
 
+export type SingleColumnConfig = RequiredPluck<IVisibleColumn, 'id'> & OptionalPluck<IVisibleColumn, 'type'>;
```
Marc-Andre-Rivet: `Partial<IVisibleColumn>` was problematic as `id` had to be redefined on top, causing potential misalignment issues + exposed the entire column interface for no good reason. Plucking and optionally plucking props from the type instead to create a new tailor-made partial type.

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -73,6 +83,24 @@ export const greaterThan: IUnboundedLexeme = R.merge({
...
 }, LEXEME_BASE);
 
+const DATE_OPTIONS: IDateValidation = {
+    allow_YY: true
+};
+
+export const likeDate: IUnboundedLexeme = R.merge({
+    evaluate: relationalEvaluator(([op, exp]) => {
+        const normalizedOp = normalizeDate(op, DATE_OPTIONS);
+        const normalizedExp = normalizeDate(exp, DATE_OPTIONS);
```
Marc-Andre-Rivet: Try to normalize content with the most flexible configuration (DATE_OPTIONS) possible

Marc-Andre-Rivet: Otherwise, would need to know the source field + column configuration for a given `evaluate` -- not sure how I'd want to expose that atm

src/dash-table/syntax-tree/lexicon/column.ts
```diff
@@ -35,6 +36,7 @@ const lexicon: ILexeme[] = [
         greaterThan,
         lessOrEqual,
         lessThan,
+        likeDate,
```
Marc-Andre-Rivet: Adding the `like` operator to global, multi column and single column queries

webpack.test.config.js
```diff
@@ -1,6 +1,14 @@
...
     },
     'development'
-);
\ No newline at end of file
+);
+
+config.module.rules.forEach(rule => {
+    if (rule.loader) {
+        rule.loader = rule.loader.replace('ts-loader', `ts-loader?${JSON.stringify({ transpileOnly: true })}`);
+    }
```
Marc-Andre-Rivet: This is a bit hackish but it prevents TS from building up a types mapping, which is very costly in terms of memory usage.. while it's good to do so in the actual build, doing so in Cypress provides very little value and just causes a `heap out of memory` exception after a while.

src/dash-table/components/FilterFactory.tsx
```diff
@@ -65,26 +65,26 @@ export default class FilterFactory {
 
         const rawGlobalFilter = R.map(
             ast => ast.query || '',
-            R.filter(ast => Boolean(ast), asts)
+            R.filter<SingleColumnSyntaxTree>(ast => Boolean(ast), asts)
         ).join(' && ');
```
alexcjohnson: not new in this PR, but what's `ast => ast.query || ''` about? If we actually ever got `''` we'd see `' &&  && '` in the result, wouldn't we?

Marc-Andre-Rivet: 🐑 You're right. Will try and figure out why I did this in the first place and replace it..

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -73,6 +83,24 @@ export const greaterThan: IUnboundedLexeme = R.merge({
...
+
+        return !R.isNil(normalizedOp) &&
+            !R.isNil(normalizedExp) &&
+            // IE11 does not support `startsWith`
+            normalizedOp.indexOf(normalizedExp) === 0;
```
alexcjohnson: if we're worried about perf, possibly faster to avoid `indexOf`:
```js
(
    normalizedOp.length === normalizedExp.length ?
    normalizedOp : normalizedOp.substr(0, normalizedExp.length)
) === normalizedExp
```

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -73,6 +83,24 @@ export const greaterThan: IUnboundedLexeme = R.merge({
...
+    }),
+    subType: RelationalOperator.LikeDate,
+    regexp: /^(like)/i
```
alexcjohnson: still not sure about naming this just `like` - at some point we may want a more sql-like `like` supporting wildcard string matching - or even if we don't add that and instead add a more powerful regex matcher, users may expect `like` to have that kind of behavior. Anyway this one is so tightly coupled to dates, it really feels like it needs to reference date in its name.

`datelike` / `likedate` / `datehas`? @chriddyp thoughts? For context, this is what @Marc-Andre-Rivet and I ended up with as the default operator for dates: match as many fields as you provide in the filter, with anything afterward - ie `like '2011-05'` matches any day, any time, in May 2011, while `like '2011-05-03 12'` matches anything in the 12:00 hour of May 3, 2011.

chriddyp: Hm..
- `datecontains`? I suppose some folks might think "Wednesday" would be valid here (capturing all Wednesdays across all time)
    - Although, perhaps the "Wednesday" filter could be "dateperiod"?
- `datewithin`?

src/dash-table/components/FilterFactory.tsx
Marc-Andre-Rivet: Ok, so here I am creating a `raw` query for caching purposes. It does not matter that the resulting query could contain `&& &&` or any other invalid "real" query combination. I am simply making sure the top-level table component has something to hang on to -- to know the user changed something in order to re-render / bust caching on the appropriate parts (e.g. filter validation / highlighting). This could probably be reworked / improved so as to not require this made up field in the future.

Otherwise, one could argue that I could filter out empty queries but that's about it. On the other hand, this other usage [here](https://github.com/plotly/dash-table/pull/410/files#diff-cb425151ed75b9da3b79ada8820b03aeR15) did need to be changed for the same reasons you mentioned above.

tests/cypress/tests/unit/dash_table_queries_test.ts
```diff
@@ -0,0 +1,107 @@
...
+                        { name: 'compares 1 to "1"', query: `${c.hideOperand ? '' : '{a} '}eq "1"`, target: { a: 1 }, valid: true, evaluate: true },
+                        { name: 'compares 1.1 to 1', query: `${c.hideOperand ? '' : '{a} '}eq 1`, target: { a: 1.1 }, valid: true, evaluate: false },
+                        { name: 'compares 1.1 to "1"', query: `${c.hideOperand ? '' : '{a} '}eq "1"`, target: { a: 1.1 }, valid: true, evaluate: false },
+                        { name: 'compares 1 to 0x1', query: `${c.hideOperand ? '' : '{a} '}eq 0x1`, target: { a: 1 }, valid: true, evaluate: true },
+                        { name: 'compares 1 to "0x1"', query: `${c.hideOperand ? '' : '{a} '}eq "0x1"`, target: { a: 1 }, valid: true, evaluate: true },
+
+                        { name: 'compares "x1" to 1', query: `${c.hideOperand ? '' : '{a} '}eq 1`, target: { a: 'x1' }, valid: true, evaluate: false },
+                        { name: 'compares "x1" to "1"', query: `${c.hideOperand ? '' : '{a} '}eq "1"`, target: { a: 'x1' }, valid: true, evaluate: false }
```
alexcjohnson: Great tests 🎉 A couple more cases I'd like for my peace of mind:
For `eq`:
`eq "1.0"` & `{a: '1'}` -> `true`
`eq " 1 "` & `{a: '\t1\t'}` -> `true`
`eq 1` & `{a: '1x'}` -> `false` (and reversed `eq '1x'` & `{a: 1}`)

For `contains`:
`contains "b"` & `{a: 'abc'}` -> `true`
`contains " b"` & `{a: 'abc'}` -> `false`
`contains "b "` & `{a: 'abc'}` -> `false`
`contains " b"` & `{a: 'a bc'}` -> `true`
`contains "b "` & `{a: 'ab c'}` -> `true`

For `like` (or whatever we call it), just a couple more showing that trailing zeros in the filter don't get dropped:
`like "2005-01-01 00"` & `{a: '2005-01-01 01'}` -> `false`
`like "2005-01-01 00"` & `{a: '2005-01-01 00:01'}` -> `true`
`like "2005-01-01 00:00"` & `{a: '2005-01-01 00:01'}` -> `false`
and a couple of 2-digit years to and from 2-digit and 4-digit years

src/dash-table/syntax-tree/lexeme/relational.ts
Marc-Andre-Rivet: > datecontains? I suppose some folks might think "Wednesday" would be valid here (capturing all Wednesdays across all time)

Variations on this discussed with Alex previously. Later on it should be possible to add functions or transformations that would allow stuff like that, e.g.

`dayOfWeek({dateField}) eq 3` or `month({dateField}) eq 10` -- going with numbers just not to get into a i18n/l10n/filtering discussion

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -47,16 +51,21 @@ const LEXEME_BASE = {
 
 export const contains: IUnboundedLexeme = R.merge({
     evaluate: relationalEvaluator(([op, exp]) =>
-        typeof op === 'string' &&
-        typeof exp === 'string' &&
-        op.indexOf(exp) !== -1
+        !R.isNil(exp) &&
+        !R.isNil(op) &&
+        (R.type(exp) === 'String' || R.type(op) === 'String') &&
+        op.toString().indexOf(exp.toString()) !== -1
```
Marc-Andre-Rivet: Reworked the `contains` logic vs. what we had previously discussed. With `operand` disappearing and lhs/rhs always handled with `expression`, it seemed unnatural to have an asymmetric `contains` operator that would do `{a -> 1} contains "1" --> true` but would do `"1" contains {a -> 1} --> false`.

Instead, it now makes sure that both statements are defined and that at least one is a string -- if those requirements are met, it coerces to string if necessary and does the check. The two queries above will now be true.

tests/cypress/tests/unit/dash_table_queries_test.ts
Marc-Andre-Rivet: ✍️ Added

src/dash-table/syntax-tree/lexeme/relational.ts
alexcjohnson: Right, we'll create lots of other functions later to do more complex and precise filtering. This one will be used almost exclusively as the default operator, so it doesn't need to have a particularly efficient name since hardly anyone will type it. But it does need to *have* a name, as it would be strange if there were behavior you can get with a simple filter that you can't replicate at all with the more complex syntax.

OK, so one more try: `datestartswith`? Still lets us add `startswith` later for strings, but makes clear that it behaves differently for dates, and as I said I don't think it needs to be short, better to be explicit.

PR #412, opened 2019-04-21
Row IDs

### LINKED ISSUE #312: Row ID Support [Sponsored: Due March 1]
Several properties provide features on a per-cell basis (per-cell dropdowns, per-cell tooltips, per-cell borders). Data in the table is provided through a `data` property and this property only contains row indices, column IDs, and the value of that cell for that particular row index and column ID.

The row indices aren’t fixed: data can be filtered (rows removed) or sorted (rows re-ordered) or updated via the backend, and the mapping between these properties and the data can become ambiguous.
To circumvent this, the DataTable needs a concept of “row IDs”, similar to the “column IDs” that currently exist. These IDs would be supplied by the dash developer and would not affect the order of the rows. It would be similar to a “hidden column”.

alexcjohnson:
Row ID support - closes #312 
- `data` - each row can have `id` key, used as row id (can also be displayed if there’s a column with `id='id'`)
- `active_cell` -> `{row, column, row_id, column_id}`
- `selected_cells` -> array like `active_cell`
- `selected_rows` -> added `selected_row_ids`
- `start_cell` and `end_cell` -> these actually didn't seem to be wired up? Now they look like `active_cell`
- `derived` -> added
    - `derived_viewport_row_ids`
    - `derived_virtual_row_ids`
    - `derived_virtual_selected_row_ids`
    - `derived_viewport_selected_row_ids`

Still need to add tests, but AFAICT it works as intended.

Simple test app with pagination (so at least `viewport` indices are different) and row selection & deletion:

```py
import json
import pandas as pd

import dash
from dash.dependencies import Input, Output

data = df.to_dict("rows")
for row in data:
    row['id'] = 'id:' + row['State'].lower()

app = dash.Dash(__name__)
app.layout = html.Div([
    DataTable(
        id='main',
        columns=[{'name': i, 'id': i} for i in df.columns],
        data=data,
        row_selectable='multi',
        row_deletable=True,
        pagination_settings={'page_size': 3, 'current_page': 0}
    ),
    html.Hr(),
    DataTable(
        id='out',
        columns=[{'name': 'prop', 'id': 'prop'}, {'name': 'val', 'id': 'val'}],
        data=[],
        style_cell_conditional=[
            {'if': {'column_id': 'val'}, 'textAlign': 'left'}
        ]
    )
])


props = [
    'active_cell', 'start_cell', 'end_cell', 'selected_cells',
    'selected_rows', 'selected_row_ids',
    'derived_viewport_indices', 'derived_viewport_row_ids',
    'derived_virtual_indices', 'derived_virtual_row_ids',
    'derived_viewport_selected_rows', 'derived_viewport_selected_row_ids',
    'derived_virtual_selected_rows', 'derived_virtual_selected_row_ids'
]


@app.callback(Output('out', 'data'), [Input('main', v) for v in props])
def show_props(*args):
    return [
        {'prop': prop, 'val': json.dumps(val)}
        for prop, val in zip(props, args)
    ]

if __name__ == '__main__':
    app.run_server(debug=True)
```


chriddyp:
🌟 Looks pretty clean!

dash_table/Format.py
```diff
@@ -1,6 +1,4 @@
 import collections
-import inspect
-import sys
```
alexcjohnson: dunno why, but locally `npm run lint` asked me to make the changes in this PR, even though on CI it passes 🤔

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -363,69 +364,88 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-        } else if (selectingRight && maxCol + 1 <= columns.length - 1) {
-            // Otherwise move selection right if possible
-            targetCells = selectedRows.map(row => [row, maxCol + 1]);
+        let startRow = start_cell && start_cell.row;
+        let startCol = start_cell && start_cell.column;
+        let endRow = end_cell && end_cell.row;
+        let endCol = end_cell && end_cell.column;
+
+        if (selectingDown) {
+            if (active_cell.row > minRow) {
```
alexcjohnson: I changed the logic here to be explicitly rectangular - ie picking out the region edges, editing these, and then recreating the `selected_cells` list from that at the end - rather than directly manipulating `selected_cells` as you had previously. I did this because it seemed both clearer and faster, I needed the new edges anyway for `end_cell`, and this way `selected_cells` will always be ordered as it appears on screen.

I considered changing these conditions from `active_cell` to `start_cell`. These are only different if you move `active_cell` after making the selection (via `tab` or `enter`), and as it stands some weird things can happen to the selection if you alter it via `shift-arrow` at that point - it's hard to predict whether `shift-down` will add a row at the bottom or remove one at the top. If I used `start_cell` instead of `active_cell` here, the cell you first clicked on would always be a corner of the selection, and `shift-arrow` would move the opposite corner. But there are two problems with that approach: (1) after you've moved `active_cell`, there's no visual indicator of which corner you started on, and (2) it's possible to shrink the selection so `active_cell` is no longer part of the selection. So in the end I left it with `active_cell`, but we could revisit that.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -363,69 +364,88 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-        });
+            end_cell: makeCell(endRow, endCol, columns, viewport),
+            selected_cells: finalSelected
+        };
+
+        const newStartRow = endRow === minRow ? maxRow : minRow;
+        const newStartCol = endCol === minCol ? maxCol : minCol;
+
+        if (startRow !== newStartRow || startCol !== newStartCol) {
+            newProps.start_cell = makeCell(
```
alexcjohnson: updating `start_cell` only for that strange situation of reshaping the selection after moving `active_cell`

src/dash-table/derived/cell/operations.tsx
```diff
@@ -2,31 +2,38 @@ import * as R from 'ramda';
...
-        newProps.active_cell = [];
-    }
 
-    if (R.is(Array, selectedRows) && R.contains(idx, selectedRows)) {
-        newProps.selected_rows = R.without([idx], selectedRows);
+    if (R.is(Array, selectedRows) && R.any(i => i >= idx, selectedRows)) {
+        newProps.selected_rows = R.map(
+            // all rows past idx have now lost one from their index
+            i => i > idx ? i - 1 : i,
+            R.without([idx], selectedRows)
```
alexcjohnson: This was a bug previously - delete a row and any selected rows after it kept the same index, not the same row contents.

src/dash-table/derived/cell/operations.tsx
```diff
@@ -2,31 +2,38 @@ import * as R from 'ramda';
...
     Indices
 } from 'dash-table/components/Table/props';
 
-function deleteRow(idx: number, activeCell: ActiveCell, data: Data, selectedRows: number[]) {
+function deleteRow(idx: number, data: Data, selectedRows: number[]) {
     const newProps: any = {
-        data: R.remove(idx, 1, data)
+        data: R.remove(idx, 1, data),
+        // We could try to adjust selection, but there are lots of edge cases
+        ...clearSelection
```
alexcjohnson: previously we left `selected_cells` intact when deleting a row, and only cleared `active_cell` if it was actually deleted. But that has some problems, and for simplicity and symmetry with [`deleteColumns`](https://github.com/plotly/dash-table/pull/412/commits/572cdf13ae78eac83fdc79ef5db5bc67468d8cc7#diff-4102a065e79334d16000051ec35189f7R46) I decided to just clear the selection here too.

src/dash-table/derived/paginator.ts
```diff
@@ -21,15 +23,15 @@ function getBackEndPagination(
     return {
         loadNext: () => {
             pagination_settings.current_page++;
-            setProps({ pagination_settings });
+            setProps({ pagination_settings, ...clearSelection });
```
alexcjohnson: Clearing cell selection on page changes. You can select *rows* on separate pages and have them stay selected... but that's with row-by-row checkboxes, an inherently stickier interaction that feels right to persist even when a checked box is on a hidden page.

src/dash-table/handlers/cellEvents.ts
```diff
@@ -1,57 +1,69 @@
...
 
+    /*
+     * In some cases this will initiate browser text selection.
+     * We've hijacked copying, so while it might be nice to allow copying part
+     * of a cell, currently you'll always get the whole cell regardless of what
+     * the browser thinks is selected.
+     * And when you've selected multiple cells the browser selection is
+     * completely wrong.
+     */
+    window.getSelection().removeAllRanges();
```
alexcjohnson: For example, shift-click to select in the first example at https://dash.plot.ly/datatable:
Screen Shot 2019-04-22 at 1 43 56 AM

src/dash-table/handlers/cellEvents.ts
```diff
@@ -1,57 +1,69 @@
-import * as R from 'ramda';
-import { SelectedCells, ICellFactoryProps } from 'dash-table/components/Table/props';
+import { min, max, set, lensPath } from 'ramda';
```
alexcjohnson: I made a couple of forays into specific ramda imports - my local linter complains about `* as R` and as @chriddyp pointed out (somewhere?) we may get a lighter build that way too.

src/dash-table/handlers/cellEvents.ts
```diff
@@ -1,57 +1,69 @@
...
-                R.max(minCol, cellLocation[1]) + 1
-            )
-        ) as any;
+    if (e.shiftKey && start_cell && start_cell.row !== undefined) {
+        newProps.selected_cells = makeSelection(
+            {
+                minRow: min(row, start_cell.row),
+                maxRow: max(row, start_cell.row),
+                minCol: min(col, start_cell.column),
+                maxCol: max(col, start_cell.column)
```
alexcjohnson: Oh hmm, here I used `start_cell` and not `active_cell` like I did in the keypress handler... and indeed it can have weird effects like `active_cell` becoming outside the selection. So one of these should change... which one?

src/dash-table/handlers/cellEvents.ts
```diff
@@ -1,57 +1,69 @@
...
+        viewport
     } = propsFn();
 
-    const selected = isCellSelected(selected_cells, idx, i);
+    const selected = isSelected(selected_cells, idx, i);
 
     // don't update if already selected
     if (selected) {
         return;
     }
```
alexcjohnson: BTW in case of `shift-click` I think this return is problematic - it means you can't shrink a selection without doing something hacky like first clicking over to the other side of the starting point.

CHANGELOG.md
```diff
@@ -14,10 +14,24 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+    - Added new props mirroring all existing row indices props:
+        - `selected_row_ids` mirrors `selected_rows`
+        - `derived_viewport_row_ids` mirrors `derived_viewport_indices`
+        - `derived_virtual_row_ids` mirrors `derived_virtual_indices`
+        - `derived_viewport_selected_row_ids` mirrors `derived_viewport_selected_rows`
+        - `derived_virtual_selected_row_ids` mirrors `derived_virtual_selected_rows`
+
 ### Changed
 [#397](https://github.com/plotly/dash-table/pull/397)
 - Rename `filtering_settings` to `filter`
```
chriddyp: I realize this wasn't done in this PR, but should there be some symmetry with `sorting_settings` -> `sort` here?

chriddyp:
Here's a "classic" filtering example. Nice use of row IDs - by only passing back a single column, we're able reduce the network size by about 5x (5 columns in this dataset). Previously, we had to pass back _all_ of the columns of data.

Also, the code is pretty elegant - using IDs to index pandas dataframes is pretty natural:
```python
filtered_df.loc[selected_row_ids, 'color'] = '#85144b'
```

Here's the full snippet:
```python
import dash
from dash.dependencies import Input, Output
import dash_core_components as dcc
import dash_table
import pandas as pd

df = pd.read_csv('dataset.csv')
# Assign row IDs
df['id'] = df.index

app = dash.Dash(__name__)

app.layout = html.Div([
    html.H3('Row IDs'),

    dash_table.DataTable(
        id='table',
        columns=[{
            'id': c,
            'label': c
        } for c in df.columns if c != 'id'],
        data=df.to_dict('records'),
        virtualization=True,
        filtering='fe',
        sorting=True,
        pagination_mode=False,
        row_selectable='multi',
        derived_virtual_selected_row_ids=[]
    ),

    dcc.Graph(id='graph')
])


@app.callback(
    Output('graph', 'figure'),
    [Input('table', 'derived_virtual_row_ids'),
     Input('table', 'derived_virtual_selected_row_ids')])
def display_graph(row_ids, selected_row_ids):
    if row_ids is not None:
        filtered_df = df.loc[row_ids, :].copy()
    else:
        filtered_df = df.copy()

    filtered_df['color'] = '#0074D9'
    filtered_df.loc[selected_row_ids, 'color'] = '#85144b'

    return {
        'data': [
            {
                'x': filtered_df['country'],
                'y': filtered_df['gdpPercap'],
                'type': 'bar',
                'marker': {
                    'color': filtered_df['color']
                }
            }
        ],
        'layout': {
            'uirevision': 'constant',
            'yaxis': {
                'title': 'gdpPercap'
            }
        }
    }


if __name__ == '__main__':
    app.run_server(debug=True)
```

row-ids

CHANGELOG.md
Marc-Andre-Rivet: Yes. Used the filtering breaking changes as an opportunity to rename the props but limited myself to those.

dash_table/Format.py
Marc-Andre-Rivet: Running the same versions? I've found that the default rules are updated frequently..

src/dash-table/dash/DataTable.js
```diff
@@ -79,12 +83,18 @@ export const defaultProps = {
...
+});
+
 export const propTypes = {
     /**
-     * The [row, column] index of which cell is currently
-     * active.
+     * The row and column indices and IDs of the currently active cell.
      */
-    active_cell: PropTypes.array,
+    active_cell: cellCoordsIDsProp,
```
Marc-Andre-Rivet: Assuming the resulting metadata.json is correct b/c the value is defined in the same file?

alexcjohnson: oh good call... looks like I can't 🌴 this in the existing `react-docgen` implementation ☹️

src/dash-table/handlers/cellEvents.ts
Marc-Andre-Rivet: Yes. Depending on your linter, it may or may not handle the typescript option for `* as R`.
The way imports are used here webpack should prune the unused content automagically but this is fine too.

tests/visual/percy-storybook/DashTable.percy.tsx
```diff
@@ -64,7 +96,7 @@ storiesOf('DashTable/With Data', module)
         ]}
     />));
 
-const columns = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j']
+const columnsA2J = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j']
```
Marc-Andre-Rivet: Does `A2J` stand for `Alex Johnson was here`? 🤔

alexcjohnson: haha I almost called it `columnsAJ` but thought that might be too selfish. It's supposed to be just `'a' -> 'j'`

dash_table/Format.py
alexcjohnson: oh actually `npm run lint` excluded a whole bunch of rules that we really don't need to exclude (as long as we're excluding the auto-generated files, which we are). I'll un-exclude them. Probably worth taking a look at this in other repos as well... it makes sense to have looser linting in tests but some of these rules are useful - like unused imports, can improve perf, only files setting up namespaces have any reason to include them.

CHANGELOG.md
alexcjohnson: Now does seem like a good time to change this, but I might vote for `sort_by` - `sort` could be interpreted to mean "only sort these columns," rather than the actual "sort all columns based on these ones"

src/dash-table/dash/DataTable.js
alexcjohnson: back to explicit prop types -> https://github.com/plotly/dash-table/pull/412/commits/397b914aa42c84f15ac513a905f2fb1b5e71b917

dash_table/Format.py
alexcjohnson: stricter linting -> https://github.com/plotly/dash-table/pull/412/commits/8f01c257b200d97bee49eb2e344ab46438d4fba9

src/dash-table/handlers/cellEvents.ts
alexcjohnson: standardized on `active_cell` like keypresses https://github.com/plotly/dash-table/pull/412#discussion_r277207194 -> https://github.com/plotly/dash-table/pull/412/commits/1c69cfbe8080524946f53d720ae5535886666136

src/dash-table/handlers/cellEvents.ts
alexcjohnson: better shift-click behavior -> https://github.com/plotly/dash-table/pull/412/commits/1c69cfbe8080524946f53d720ae5535886666136 (and https://github.com/plotly/dash-table/pull/412/commits/3e31a7f33418bc0f2b1830d6cf75f6e67830c736)

alexcjohnson:
@Marc-Andre-Rivet ready for review. I didn't change `sorting_settings` (https://github.com/plotly/dash-table/pull/412#discussion_r278364448), would be happy to do that here or in a separate PR.

package.json
```diff
@@ -22,7 +22,7 @@
     "private::host_dash8083": "python tests/cypress/dash/v_fe_page.py",
     "private::host_js": "http-server ./dash_table -c-1 --silent",
     "private::lint:ts": "tslint '{src,demo,tests}/**/*.{js,ts,tsx}' --exclude '**/@Types/*.*'",
-    "private::lint:py": "flake8 --exclude=DataTable.py,__init__.py,_imports_.py --ignore=E501,F401,F841,F811,F821 dash_table",
+    "private::lint:py": "flake8 --exclude=DataTable.py,__init__.py,_imports_.py dash_table",
```
Marc-Andre-Rivet: Thanks for cleaning that up / removing the exceptions

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -330,6 +331,13 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         // TAB
         event.preventDefault();
 
+        if (!active_cell) {
+            // there should always be an active_cell if we got here...
+            // but if for some reason there isn't, bail out rather than
+            // doing something unexpected
+            return;
```
Marc-Andre-Rivet: If this is unexpected, maybe add a `Logger.warning` here

Issue #207, opened 2018-11-02
Breaking Changes Log


chriddyp:
👋 Dash Community 👋 

We'll keep this issue updated when we make breaking changes to the `DataTable`. Subscribe to this issue to stay aware of these changes.

Note that we'll indicate breaking changes through our versioning system ([semver](https://semver.org/)). If we bump the major version number (`major.minor.micro`), then it means that we've made a breaking change to the API.

Not all breaking changes will be large, so don't worry. We'll always include steps for upgrading between major versions in our CHANGELOG.md.

Thank you for your diligence and support ❤️

GrahamFord:
I only had to update Dash to the latest version (currently 0.31.1) to fix this problem.

chriddyp:
**Breaking change notice**

In 3.2.0, `column.type: 'dropdown'` has been moved to `column.presentation: 'dropdown'`. 

That is, before you had:
```
DataTable(columns=[{'id': 'column-1', 'name': 'Column 1', 'type': 'dropdown'}]
```
This would now be:
```
DataTable(columns=[{'id': 'column-1', 'name': 'Column 1', 'presentation': 'dropdown'}]
```

GitTorres:
Thanks for incorporating this component!

Is `DataTable` still officially in Alpha, as mentioned in the <a href=https://dash.plot.ly/datatable>docs</a>, dated Nov 1, 2018? I'd like to use this for a work project but there will be certain push-back if so.  Thanks!

alexcjohnson:
#446 introduced a big chunk of breaking changes. This is released in dash-table 4.0.0, included in dash 1.0.0, and officially removes the alpha designation - we will not be introducing any more breaking changes in the forseeable future, and if we do it would constitute a new major version of dash and dash-table.

From the CHANGELOG:

#### NEW
    - `column.sort_as_null`: Allows sorting behavior customization.
        Accepts an array of string, number or booleans.

#### REMOVED
    - `column.clearable`: Allows clearing the value of a dropdown cell.
        Removed in favor of `dropdown_**` `clearable` nested property.
    - `column.options`
        Removed. Redundant with `dropdown`.
    - `pagination_settings`
        Replaced by two props `page_current` and `page_size`.

#### RENAMED
    - `column_static_tooltip`
        Renamed to `tooltip`.
    - `column_conditional_tooltips`
        Renamed to `tooltip_conditional`.
    - `filter`
        Renamed to `filter_query`.
    - `sort_type`
[...]

PR #417, opened 2019-04-25
change sorting_settings to sort_by

alexcjohnson:
As discussed https://github.com/plotly/dash-table/pull/412#discussion_r278364448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 [#397](https://github.com/plotly/dash-table/pull/397)
 - Rename `filtering_settings` to `filter`
 
+[#412](https://github.com/plotly/dash-table/pull/412)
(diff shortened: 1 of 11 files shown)


chriddyp:
💃 

In the future, one idea to ease the transition for our users might be to keep the old prop type available in python but then raise a helpful error in the `constructor` that `dev-tools` will pick up.

I'm OK doing a hard switch for now while we're still in alpha.

Once this is merged, could you add a comment to https://github.com/plotly/dash-table/issues/207?

alexcjohnson:
> In the future, one idea to ease the transition for our users might be to keep the old prop type available in python but then raise a helpful error in the `constructor` that `dev-tools` will pick up.

Possibly - would be nice if there were some indication on the python side that you shouldn't use it. Maybe there's a way we could build "obsolete props" into `dash-component-system`

> Once this is merged, could you add a comment to #207?

👍

PR #402, opened 2019-03-22
Border colors rework

valentijnnieman:
This PR aims to replace the way we handle border colors. Still a WIP - pushing up so others can see progress!

Screen Shot 2019-03-22 at 5 29 42 PM

PR #424, opened 2019-04-30
Border colors 2

Marc-Andre-Rivet:
This is a continuation of the previous border colors [PR](https://github.com/plotly/dash-table/pull/402). This takes the essential elements from the PR and while keeping the same underlying ideas that were developed there, makes the implementation a bit tighter.

- [x] use subset of `border**` CSS rules to style the table's borders (border, border-left, border-right, border-bottom, border-top) --> while only supporting a subset might feel limiting, it covers pretty much all cases with very little redundancy vs. the full rules set and simplifies implementation greatly
- [x] omit all `border**` CSS rules from general cell/data/headers/filter/ops styling
- [x] apply default styling
- [x] reconcile common borders between various types & with n_fixed_rows & n_fixed_columns (priority order: data > filter > headers > operations > fillers)

n_fixed_rows, n_fixed_columns cases are handled through CSS: https://github.com/plotly/dash-table/pull/424/files#diff-02b1c6154ad8311a2face953de21380dR159

headers, filters, data and operations are reconciled with priority
1. data
2. filters
3. headers
4. operations

Given that all the items above share style_cell and style_cell_conditional, and that in all cases these are the lowest weight styles applied,

p1 will win on p2 if:
- p1's edge is not the table's default and not from style_cell or style_cell_conditional
- p1's edge weight is greater than p2's edge weight -- the weight is equal to the source style's index

- [x] unit tests for edge matrices generation
- [x] visual tests for table rendering

Note: Percy diffs will be all over the place because of the pixel-differences between `box-shadow` and `border` rendering. We'll need to be careful with that one.


src/core/type/index.ts
```diff
@@ -1,2 +1,10 @@
+export type OptionalMap<T, TR extends keyof T, M> = {
+    [tr in TR]?: M
+};
 export type RequiredPluck<T, R extends keyof T> = { [r in R]: T[r] };
 export type OptionalPluck<T, R extends keyof T> = { [r in R]?: T[r] };
+
+export type RequiredProp<T, R extends keyof T> = T[R];
+export type OptionalProp<T, R extends keyof T> = T[R] | undefined;
+
+export type PropOf<T, R extends keyof T> = R;
```
Marc-Andre-Rivet: Just defining some types to do TypeScript magic later..

src/core/math/arrayZipMap.ts
```diff
@@ -2,7 +2,16 @@ import * as R from 'ramda';
...
 
-export function arrayMap<T1, T2, TR>(
+export function arrayMap<T1, TR>(
+    a1: Array<T1>,
+    cb: (d1: T1, i: number) => TR
+) {
+    const mapArray = R.addIndex<T1, TR>(R.map);
+
+    return mapArray((iValue, i) => cb(iValue, i), a1);
+}
```
alexcjohnson: This seems like a lot of trouble to go through to replicate `a1.map(cb)` - what's the benefit?

Marc-Andre-Rivet: None. Only benefit is keeping with a relatively consistent ramda usage.

src/core/math/matrixZipMap.ts
```diff
@@ -2,70 +2,109 @@ import * as R from 'ramda';
...
     const mapRow = R.addIndex<T1, TR>(R.map);
 
     return mapMatrix((iRow, i) =>
         mapRow(
-            (ijValue, j) => cb(ijValue, m2[i][j], i, j),
+            (ijValue, j) => cb(ijValue, i, j),
+            iRow
+        ), m1
+    );
+}
```
Marc-Andre-Rivet: Adding a 1-matrix map -- otherwise all below is essentially dealing with undefined cases

src/dash-table/components/CellFactory.tsx
```diff
@@ -100,15 +102,28 @@ export default class CellFactory {
...
-            contents,
-            (w, s, c) => React.cloneElement(w, { children: [c], style: s })
+            cellWrappers,
+            cellStyles,
+            cellContents,
+            (w, s, c, i, j) => React.cloneElement(w, {
+                children: [c],
+                style: R.mergeAll([
+                    s,
+                    dataEdges && dataEdges.getStyle(i, j)
```
Marc-Andre-Rivet: Here and above in ops, obtaining the calculated borders for this specific cell

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -735,13 +717,30 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
             virtualized.offset.rows
         );
 
+        const classes = [
+            'dash-spreadsheet',
+            ...(virtualization ? ['dash-virtualized'] : []),
+            ...(n_fixed_rows ? ['dash-freeze-top'] : []),
+            ...(n_fixed_columns ? ['dash-freeze-left'] : []),
+            ...(style_as_list_view ? ['dash-list-view'] : []),
+            ...(empty[0][1] ? ['dash-empty-01'] : []),
```
Marc-Andre-Rivet: Adding a few marker classes for various cases
- empty content & fixed contents
- no data, no columns, no filter

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { EdgesMatrices, IEdgesMatrices } from 'dash-table/derived/edges/type';
+
+import getHeaderRows from 'dash-table/derived/header/headerRows';
+
+import { derivedRelevantCellStyles, derivedRelevantFilterStyles, derivedRelevantHeaderStyles } from 'dash-table/derived/style';
+import { Style, Cells, DataCells, BasicFilters, Headers } from 'dash-table/derived/style/props';
+
+import { ControlledTableProps, VisibleColumns, IViewportOffset, Data } from './Table/props';
+
+export default class EdgeFactory {
```
Marc-Andre-Rivet: This is the central change of this PR -- evaluates styles in isolation & reconciles styles for boundary cases (e.g. shared data / filter borders, etc.)

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,228 @@
...
+        R.forEach(j =>
+            (
+                hNext.getWeight(iNext, j) > cutoffWeight ||
+                (hNext.getWeight(iNext, j) > hTarget.getWeight(iTarget, j))
+            ) && hTarget.setEdge(iTarget, j, undefined, -Infinity, true),
+            R.range(0, hTarget.columns)
+        );
+
+        return target;
+    }
```
Marc-Andre-Rivet: Evaluates boundary horizontal edges between the target edges matrix and the "next" edges matrix.

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,228 @@
...
+        R.forEach(i =>
+            (
+                vNext.getWeight(i, jNext) > cutoffWeight ||
+                (vNext.getWeight(i, jNext) > vTarget.getWeight(i, jTarget))
+            ) && vTarget.setEdge(i, jTarget, undefined, -Infinity, true),
+            R.range(0, vTarget.rows)
+        );
+
+        return target;
+    }
```
Marc-Andre-Rivet: Evaluates boundary vertical edges between the target edges matrix and the "next" edges matrix.

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,228 @@
...
+        const hNext = next.getMatrices().horizontal;
+        const hTarget = target.getMatrices().horizontal;
+
+        const iNext = 0;
+        const iTarget = hTarget.rows - 1;
+
+        R.forEach(j =>
+            (
+                hNext.getWeight(iNext, j) > cutoffWeight ||
+                (hNext.getWeight(iNext, j) > hTarget.getWeight(iTarget, j))
```
Marc-Andre-Rivet: For each edge, if the "next" edge has a weight beyond the cutoff (not style_cell or style_cell_conditional) or if it's below-cutoff weight is higher than the target's, clear the target's edge

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,228 @@
...
+        const vTarget = target.getMatrices().vertical;
+
+        const jNext = 0;
+        const jTarget = vTarget.columns - 1;
+
+        R.forEach(i =>
+            (
+                vNext.getWeight(i, jNext) > cutoffWeight ||
+                (vNext.getWeight(i, jNext) > vTarget.getWeight(i, jTarget))
+            ) && vTarget.setEdge(i, jTarget, undefined, -Infinity, true),
```
Marc-Andre-Rivet: For each edge, if the "next" edge has a weight beyond the cutoff (not style_cell or style_cell_conditional) or if it's below-cutoff weight is higher than the target's, clear the target's edge

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,228 @@
...
+        style_cell_conditional: Cells,
+        style_data: Style,
+        style_data_conditional: DataCells,
+        style_filter: Style,
+        style_filter_conditional: BasicFilters,
+        style_header: Style,
+        style_header_conditional: Headers,
+        data: Data,
+        offset: IViewportOffset
+    ) => {
```
Marc-Andre-Rivet: The overall edges calculation is expensive and cached.
The specific edges calculation are cached.
The style calculation for each section are cache.

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,216 @@
...
+        const cutoffWeight = (style_cell ? 1 : 0) + style_cell_conditional.length - 1;
+
+        headerEdges = this.hReconcile(headerEdges, filterEdges || dataEdges, cutoffWeight);
+        headerOpEdges = this.hReconcile(headerOpEdges, filterOpEdges || dataOpEdges, cutoffWeight);
+        filterEdges = this.hReconcile(filterEdges, dataEdges, cutoffWeight);
+        filterOpEdges = this.hReconcile(filterOpEdges, dataOpEdges, cutoffWeight);
+
+        headerOpEdges = this.vReconcile(headerOpEdges, headerEdges, cutoffWeight);
+        filterOpEdges = this.vReconcile(filterOpEdges, filterEdges, cutoffWeight);
+        dataOpEdges = this.vReconcile(dataOpEdges, dataEdges, cutoffWeight);
```
Marc-Andre-Rivet: As they are cached, these need to be treated as immutable even though they are not. The reconciliation step will clone if required, afterwards the edges are returned as a read-only interface

src/dash-table/components/Table/Table.less
```diff
@@ -1,183 +1,5 @@
 @import (reference) '~dash-table/style/reset.less';
 
-.inset-shadow(@color, @left, @top, @right, @bottom) {
```
Marc-Andre-Rivet: The box shadow declaration/rules are not required anymore 🔪

src/dash-table/components/Table/Table.less
```diff
@@ -334,6 +156,28 @@
...
+            .cell-0-0,
+            .cell-1-0 {
+                tr {
+                    td:last-of-type,
+                    th:last-of-type {
+                        border-right: none !important;
+                    }
+                }
+            }
+        }
```
Marc-Andre-Rivet: Handles the boundary cases for n_fixed_rows and n_fixed_columns -- only apply if the neighbouring table fragment is not empty

src/dash-table/dash/DataTable.js
```diff
@@ -77,8 +77,9 @@ export const defaultProps = {
     row_selectable: false,
 
     style_table: {},
-    style_data_conditional: [],
     style_cell_conditional: [],
+    style_data_conditional: [],
+    style_filter_conditional: [],
```
Marc-Andre-Rivet: default was not defined for `style_filter_conditional`

src/dash-table/derived/edges/type.ts
```diff
@@ -0,0 +1,197 @@
...
+    'borderRight',
+    'borderTop'
+];
+
+export const BORDER_PROPERTIES_AND_FRAGMENTS: string[] = R.uniq(
+    R.filter(
+        p => p.indexOf('border') === 0,
+        Array.from(py2jsCssProperties.values())
+    )
+);
```
Marc-Andre-Rivet: For use later, will omit all `border**` rules from the styles applied to the table cells

src/dash-table/derived/edges/type.ts
```diff
@@ -0,0 +1,197 @@
...
+    constructor(
+        rows: number,
+        columns: number,
+        defaultEdge?: Edge
+    );
+    constructor(
+        rowsOrMatrix: number | EdgesMatrix,
+        columns?: number,
+        defaultEdge?: Edge
+    ) {
```
Marc-Andre-Rivet: Fwiw, this is how a ctor is overloaded in typescript -- define n ctors with their own definitions and then declare one that encompasses them all. As TS generates absolutely no new code, the developer is responsible for making it all work.

tests/cypress/tests/unit/edges_test.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    it('without data has no edges', () => {
+        const res = edgesFn(
+            [{ id: 'id', name: 'id' }],
+            [],
+            [],
+            { columns: 0, rows: 0 }
+        );
+
+        expect(res !== undefined).to.equal(false);
```
alexcjohnson: Nice triple negative 😏 
`expect(res).to.be.undefined`?

tests/cypress/tests/unit/edges_test.ts
```diff
@@ -0,0 +1,319 @@
...
+
+            expect(vertical.length).to.equal(1);
+            expect(vertical[0].length).to.equal(3);
+            expect(vertical[0][0]).to.equal('1px solid green');
+            expect(vertical[0][1]).to.equal('1px solid green');
+            expect(vertical[0][2]).to.equal('1px solid darkgreen');
+        }
+    });
+
+    it('applies `borderLeft` overridden by higher precedence `border`', () => {
```
alexcjohnson: can we test the opposite case too, `borderLeft` overriding only the left with `border` still setting the other three sides

alexcjohnson:
@Marc-Andre-Rivet the percy diffs look fine to me. There's one where the baseline and new image both look wrong: **DashTable/Width percentages: with frozen rows** - there's misalignment between headers and data. Your call if you want to investigate this now or just make an issue for it, but I would like to see screenshots of variable-width borders (a big difference like 1 vs 4 px) to verify they don't cause misalignment:
- thick border on the headers, thin borders on data
- thin borders everywhere except thick borders on one highlighted data cell
- the same two cases between fixed and movable columns.

Also, do we already have a test case for conditional styling by both row number and data value while also sorting/filtering? Can we include borders there?

alexcjohnson:
Oh I didn't look at the python percy diffs at first - two of these seem to have broken:
- **editor-with-configurable-options** - the row selection column got much smaller? I actually kind of like it smaller, not sure why it had the size it did, but it's a substantial change.
- **dataframe-updating-graph-fe** actually the headers were already given sizes unrelated to the filters and data, but the two auto-generated initial columns collapsed making it worse

Marc-Andre-Rivet:
I'm seeing issues with at least:
- filter styling with invalid column query
- active cell border

Need to add more tests for borders. I'm barely covering the basics what the visual tests atm

Marc-Andre-Rivet:
@alexcjohnson Update:

Main problematic visual change I still see in the storybook is:
https://percy.io/plotly/dash-table/builds/1815314/view/118167441/1280?mode=diff&browser=chrome&snapshot=118167441 -- with the last bottom border probably hidden by the dropdown, to be checked.

Not sure about the percy python diffs:
https://percy.io/plotly/dash-table-python-v0/builds/1815331/view/118168907/1280?mode=diff&browser=chrome&snapshot=118168907 -- only seems to be affecting the "operation" column -- I don't see a significant change in the storybook snapshots for the operation columns.

https://percy.io/plotly/dash-table-python-v0/builds/1815331/view/118168743/1280?mode=diff&browser=chrome&snapshot=118168743 -- alignment of the header rows is incorrect, assuming n_fixed_rows and/or n_fixed_columns is set.

Still haven't addressed the issues you found earlier, still need to add more tests for more scenarios.

Not satisfied with the current filter / invalid styling solution (CSS / !important) -- will probably be problematic when styled with higher weight data rows. Will see when I add tests for those..

Marc-Andre-Rivet:
Changed test slightly for https://percy.io/plotly/dash-table/builds/1815314/view/118167441/1280?mode=diff&browser=chrome&snapshot=118167441 -- made table a bit higher to it's clear that the border is rendering.. the change was due to the small row height change with the border vs. the inset-box shadow.

https://percy.io/plotly/dash-table-python-v0/builds/1815331/view/118168907/1280?mode=diff&browser=chrome&snapshot=118168907 and https://percy.io/plotly/dash-table-python-v0/builds/1815331/view/118168743/1280?mode=diff&browser=chrome&snapshot=118168743 -- these are Python/Percy test artefacts, somehow with the implementation changes the old and the new artefacts are different. Rendering seems fine when run locally.

tests/cypress/tests/unit/edges_test.ts
Marc-Andre-Rivet: adding

Marc-Andre-Rivet:
Last known issue: Filter case, using CSS the left border gets overridden by border-collapse giving precedence to the previous cell's borderRight
image

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+        R.forEach(i =>
+            EdgeFactory.hasPrecedence(
+                hPrevious.getWeight(i, jPrevious),
+                hTarget.getWeight(i, jTarget),
+                cutoffWeight
+            ) && hTarget.setEdge(i, jTarget, hPrevious.getEdge(i, jPrevious), Infinity, true)
+            , R.range(0, hPrevious.rows)
+        );
+    }
```
Marc-Andre-Rivet: Evaluates boundary edges between the previous and the target -- and overrides the target's value with the previous value if it does not have precedence. This handles the fixed columns/rows cases

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,314 @@
...
+            style_cell_conditional,
+            style_filter_conditional
+        );
+
+        const headerStyles = this.headerStyles(
+            style_cell,
+            style_header,
+            style_cell_conditional,
+            style_header_conditional
+        );
```
Marc-Andre-Rivet: Calculate all the applicable styles for the various sub-sections of the table

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,314 @@
...
+            headerStyles,
+            style_as_list_view
+        );
+
+        let headerOpEdges = this.getHeaderOpEdges(
+            operations,
+            headerRows,
+            headerStyles,
+            style_as_list_view
+        );
```
Marc-Andre-Rivet: Calculate the edges for each section

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+        const cutoffWeight = (style_cell ? 1 : 0) + style_cell_conditional.length - 1;
+
+        headerEdges = EdgeFactory.clone(headerEdges);
+        headerOpEdges = EdgeFactory.clone(headerOpEdges);
+        filterEdges = EdgeFactory.clone(filterEdges);
+        filterOpEdges = EdgeFactory.clone(filterOpEdges);
+        dataEdges = EdgeFactory.clone(dataEdges);
+        dataOpEdges = EdgeFactory.clone(dataOpEdges);
```
Marc-Andre-Rivet: Clone everything before operating on it... we want to keep the previous memoized results clean

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+        this.hReconcile(headerEdges, filterEdges || dataEdges, cutoffWeight);
+        this.hReconcile(headerOpEdges, filterOpEdges || dataOpEdges, cutoffWeight);
+        this.hReconcile(filterEdges, dataEdges, cutoffWeight);
+        this.hReconcile(filterOpEdges, dataOpEdges, cutoffWeight);
+
+        this.vReconcile(headerOpEdges, headerEdges, cutoffWeight);
+        this.vReconcile(filterOpEdges, filterEdges, cutoffWeight);
+        this.vReconcile(dataOpEdges, dataEdges, cutoffWeight);
```
Marc-Andre-Rivet: Reconcile the boundaries between the table sections

src/dash-table/components/EdgeFactory.tsx
```diff
@@ -0,0 +1,314 @@
...
+        } else if (filtering && n_fixed_rows === headerRows + 1) {
+            this.hOverride(filterEdges, dataEdges, cutoffWeight);
+            this.hOverride(filterOpEdges, dataOpEdges, cutoffWeight);
+        }
+
+        if (_n_fixed_columns === operations) {
+            this.vOverride(headerOpEdges, headerEdges, cutoffWeight);
+            this.vOverride(filterOpEdges, filterEdges, cutoffWeight);
+            this.vOverride(dataOpEdges, dataEdges, cutoffWeight);
+        }
```
Marc-Andre-Rivet: Override styles when necessary around the fixed rows & columns

Marc-Andre-Rivet:
@alexcjohnson This is finally ready for another round. It's a bit of a mess by now though :(

Marc-Andre-Rivet:
Found three more issues while demoing this to @cldougl 

- ops columns don't apply style for conditionals `if > row_index`
- `if > row_index` does not take paging into consideration
- filter columns with invalid values get erased when a new valid value is entered

PR #428, opened 2019-05-10
Upgrade core js

Marc-Andre-Rivet:
Keeping up with the latest and greatest of various dependencies. Been preventing this one from updating through the renovate bot for a while now as it causes a bunch of issues in the tests.

What's the point? Core-js is a library that polyfills newer ES201x functionalities into things the `@babel/env` target browsers will all be able to use.

Fixing all of those here!

### DIFF
--- a/.config/webpack/base.js
+++ b/.config/webpack/base.js
@@ -8,11 +8,13 @@ module.exports = (options = {}) => {
     const babel = options.babel || undefined;
     const preprocessor = options.preprocessor || {};
     const mode = options.mode || 'development';
+    const ts = options.ts || {};
 
     console.log('********** Webpack Environment Overrides **********');
     console.log('Preprocessor', JSON.stringify(preprocessor));
     console.log('mode', mode);
     console.log('babel', JSON.stringify(babel));
+    console.log('ts', JSON.stringify(ts));
 
     return {
         entry: {
@@ -57,7 +59,7 @@ module.exports = (options = {}) => {
                     exclude: /node_modules/,
                     use: [
                         { loader: 'babel-loader', options: babel },
-                        { loader: 'ts-loader' },
+                        { loader: 'ts-loader', options: ts },
                         { loader: 'webpack-preprocessor', options: JSON.stringify(preprocessor) }
                     ]
                 },
--- a/.config/webpack/base.js
+++ b/.config/webpack/base.js
@@ -4,10 +4,15 @@ const packagejson = require('./../../package.json');
 
 const dashLibraryName = packagejson.name.replace(/-/g, '_');
 
-module.exports = (preprocessor = {}, mode = 'development') => {
+module.exports = (options = {}) => {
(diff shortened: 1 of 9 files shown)


.storybook/babel.config.js
```diff
@@ -0,0 +1,9 @@
+const presets = [
+    ['@babel/env', {
+        useBuiltIns: 'entry',
```
Marc-Andre-Rivet: This is the main difference between the storybook configuration and the build/release configuration: `entry` guarantees all core-js stuff is present / not resolved on a per-usage basis like with `usage`. For some reason in storybook per-usage does not identify what's needed correctly..

.config/webpack/base.js
```diff
@@ -50,13 +55,19 @@ module.exports = (preprocessor = {}, mode = 'development') => {
...
                 },
                 {
                     test: /\.js$/,
                     exclude: /node_modules/,
-                    loader: `babel-loader!webpack-preprocessor?${JSON.stringify(preprocessor)}`
-
+                    use: [
+                        { loader: 'babel-loader', options: babel },
+                        { loader: 'webpack-preprocessor', options: JSON.stringify(preprocessor) }
+                    ]
```
Marc-Andre-Rivet: Expanded loaders config, easier to pass options..

package.json
```diff
@@ -62,7 +62,7 @@
     "@types/react-dom": "^16.8.4",
     "@types/react-select": "^1.3.4",
     "babel-loader": "^8.0.5",
-    "core-js": "^2.6.5",
+    "core-js": "^3.0.1",
```
Marc-Andre-Rivet: the updated dep..

tests/cypress/tests/unit/edges_test.ts
```diff
@@ -1,6 +1,15 @@
...
 
+const converter = {
+    checksColumn: () => true,
+    checksFilter: () => true,
+    checksRow: () => true,
+    matchesColumn: () => true,
+    matchesFilter: () => true,
+    matchesRow: () => true
+};
+
```
Marc-Andre-Rivet: typing info goes through better than before.. need to respect interface!

.config/webpack/base.js
```diff
@@ -4,10 +4,17 @@ const packagejson = require('./../../package.json');
...
+    const babel = options.babel || undefined;
+    const preprocessor = options.preprocessor || {};
+    const mode = options.mode || 'development';
+    const ts = options.ts || {};
+
     console.log('********** Webpack Environment Overrides **********');
     console.log('Preprocessor', JSON.stringify(preprocessor));
     console.log('mode', mode);
+    console.log('babel', JSON.stringify(babel));
+    console.log('ts', JSON.stringify(ts));
```
Marc-Andre-Rivet: Have to pass in ts and babel configurations explicitly otherwise they get resolved through some mechanism that does not behave the way I need it to!

PR #442, opened 2019-05-26
Initial commit of R package providing dash-table functionality

### LINKED ISSUE #456: R package support for dash-table
An R package providing support for the `dash-table` component should be made available from this repo, including complete documentation which mirrors the Python docstring content, and is accessible via R's help viewer.

Argument typing information for the component should be edited as needed to better match R's data structures (e.g. "named list" instead of `dict` or "unnamed list" instead of `list`).

rpkyle:
This PR proposes to include an R package providing support for the `DataTable` component.

The command used to build this package was

`dash-generate-components ./src/dash-table/dash dash_table --r-prefix="dash"`

`dashTable` was built using an updated version of the package generation code that incorporates documentation for nested props found in the source JavaScript, which previously were not included in the `dashDataTable.Rd` help file. 

In brief, generating documentation for `PropTypes.exact` entries is now possible.

A helper function is also proposed which simplifies passing in `data.frames` in R as inputs to `data`, similar to the `.to_dict()` method in `pandas`:

https://github.com/plotly/dash-table/blob/ae03f11a6946bdd680259ea31ae8a3f1a8db4a19/R/df_to_list.R#L1-L8

Closes #456.


rpkyle:
Noticed a Percy test indicated that "dash green" should appear, cell border colour was actually "dashed blue." Chatted with @Marc-Andre-Rivet, now fixed in 81ec923.

PR #443, opened 2019-05-27
Fix monorepo build

Marc-Andre-Rivet:


Marc-Andre-Rivet:
As per [comment](https://github.com/plotly/dash-html-components/pull/115#discussion_r288136267), updating to shallow clone everywhere.

Issue #435, opened 2019-05-22
Table selected cell


Marc-Andre-Rivet:
When a cell with default background color is active, a highlight color is applied.
![image](https://user-images.githubusercontent.com/8092993/58188953-8b3d7280-7c87-11e9-9c2c-a4ef9c7bf515.png)

When a cell with a `style_**` background color is active, the highlight color is not visible.
![image](https://user-images.githubusercontent.com/8092993/58188995-a01a0600-7c87-11e9-8d0e-d293f80df5eb.png)

This also applies to selected cells.
![image](https://user-images.githubusercontent.com/8092993/58189044-b32cd600-7c87-11e9-8f32-69900b914190.png)

The active/selected styling is currently done through css [here](https://github.com/plotly/dash-table/blob/master/src/dash-table/components/Table/Table.less#L249) and [here](https://github.com/plotly/dash-table/blob/master/src/dash-table/components/Table/Table.less#L261) -- this styling should be done through the styling pipeline / derivations. For example, the data cells [here](https://github.com/plotly/dash-table/blob/master/src/dash-table/derived/cell/wrapperStyles.ts)

The change should be similar to what has already been done for the active cell's border [here](https://github.com/plotly/dash-table/blob/master/src/dash-table/derived/edges/data.ts#L78)

PR #444, opened 2019-05-28
Issue 435 - Table selected cell  

alinastarkov:
Fixes https://github.com/plotly/dash-table/issues/435


generator/cssPropertiesGenerator.js
```diff
@@ -300,9 +300,6 @@ snakes.forEach(([snake, camel]) => map.set(snake, camel));
 kebabs.forEach(([kebab, camel]) => map.set(kebab, camel));
 camels.forEach(([camel]) => map.set(camel, camel));
 
-map.forEach((value, key) => console.log(value, key));
-console.log(map.size);
-
```
Marc-Andre-Rivet: :hocho: :hole: :+1:

src/dash-table/components/Table/Table.less
```diff
@@ -260,7 +254,7 @@
             &.cell--selected {
                 background-color: var(--selected-background);
             }
-
+
```
Marc-Andre-Rivet: I think we can slice out that one ☝️

There's also an entry in `Dropdown.css` that can be removed (the `border-radius` is redundant anyway as the `border: none` is applied elsewhere in that file..

```
.dash-spreadsheet .cell--selected .Select-control {
    background-color: var(--selected-background);
    border-radius: 0;
}
```

tests/visual/percy-storybook/fixtures.ts
```diff
@@ -444,6 +445,7 @@ export default [
             id: 'table',
             editable: true,
             selected_cells: [[1, 1], [1, 2], [1, 3], [2, 1], [2, 2], [2, 3], [3, 1], [3, 2], [3, 3]],
+            active_cell: [1, 1],
```
Marc-Andre-Rivet: Hum.. one has to wonder why these still work 🤔

src/dash-table/derived/cell/wrapperStyles.ts
```diff
@@ -25,7 +26,10 @@ function getter(
                 columnStyles
             )
         );
-
+        let isSelectedCell: boolean = selectedCells.some(cell => cell.row === index && cell.column === columnIndex);
```
Marc-Andre-Rivet: For consistency, use Ramda instead

PR #454, opened 2019-05-31
Refactor sorting

Marc-Andre-Rivet:
- [x] tests for new `sort_as_none`
- [x] what to do with the python global test (usefulness?)

@alexcjohnson Not quite ready for a full-on review, but I'd like to review this individually before folding it into the bigger refactoring branch as the changes are spanning over most of the table by now.

Taking an approach similar to `locale_format`: a per-column nested prop and a global prop as the default configuration for the whole table. Defaults to `[]` (only None gets processed as None!) - Users can pass strings, numbers and booleans.


src/core/type/index.ts
```diff
@@ -9,5 +9,4 @@ export type OptionalProp<T, R extends keyof T> = T[R] | undefined;
 
 export type PropOf<T, R extends keyof T> = R;
 
-export type Merge<M, N> = Omit<M, Extract<keyof M, keyof N>> & N;
-export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
```
Marc-Andre-Rivet: Typescript 3.5 introduced `Omit` as a first class type -- taking it back out :)

src/dash-table/components/Table/props.ts
```diff
@@ -359,10 +361,13 @@ interface IDerivedProps {
...
-export type SanitizedProps = Merge<PropsWithDefaults, {
-    fixed_columns: number;
-    fixed_rows: number;
-}>;
+export type SanitizedProps = Omit<
+    Omit<
+        Merge<PropsWithDefaults, { fixed_columns: number; fixed_rows: number; }>,
+        'locale_format'
+    >,
+    'sort_as_none'
```
Marc-Andre-Rivet: Sadly can't omit an array of props

src/dash-table/dash/DataTable.js
```diff
@@ -105,7 +106,7 @@ export const propTypes = {
      * Columns describes various aspects about each individual column.
      * `name` and `id` are the only required parameters.
      */
-    columns: PropTypes.arrayOf(PropTypes.shape({
+    columns: PropTypes.arrayOf(PropTypes.exact({
```
Marc-Andre-Rivet: Side-quest: We support `exact` now, use it everywhere we can!

src/dash-table/dash/DataTable.js
```diff
@@ -870,13 +877,11 @@ export const propTypes = {
...
-     * valid values (they will appear first when sorting ascending).
-     * If True, empty strings will be ignored, causing these cells to always
-     * appear last.
-     */
-    sorting_treat_empty_string_as_none: PropTypes.bool,
+    sort_as_none: PropTypes.arrayOf(PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
+        PropTypes.bool
+    ])),
```
Marc-Andre-Rivet: Like `locale_format`, provide a table-level `sort_as_none` that can be overridden by columns

alexcjohnson:
Looks great! No complaints on the functionality, I love the idea of per-column overriding a global. Just needs docstrings.

Not sure what you mean by "what to do with the python global test (usefulness?)" - are you proposing removing it, or asking if it should be extended with the new features?

Marc-Andre-Rivet:
> Not sure what you mean by "what to do with the python global test (usefulness?)" - are you proposing removing it, or asking if it should be extended with the new features?

This is a test app that was added a while back to tweak & test parameters with minimally wiring. I don't have a problem with its existence per say, I just find that it (1) causes visual diffs often (as do a lot of our Percy+Selenium tests) and (2) it provides very little value vs. other visual tests that exist (what is this testing really beside a default-ish table + a bunch of dcc inputs / checkbox / radio buttons?

In fact this visual test is part of a group of 3 tests that don't really have a clear purpose and change often. IMO we either scrap them altogether or make them more stable and more valuable to justify their existence.

alexcjohnson:
Got it. As long as we have at least one end-to-end test and the same front-end functionality is tested elsewhere, I have no problem with removing it.

Marc-Andre-Rivet:
No real e2e testing is done with these apps. They are loaded, a snapshot is taken, the app is closed. 

Removing the "parameter tweak" test as it's the most problematic one and the one with the least redeeming value. The other two involve setting up an app with callbacks involving tables and graphs -- which is not covered elsewhere; while they are unlikely to find any problem smaller than "dash as a whole is broken" on their own it's better than nothing at all.

alexcjohnson:
> Removing the "parameter tweak" test as it's the most problematic one and the one with the least redeeming value. The other two are setting up apps with callbacks involving tables and graphs -- which is not covered elsewhere; while they are unlikely to find any problem smaller than "dash as a whole is broken" on their own it's better than nothing at all.

Sounds good. Yes, "dash as a whole is broken" - in some way that we didn't catch elsewhere but shows up in table because it's the most complicated component we have - is the main reason I'm keen to keep at least one end-to-end test.

Marc-Andre-Rivet:
@alexcjohnson As we seem good with the overall approach, will fold this into the main branch -- this will get final review there in any case.

Issue #320, opened 2018-12-19
ability to conditionally format columns if editing is disabled [Sponsored: Due May 1]


chriddyp:
Enable Dash developers to update the style depending on whether the column is editable or not.
This would likely be an additional option in our `style_data_conditional` property (see e.g. https://dash.plot.ly/datatable/style)

Marc-Andre-Rivet:
All `style_**_conditional` to get an additional `is_editable: boolean` nested prop inside the `if`.

For each, we need to modify the calculation pipeline to take the new condition into consideration.
Have a look at `dash-table/derived/style/index.ts` to see how it all glues together.

alexcjohnson:
Closed by #462

PR #462, opened 2019-06-07
Column editable props

alinastarkov:
- Add colums_editable props to style_data_conditional, style_header_conditional, style _filter_conditional (https://github.com/plotly/dash-table/issues/320)
- Check if a column is editable in sanitize.ts and clean up.


src/dash-table/conditional/index.ts
```diff
@@ -21,8 +21,12 @@ export interface ITypedElement {
     column_type?: ColumnType;
 }
 
+export interface IActiveElement {
```
Marc-Andre-Rivet: IEditableElement instead

PR #463, opened 2019-06-10
Improve performance - Memoize the world

Marc-Andre-Rivet:
Improve table performance for various scenarios (not including cell selection)


src/dash-table/components/CellFactory.tsx
```diff
@@ -111,35 +112,65 @@ export default class CellFactory {
...
+        styles,
+        contents,
+        (w, s, c, i, j) => React.cloneElement(w, {
+            children: [c],
+            style: R.mergeAll([
+                s,
+                edges && edges.getStyle(i, j)
+            ])
+        })
+    ));
```
Marc-Andre-Rivet: ops and cells got regenerated on every render, memoize!

src/dash-table/components/FilterFactory.tsx
```diff
@@ -162,18 +162,49 @@ export default class FilterFactory {
...
+        ops,
+        styles,
+        (o, s, j) => React.cloneElement(o, {
+            style: R.mergeAll([
+                edges && edges.getStyle(0, j),
+                s,
+                o.props.style
+            ])
+        })
+    ));
```
Marc-Andre-Rivet: ops and filters got regenerated on every render, memoize!

src/dash-table/components/FilterFactory.tsx
```diff
@@ -118,7 +120,7 @@ export default class FilterFactory {
         } = this.props;
 
         if (filter_action === TableAction.None) {
-            return [];
+            return NO_FILTERS;
```
Marc-Andre-Rivet: use the same array, always, for no filters

src/dash-table/components/HeaderFactory.tsx
```diff
@@ -53,10 +55,7 @@ export default class HeaderFactory {
 
         const headerRows = getHeaderRows(columns);
 
-        const labels = getLabels(columns, headerRows);
-        const indices = getIndices(columns, labels, merge_duplicate_headers);
-
-        const labelsAndIndices = R.zip(labels, indices);
+        const labelsAndIndices = this.getLabelsAndIndices(columns, headerRows, merge_duplicate_headers);
```
Marc-Andre-Rivet: R.zip resulted in a new object each time, memoize!

src/dash-table/components/HeaderFactory.tsx
```diff
@@ -89,41 +88,85 @@ export default class HeaderFactory {
...
+        styles,
+        contents,
+        (w, s, c, i, j) => React.cloneElement(w, {
+            children: [c],
+            style: R.mergeAll([
+                s,
+                edges && edges.getStyle(i, j)
+            ])
+        })
+    ));
```
Marc-Andre-Rivet: ops and headers got regenerated on every render, memoize!

src/dash-table/derived/header/content.tsx
```diff
@@ -8,18 +8,18 @@ import singleUpdate from 'core/sorting/single';
...
     VisibleColumns,
     IVisibleColumn,
     SetProps,
-    ControlledTableProps,
     TableAction
 } from 'dash-table/components/Table/props';
 import * as actions from 'dash-table/utils/actions';
 
-function deleteColumn(column: IVisibleColumn, columns: VisibleColumns, columnRowIndex: any, setProps: SetProps, options: ControlledTableProps) {
+function deleteColumn(column: IVisibleColumn, columns: VisibleColumns, columnRowIndex: any, setProps: SetProps, data: Data) {
```
Marc-Andre-Rivet: Cleaning up impossible-to-control `options` (equals `props`) parameter

src/dash-table/derived/table/fragments.tsx
```diff
@@ -65,4 +74,4 @@ export default (
             [isEmpty(fixedColumnCells), isEmpty(cells)]
         ]
     };
-};
\ No newline at end of file
+});
```
Marc-Andre-Rivet: Fragments did *not* respect immutability of cells/filters/headers factories output. Switch out `splice` for `slice`

src/dash-table/derived/table/fragments.tsx
```diff
@@ -46,15 +47,23 @@ export default (
             cells) :
         null;
 
+    cells = R.addIndex<JSX.Element[]>(R.map)((row, i) => row.slice(
+        (fixedColumnCells && fixedColumnCells[i].length) || 0
+    ), cells);
```
Marc-Andre-Rivet: Fixed columns is impacted by merged headers, re-use the previous calculation to slice in the right place

src/dash-table/derived/table/index.tsx
```diff
@@ -64,5 +43,23 @@ export default (propsFn: () => ControlledTableProps) => {
...
+
+    return () => {
+        const edges = edgeFactory.createEdges();
+
+        const dataCells = cellFactory.createCells(edges.dataEdges, edges.dataOpEdges);
+        const filters = filterFactory.createFilters(edges.filterEdges, edges.filterOpEdges);
+        const headers = headerFactory.createHeaders(edges.headerEdges, edges.headerOpEdges);
+
+        return merge(dataCells, filters, headers);
+    };
```
Marc-Andre-Rivet: We've memoized everything underneath.. let's memoize the whole table

src/dash-table/components/CellFactory.tsx
```diff
@@ -111,35 +112,65 @@ export default class CellFactory {
...
         );
     }
+
+    getCells = memoizeOne((
+        opCells: JSX.Element[][],
+        dataCells: JSX.Element[][]
+    ) => arrayMap2(
+        opCells,
+        dataCells,
+        (o, c) => [...o, ...c]
```
alexcjohnson: no opCells must be a very common case - could we detect it here and avoid this `arrayMap2` altogether?

Marc-Andre-Rivet: Sure. Out of curiosity, checked out the speed of various ways of doing this
image

Only Chrome shown above, but the spread operator is consistently 2.5-4x slower than `concat`

https://jsperf.com/concat-some-arrays

alexcjohnson: weird... we're still transpiling this, right? I wonder what this expression transpiles into? Maybe Chrome just hasn't bothered optimizing spread yet since presumably most people are still transpiling it.

Marc-Andre-Rivet: The above example is native Chrome support, not what the table outputs. Spread and destructuring operators are not particularly efficient as they require more checks under the hood and have not been optimized over two decades. In most cases this is not a problem..

In the table's case, we transpile with Babel for the lowest common denominator: IE11 -- which has absolutely no ES6 support. `core-js` is used to polyfill a lot of features but this is pure Babel transpilation.

```
return [...o, ...c]; 
```

becomes:

```
function _toConsumableArray(arr) { return _arrayWithoutHoles(arr) || _iterableToArray(arr) || _nonIterableSpread(); }

function _nonIterableSpread() { throw new TypeError("Invalid attempt to spread non-iterable instance"); }

function _iterableToArray(iter) { if (Symbol.iterator in Object(iter) || Object.prototype.toString.call(iter) === "[object Arguments]") return Array.from(iter); }

function _arrayWithoutHoles(arr) { if (Array.isArray(arr)) { for (var i = 0, arr2 = new Array(arr.length); i < arr.length; i++) { arr2[i] = arr[i]; } return arr2; } }

return _toConsumableArray(o).concat(_toConsumableArray(c));
```

PR #465, opened 2019-06-12
Improve styles and edges resolution

Marc-Andre-Rivet:
Ham fisted improvements for the table's render performance when the active and selected cell(s) change. This is a continuation of https://github.com/plotly/dash-table/pull/463 for the more general case.

On my laptop, brings down render time from ~550-650ms back down to ~180-200ms, right in the acceptable range.. and possibly faster than it was before all the new features were added.. 🤔 

Mostly this is done through (1) `partial` calculations that do not take active/selected into consideration, (2) memoizing data cells individually instead of as a matrix, (3) some general improvements for style/edges resolution and shallow cloning of matrices, (4) jumping some hoops to get around the edges objects always changing from render to render.

For a 250rows x 12 cols (10+2 ops) table, the `render` time itself goes from ~170ms to ~50ms, React side-effects from ~350ms down to ~50ms and reflow (on input focus) stays stable at around ~100ms. More optimization is possible but this seems like as good a point as any to stop (with 50%-50% divide, we run into significantly diminishing returns)


src/core/math/matrixZipMap.ts
```diff
@@ -2,6 +2,30 @@ import * as R from 'ramda';
 
 type Matrix<T> = T[][];
 
+export const shallowClone: <T>(
+    m: Matrix<T>
+) => Matrix<T> = R.map(row => row.slice(0));
```
Marc-Andre-Rivet: Utility. This is significantly faster than using `R.clone`

src/core/math/matrixZipMap.ts
```diff
@@ -2,6 +2,30 @@ import * as R from 'ramda';
...
+
+export const traverse2 = <T1, T2, TR>(
+    a1: T1[],
+    a2: T2[],
+    fn: (d1: T1, d2: T2, i1: number, i2: number) => TR
+) => R.addIndex<T1>(R.forEach)((d1, i1) =>
+    R.addIndex<T2>(R.forEach)((d2, i2) =>
+        fn(d1, d2, i1, i2),
+        a2),
+    a1);
```
Marc-Andre-Rivet: Utility. Apply `fn` for all combinations `(a,b)` of lists `A, B`

src/core/math/matrixZipMap.ts
```diff
@@ -2,6 +2,30 @@ import * as R from 'ramda';
...
+export const traverseMap2 = <T1, T2, TR>(
+    a1: T1[],
+    a2: T2[],
+    fn: (d1: T1, d2: T2, i1: number, i2: number) => TR
+): TR[][] => R.addIndex<T1, TR[]>(R.map)((d1, i1) =>
+    R.addIndex<T2, TR>(R.map)((d2, i2) =>
+        fn(d1, d2, i1, i2),
+        a2),
+    a1);
+
```
Marc-Andre-Rivet: Utility. Apply `fn: (a,b) -> res` for all combinations of `(a,b)` of lists `A, B`.

src/dash-table/components/CellFactory.tsx
```diff
@@ -93,15 +100,29 @@ export default class CellFactory {
             setProps
         );
 
-        const cellWrappers = this.cellWrappers(
-            active_cell,
+        const partialCellWrappers = this.cellWrappers.partialGet(
             columns,
             virtualized.data,
+            virtualized.offset
+        );
```
Marc-Andre-Rivet: Wrappers baseline. Without taking user interaction (active, selected cells) into consideration.

src/dash-table/components/CellFactory.tsx
```diff
@@ -93,15 +100,29 @@ export default class CellFactory {
...
+
+        const cellWrappers = this.cellWrappers.get(
+            partialCellWrappers,
             virtualized.offset,
+            active_cell,
             selected_cells
         );
```
Marc-Andre-Rivet: Using the baseline wrappers as default, swap out active/selected for new values

src/dash-table/components/CellFactory.tsx
```diff
@@ -138,22 +159,52 @@ export default class CellFactory {
         (o, c) => o.length ? o.concat(c) : c
     ));
```
Marc-Andre-Rivet: Same as for wrappers. Baseline content + active/selected swapping out

src/dash-table/components/CellFactory.tsx
```diff
@@ -138,22 +159,52 @@ export default class CellFactory {
...
 
+    getDataOpCell = memoizerCache<[number, number]>()((
+        operation: JSX.Element,
+        style: CSSProperties | undefined,
+        borderBottom: any,
+        borderLeft: any,
+        borderRight: any,
+        borderTop: any
```
Marc-Andre-Rivet: The current `edges` resolution always results in new objects, making caching impossible for many scenarios. Break it down into the various borders allows for caching.

This is a caching layer on the cells themselves, as `getDataOpCells` often can't provide meaningful caching at the "table" level.

This is rather ham fisted but works within the current structure.

src/dash-table/components/CellFactory.tsx
```diff
@@ -138,22 +159,52 @@ export default class CellFactory {
...
-            style: R.mergeAll([
-                edges && edges.getStyle(i, j),
-                s,
-                o.props.style
-            ])
-        })
+        (o, s, i, j) => {
+            const edge = edges && edges.getStyle(i, j);
+
+            return this.getDataOpCell.get(i, j)(o, s, edge && edge.borderBottom, edge && edge.borderLeft, edge && edge.borderRight, edge && edge.borderTop);
```
Marc-Andre-Rivet: Defer the resolution of the cell to the cache above

src/dash-table/components/CellFactory.tsx
```diff
@@ -163,12 +214,10 @@ export default class CellFactory {
...
-            children: [c],
-            style: R.mergeAll([
-                s,
-                edges && edges.getStyle(i, j)
-            ])
-        })
+        (w, s, c, i, j) => {
+            const edge = edges && edges.getStyle(i, j);
+
+            return this.getDataCell.get(i, j)(w, c, s, edge && edge.borderBottom, edge && edge.borderLeft, edge && edge.borderRight, edge && edge.borderTop);
```
Marc-Andre-Rivet: Same as for op cells but for the data cells themselves.

The same has not been done for headers and filters as, while it will improve performance slightly, the bulk of the gains are in the data.

src/dash-table/derived/cell/wrapperStyles.ts
```diff
@@ -2,70 +2,56 @@ import * as R from 'ramda';
...
-        if (isSelectedCell) {
-            relevantStyles.push({backgroundColor:  'var(--selected-background)'});
+) => {
+    styles = shallowClone(styles);
+
+    R.forEach(({ row: i, column: j }) => {
+        i -= offset.rows;
+        j -= offset.columns;
+
+        if (i < 0 || j < 0 || styles.length <= i || styles[i].length <= j) {
```
Marc-Andre-Rivet: Now looping on `selected_cells` instead of data/column. Need to check boundaries.

src/dash-table/derived/cell/wrappers.tsx
```diff
@@ -19,40 +19,76 @@ class Wrappers {
 
     }
 
+    partialGet = memoizeOne((
+        columns: VisibleColumns,
+        data: Data,
+        _offset: IViewportOffset
```
Marc-Andre-Rivet: `data` needs to come with `offset` for cache busting purposes but it's not used here. Prefixing with `_` disables parameter usage check with the current TS config

src/dash-table/derived/cell/wrappers.tsx
```diff
@@ -19,40 +19,76 @@ class Wrappers {
...
-                    columnIndex,
-                    column.id,
-                    rowIndex,
-                    this.handlers(Handler.Enter, rowIndex, columnIndex),
-                    this.handlers(Handler.Leave, rowIndex, columnIndex),
-                    this.handlers(Handler.Move, rowIndex, columnIndex)
-                );
-            }, columns
-        ), data));
+            if (i < 0 || j < 0 || wrappers.length <= i || wrappers[i].length <= j) {
```
Marc-Andre-Rivet: Now looping on `selected_cells` instead of data/column. Need to check boundaries.

src/dash-table/derived/style/index.ts
```diff
@@ -125,4 +127,20 @@ export const derivedTableStyle = memoizeOneFactory(
...
+        )
+    );
+}
+
+export const getDataCellStyle = (datum: Datum, i: number, column: IVisibleColumn) => R.compose(resolveStyle, matchesDataCell(datum, i, column));
+export const getDataOpCellStyle = (datum: Datum, i: number) => R.compose(resolveStyle, matchesDataOpCell(datum, i));
+export const getFilterCellStyle = (column: IVisibleColumn) => R.compose(resolveStyle, matchesFilterCell(column));
+export const getFilterOpCellStyle = () => R.compose(resolveStyle, getFilterOpStyles);
+export const getHeaderCellStyle = (i: number, column: IVisibleColumn) => R.compose(resolveStyle, matchesHeaderCell(i, column));
+export const getHeaderOpCellStyle = (i: number) => R.compose(resolveStyle, getHeaderOpStyles(i));
```
Marc-Andre-Rivet: Define composed functions for how to filter and reduce styles for various usage scenarios (data, dataOp, filters, etc.)

src/dash-table/derived/edges/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    return res;
+}
+
+export const getDataCellEdges = (datum: Datum, i: number, column: IVisibleColumn) => R.compose(resolveEdges, matchesDataCell(datum, i, column));
+export const getDataOpCellEdges = (datum: Datum, i: number) => R.compose(resolveEdges, matchesDataOpCell(datum, i));
+export const getFilterCellEdges = (column: IVisibleColumn) => R.compose(resolveEdges, matchesFilterCell(column));
+export const getFilterOpCellEdges = () => R.compose(resolveEdges, getFilterOpStyles);
+export const getHeaderCellEdges = (i: number, column: IVisibleColumn) => R.compose(resolveEdges, matchesHeaderCell(i, column));
+export const getHeaderOpCellEdges = (i: number) => R.compose(resolveEdges, getHeaderOpStyles(i));
```
Marc-Andre-Rivet: Define composed functions for how to filter and reduce styles into edges for various usage scenarios (data, dataOp, filters, etc.)

src/dash-table/conditional/index.ts
```diff
@@ -82,4 +85,36 @@ export function ifEditable(condition: IEditableElement | undefined, isEditable:
...
+));
+
+export const getFilterOpStyles: Filter<IConvertedStyle> = R.filter<IConvertedStyle>((style =>
+    !style.checksColumn()
+));
+
+export const getHeaderOpStyles = (i: number): Filter<IConvertedStyle> => R.filter<IConvertedStyle>((style =>
+    style.matchesRow(i) &&
+    !style.checksColumn()
+));
```
Marc-Andre-Rivet: Flavored filter helpers for various usage scenarios (data, dataOp, filter, etc.) -- used as building blocks by edges and style calcs

src/core/math/matrixZipMap.ts
alexcjohnson: I have to wonder whether all these little utilities wouldn't be better off without `ramda` at all - in plotly.js we don't even use `Array.forEach` for really hot paths, only `for`:
```js
for(let i1 = 0; i1 < a1.length; i1++) {
    for(let i2 = 0; i2 < a2.length; i2++) {
        fn(a1[i1], a2[i2], i1, i2);
    }
}
```

Marc-Andre-Rivet: Based on this comment, removed Ramda from the utility functions to see what sort of impact it has.

On average, through 7 renders (don't ask.. glad you did!.. my performance tab was opened in such a way that ~7 renders would fit..) seeing a small but consistent 10-15% speed increase for the utility calls with the changes in https://github.com/plotly/dash-table/pull/465/commits/62e519d2bee18b6a305ab6438cb3ee3935c4581b. Real world gains are in the order of 3-4ms per render cycle.. or about 2% of the render time.

This code is very isolated/abstracted so I see no impact to adding these changes in terms maintainability as long as all the tests are still passing. To be confirmed!

Marc-Andre-Rivet:
Turns out 62e519d had side-effects. Reverting. Will explore the causes after this is merged instead.

PR #466, opened 2019-06-13
Optimize utility functions

Marc-Andre-Rivet:
This does 2 things:

1. Optimize the utility functions (mostly removing Ramda for/map/addIndex in favor of native `for` loops)
This drops average render time from ~50ms down to ~35ms.

2. Optimize the previously optimized edges & styles calculations (https://github.com/plotly/dash-table/pull/465). This is done by removing Ramda where possible and through some other optimizations.
This drops the average render time from ~35ms to ~23ms.

Reflow and react internals remain unchanged with ~100ms and ~50ms per render respectively while table `render` itself is slashed in half. This only represents a ~10% improvement.

At this point ~45% (~10-10.5ms) of the table's `render` time is accounted for by cache busting / checks on memoized functions.

### DIFF
--- a/src/core/math/arrayZipMap.ts
+++ b/src/core/math/arrayZipMap.ts
@@ -1,24 +1,34 @@
-import * as R from 'ramda';
-
 type Array<T> = T[];
 
 export function arrayMap<T1, TR>(
     a1: Array<T1>,
     cb: (d1: T1, i: number) => TR
-) {
-    const mapArray = R.addIndex<T1, TR>(R.map);
+): Array<TR> {
+    const _a1_ = a1.length;
 
-    return mapArray((iValue, i) => cb(iValue, i), a1);
+    const res: Array<TR> = [];
+
+    for (let i = 0; i < _a1_; ++i) {
+        res.push(cb(a1[i], i));
+    }
+
+    return res;
 }
 
 export function arrayMap2<T1, T2, TR>(
     a1: Array<T1>,
     a2: Array<T2>,
     cb: (d1: T1, d2: T2, i: number) => TR
-) {
-    const mapArray = R.addIndex<T1, TR>(R.map);
+): Array<TR> {
+    const _a1_ = a1.length;
 
-    return mapArray((iValue, i) => cb(iValue, a2[i], i), a1);
+    const res: Array<TR> = [];
+
+    for (let i = 0; i < _a1_; ++i) {
+        res.push(cb(a1[i], a2[i], i));
+    }
+
+    return res;
 }
 
 export function arrayMap3<T1, T2, T3, TR>(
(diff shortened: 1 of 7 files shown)


src/dash-table/derived/style/index.ts
```diff
@@ -130,17 +130,18 @@ export const derivedTableStyle = memoizeOneFactory(
...
-            styles
-        )
-    );
+    let res: CSSProperties = {};
+
+    for (let i = 0; i < styles.length; ++i) {
+        Object.assign(res, styles[i].style);
+    }
+
+    return R.omit(BORDER_PROPERTIES_AND_FRAGMENTS, res);
```
Marc-Andre-Rivet: Makes no sense to make this on all styles.. just resolve once at the end

tslint.json
```diff
@@ -80,6 +80,7 @@
         ],
         "ordered-imports": false,
         "prefer-const": false,
+        "prefer-for-of": false,
```
Marc-Andre-Rivet: By default tslint want you to use `for(let v in vs)` instead of a basic for loop. It's significantly less performant. Disabling the warning.

src/core/math/arrayZipMap.ts
```diff
@@ -1,46 +1,67 @@
...
-    const mapArray = R.addIndex<T1, TR>(R.map);
+    fn: (d1: T1, i: number) => TR
+): Array<TR> {
+    const _a1_ = a1.length;
+
+    const res: Array<TR> = [];
 
-    return mapArray((iValue, i) => cb(iValue, i), a1);
+    for (let i = 0; i < _a1_; ++i) {
+        res.push(fn(a1[i], i));
```
alexcjohnson: As long as we're hyperoptimizing, I think `const res = new Array(_a1_)` and `res[i] = fn(a1[i], i);` is a bit faster?

Marc-Andre-Rivet: Yes!

src/dash-table/derived/edges/index.ts
```diff
@@ -1,27 +1,30 @@
...
-export const getHeaderCellEdges = (i: number, column: IVisibleColumn) => R.compose(resolveEdges, matchesHeaderCell(i, column));
-export const getHeaderOpCellEdges = (i: number) => R.compose(resolveEdges, getHeaderOpStyles(i));
\ No newline at end of file
+export const getDataCellEdges = (datum: Datum, i: number, column: IVisibleColumn) => (styles: IConvertedStyle[]) => resolveEdges(matchesDataCell(datum, i, column)(styles));
+export const getDataOpCellEdges = (datum: Datum, i: number) => (styles: IConvertedStyle[]) => resolveEdges(matchesDataOpCell(datum, i)(styles));
+export const getFilterCellEdges = (column: IVisibleColumn) => (styles: IConvertedStyle[]) => resolveEdges(matchesFilterCell(column)(styles));
+export const getFilterOpCellEdges = () => (styles: IConvertedStyle[]) => resolveEdges(getFilterOpStyles(styles));
+export const getHeaderCellEdges = (i: number, column: IVisibleColumn) => (styles: IConvertedStyle[]) => resolveEdges(matchesHeaderCell(i, column)(styles));
+export const getHeaderOpCellEdges = (i: number) => (styles: IConvertedStyle[]) => resolveEdges(getHeaderOpStyles(i)(styles));
```
alexcjohnson: ha good call - `compose` when you're not currying seems particularly silly to me.

PR #446, opened 2019-05-28
Refactor table API

Marc-Andre-Rivet:
Fixes https://github.com/plotly/dash-core/issues/58

See https://github.com/plotly/dash-core/issues/58#issuecomment-496593455 for detailed entries.

- [x] update prop types documentation
- [x] update changelog

PR #468, opened 2019-06-14
Regression fix multi table typing

Marc-Andre-Rivet:
#446 introduced a regression in the table sanitation logic causing a memoized functions to be shared across tables, resulting in an infinite loop re-rendering of all tables involve.

- fix the issue
- use two existing `edit` tests for the new 'two tables' scenario (w/o fix: https://circleci.com/gh/plotly/dash-table/9324, w/ fix: https://circleci.com/gh/plotly/dash-table/9333
- no changelog entry as this folds in with the API refactor / hasn't been published


demo/App.tsx
```diff
@@ -0,0 +1,84 @@
...
+            </div>);
+        } else if (mode === AppMode.TaleOfTwoTables) {
+            const props: any = {};
+            Object.entries(this.state.tableProps).forEach(([key, value]) => {
+                props[key] = this.propCache.get(key)(value);
+            });
+
+            return (<DataTable
+                {...props}
+            />);
```
Marc-Andre-Rivet: Two tables mode: shallow copy each prop & memoize -- feed the 2nd table with these new values.

demo/App.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { DataTable } from 'dash-table/index';
+import Environment from 'core/environment';
+import { memoizeOne } from 'core/memoizer';
+import Logger from 'core/Logger';
+import AppState, { AppMode } from './AppMode';
+import memoizerCache from 'core/cache/memoizer';
+
+import './style.less';
+
+class App extends Component<any, any> {
```
Marc-Andre-Rivet: Renaming to `.tsx` file but mostly just filling in with `any`. Can improve some more later. At least now it's TS too.

src/dash-table/dash/DataTable.js
```diff
@@ -8,21 +8,22 @@ import Logger from 'core/Logger';
...
+        this.sanitizer = new Sanitizer();
     }
 
     render() {
         if (!isValidProps(this.props)) {
             return (<div>Invalid props combination</div>);
         }
 
-        const sanitizedProps = sanitizeProps(this.props);
+        const sanitizedProps = this.sanitizer.sanitize(this.props);
```
Marc-Andre-Rivet: Each table instance has its sanitizer instance

src/dash-table/dash/Sanitizer.ts
```diff
@@ -64,16 +47,37 @@ const getFixedRows = (
...
+            columns: Sanitizer.applyDefaultsToColumns(locale_format, props.sort_as_null, props.columns, props.editable),
+            fixed_columns: getFixedColumns(props.fixed_columns, props.row_deletable, props.row_selectable),
+            fixed_rows: getFixedRows(props.fixed_rows, props.columns, props.filter_action),
+            locale_format
+        });
+    }
+
+    private static readonly applyDefaultToLocale = memoizeOne(applyDefaultToLocale);
+
+    private static readonly applyDefaultsToColumns = memoizeOne(applyDefaultsToColumns);
```
Marc-Andre-Rivet: `static readonly` on purpose to see the tests fail for this mode. This is essentially equivalent to the previous implementation. Will make `readonly` once a test run fails to demonstrate the impact.

tests/cypress/tests/standalone/edit_cell_test.ts
```diff
@@ -4,6 +4,34 @@ import Key from 'cypress/Key';
...
+        it('can edit when clicking outside of cell', () => {
+            DashTable.getCell(0, 1).click();
+            DOM.focused.type(`abc`);
+            cy.wait(1000);
+            DashTable.getCell(0, 0).click();
+            DashTable.getCell(0, 1).within(() => cy.get('.dash-cell-value').should('have.html', `abc`));
+        });
+    });
+});
+
```
Marc-Andre-Rivet: Subset of `edit` tests that are known to be impacted by the presence of a second table.

Pulled out from the tests below.

tests/cypress/tests/standalone/edit_cell_test.ts
```diff
@@ -4,6 +4,34 @@ import Key from 'cypress/Key';
...
+            DashTable.getCell(0, 1).click();
+            cy.wait(1000);
+            DOM.focused.type(`abc${Key.Enter}`);
+            DashTable.getCell(0, 1).within(() => cy.get('.dash-cell-value').should('have.html', `abc`));
+        });
+
+        it('can edit when clicking outside of cell', () => {
+            DashTable.getCell(0, 1).click();
+            DOM.focused.type(`abc`);
+            cy.wait(1000);
```
Marc-Andre-Rivet: The delay is not great but need to guarantee the table has had time to re-render if stuck in a loop.

alexcjohnson: Hmm OK. I guess if it's an async system causing the infinite loop that makes sense. Please put a comment by each of these waits referencing this PR so we know why it's there and we don't try to optimize it away later, since presumably the test will still pass if it were removed - it just wouldn't be testing what it's supposed to be testing.

PR #488, opened 2019-07-03
Handle external 'prop-types' correctly in prod build

Marc-Andre-Rivet:
This PR correctly externalizes `prop-types` for usage as a library. It also exposes the library as a prop on `window` to be consistent with other Dash libraries.

Would normally not update the build as part of a PR but this is included in another project through package.json and makes everything easier to handle on the other end.

`"dash-table": "github:plotly/dash-table#prod-prop-types"`


.config/webpack/base.js
```diff
@@ -19,29 +19,19 @@ module.exports = (options = {}) => {
...
-                root: 'React'
-            },
-            'react-dom': {
-                commonjs: 'react-dom',
-                commonjs2: 'react-dom',
-                amd: 'ReactDOM',
-                root: 'ReactDOM'
-            },
+            react: 'React',
+            'react-dom': 'ReactDOM',
```
Marc-Andre-Rivet: This can be simplified as it's target is `window` now

.config/webpack/base.js
```diff
@@ -19,29 +19,19 @@ module.exports = (options = {}) => {
...
-            demo: ['./demo/index.js', './demo/index.html']
+            demo: ['./demo/index.html', './demo/index.js']
         },
         mode: mode,
         output: {
             path: path.resolve(__dirname, `./../../${dashLibraryName}`),
             filename: '[name].js',
             library: dashLibraryName,
-            libraryTarget: 'umd'
+            libraryTarget: 'window'
```
Marc-Andre-Rivet: Loaded unto `window.dash_table` in the browser

wbrgss: Will changing how this is exposed have any side-effects for non-Dash users? Could you imagine anyone using RequireJS or anything else that expects a `umd` in their JavaScript project? I suppose, obviously, we wouldn't expect Node.js users (without a `window` global namespace) to be using this project. I'm also not sure if there's any webpack magic that makes this a non-issue.

Marc-Andre-Rivet: This should not have an impact.

Marc-Andre-Rivet:
@wbrgss Yes. It will fix the missing prop types dep

Marc-Andre-Rivet:
Adding a changelog entry and will merge afterwards.

Issue #495, opened 2019-07-10
Create a contextual menu for column headers


Marc-Andre-Rivet:
Do after #315 and #317 

Leave selection and sorting in the header itself and move rename, delete, clear into a contextual menu accessible through '...' or something similar.

PR #497, opened 2019-07-11
Issue 315: Clearable columns

### LINKED ISSUE #315: Ability to clear the contents of a column
- The UI for this feature may overload the existing `x` button on the column or it may introduce a new icon or an additional context menu.
- Deleting the contents of a column will delete the contents across all pages (if pagination is enabled)
- If a filtering expression is applied on a column of which the contents are deleted, the filter will be cleared as well.

Marc-Andre-Rivet:
Closes #315.

- [x] Update CHANGELOG with information about the feature and ref to this PR/issue
- [x] New tests for the feature
- [x] Visual tests for the column actions
- [ ] Documentation (`clearable` usage, action icon override through CSS & table css prop)

- Adds the ability to clear column (columns when merged) content without deleting the column itself.
- When the `clear` action is done, the associated filters are also cleared. The behavior of `delete` as been made the same as for `clear` with regard to the fitlers.
- some code rework to share filter state between FilterFactory and HeaderFactory now that header actions tie in with filtering

Re-using the currently in review changes done by @alinastarkov to have a different behavior based on whether columns are merged or not (https://github.com/plotly/dash-table/blob/f69106e5a9639ab1f7831f4cdabb0860bd1579f0/src/dash-table/utils/actions.js)


src/dash-table/derived/filter/map.ts
```diff
@@ -61,20 +61,49 @@ export default memoizeOneFactory((
...
     if (value && value.length) {
-        newMap.set(safeColumnId, new SingleColumnSyntaxTree(value, column));
+        newMap.set(id, new SingleColumnSyntaxTree(value, column));
     } else {
-        newMap.delete(safeColumnId);
+        newMap.delete(id);
     }
 
     return newMap;
+}
```
Marc-Andre-Rivet: Isolating the map update logic - can be called for a single column update (user edits filter) or through header `clear` and `delete` column actions (which may impact multiple columns at the same time)

src/dash-table/derived/filter/map.ts
```diff
@@ -61,20 +61,49 @@ export default memoizeOneFactory((
...
+    const asts = Array.from(map.values());
+    const globalFilter = getMultiColumnQueryString(asts);
+
+    const rawGlobalFilter = R.map(
+        ast => ast.query || '',
+        R.filter<SingleColumnSyntaxTree>(ast => Boolean(ast), asts)
+    ).join(' && ');
+
+    setFilter(globalFilter, rawGlobalFilter, map);
+}
```
Marc-Andre-Rivet: Same idea here, isolating the state update logic for both scenarios.

src/dash-table/derived/header/content.tsx
```diff
@@ -13,15 +13,40 @@ import {
...
+        }
+    }, actions.getAffectedColumns(column, columns, columnRowIndex));
 
-function deleteColumn(column: IVisibleColumn, columns: VisibleColumns, columnRowIndex: any, setProps: SetProps, data: Data) {
-    return () => {
-        setProps(actions.deleteColumn(column, columns, columnRowIndex, data));
-    };
-}
+    clearColumnsFilter(map, affectedColumns, setFilter);
+};
```
Marc-Andre-Rivet: Delete and clear apply the same logic, only the `action` changes

src/dash-table/derived/header/content.tsx
```diff
@@ -79,14 +104,22 @@ function getSortingIcon(columnId: ColumnId, sortBy: SortBy) {
     }
 }
 
+function getColumnFlag(i: number, flag?: boolean | boolean[]): boolean {
+    return typeof flag === 'boolean' ?
+        flag :
+        !!flag && flag[i];
+}
```
Marc-Andre-Rivet: renamable, clearable, deletable all use the same logic to decide whether the action is available

src/dash-table/derived/table/index.tsx
```diff
@@ -35,12 +35,15 @@ export default (propsFn: () => ControlledTableProps) => {
...
         const props = propsFn();
 
-        return filterPropsFn(propsFn, setFilter(props.setProps, props.setState));
-    });
-    const headerFactory = new HeaderFactory(propsFn);
+        return propsAndMapFn(propsFn, setFilter(props.setProps, props.setState));
+    };
+
+    const filterFactory = new FilterFactory(augmentedPropsFn);
+    const headerFactory = new HeaderFactory(augmentedPropsFn);
```
Marc-Andre-Rivet: headers now need access to map/setFilter too, generalize usage

Marc-Andre-Rivet:
@alexcjohnson Good idea, adding a few visual tests. Keeping in mind that we have https://github.com/plotly/dash-table/issues/495 as a follow up.

alexcjohnson:
Thanks for the visual tests!
I wonder if `ø` might look more “icon-like” than `Ø`
Also, a thought: what if we put `ø` and `×` into a css `::before` or something, so users could override these (without us having to add another prop 😏)?

Marc-Andre-Rivet:
Action "icons" can now be overridden through css or the table's `css` prop

```
css = [
    { selector: '.column-header--clear::before', rule: `content: 'C';` },
    { selector: '.column-header--delete::before', rule: `content: 'D';` },
    { selector: '.column-header--edit::before', rule: `content: 'E';` }
]
```

PR #498, opened 2019-07-11
Issue 491

### LINKED ISSUE #491: Changing one header cell changes all the header cells on the left
bug2

This only happened once when the cell is first changed.

alinastarkov:
Fixes #491


src/dash-table/utils/actions.js
```diff
@@ -1,12 +1,20 @@
 import * as R from 'ramda';
+import merge from 'ramda/es/merge';
```
Marc-Andre-Rivet: Use `R.merge` instead

src/dash-table/utils/actions.js
```diff
@@ -53,23 +61,36 @@ export const clearSelection = {
...
 };
 
-export function editColumnName(column, columns, headerRowIndex) {
+export function changeColumnHeader(column, columns, headerRowIndex, mergeDuplicateHeaders, newColumnName) {
+    let cloneColumn = Object.assign({}, column);
+    if ((typeof column.name === 'string') || (column.name instanceof String)) {
+            const columnHeaders = columns.map(col => col.name);
+            const maxLength = findMaxLength(columnHeaders);
+            const newColumnNames = Array(maxLength).fill(column.name);
+            cloneColumn = Object.assign({}, column, {name: newColumnNames});
```
Marc-Andre-Rivet: No need to transform into an array if there is only one header row.

src/dash-table/utils/actions.js
```diff
@@ -53,23 +61,36 @@ export const clearSelection = {
...
+    }
+    const transformedColumns = columns.map(col => {
+        if (((typeof column.name === 'string') || (column.name instanceof String)) && col.id === column.id) {
```
Marc-Andre-Rivet: `instanceof String` is only useful if we use the `new String(...)`, which we do not (guaranteed through tslint no-construct rule), `typeof` should be enough.

src/dash-table/utils/actions.js
```diff
@@ -1,12 +1,20 @@
...
 
-function getGroupedColumnIndices(column, columns, headerRowIndex) {
-    const columnIndex = columns.indexOf(column);
+function findMaxLength(array) {
+    let maxLength = 0;
+    R.forEach(row => { if (row instanceof Array && row.length > maxLength) {
+        maxLength = row.length;
+    }}, array);
+    return maxLength;
+}
```
Marc-Andre-Rivet: Use https://github.com/plotly/dash-table/blob/master/src/dash-table/derived/header/headerRows.ts instead.

src/dash-table/utils/actions.js
```diff
@@ -53,23 +61,36 @@ export const clearSelection = {
     selected_cells: []
 };
 
-export function editColumnName(column, columns, headerRowIndex) {
+export function changeColumnHeader(column, columns, headerRowIndex, mergeDuplicateHeaders, newColumnName) {
+    let cloneColumn = Object.assign({}, column);
```
Marc-Andre-Rivet: Use `R.merge` or `R.mergeAll` for consistency. The nice advantage is that it treats all arguments as immutables -- so you do not have to add a target `{ }`.

src/dash-table/utils/actions.js
```diff
@@ -53,23 +61,36 @@ export const clearSelection = {
...
+            const newColumnNames = Array(maxLength).fill(column.name);
+            cloneColumn = Object.assign({}, column, {name: newColumnNames});
+    }
+    const transformedColumns = columns.map(col => {
+        if (((typeof column.name === 'string') || (column.name instanceof String)) && col.id === column.id) {
+            return cloneColumn;
+        } else {
+            return col;
+        }
+    });
```
Marc-Andre-Rivet: Instead of a map you can:

newColumns = columns.slice(0) // shallow clone of array
newColumns[newColumns.indexOf(column)] = newColumn // replace the modified entry

tests/cypress/tests/standalone/edit_cell_test.ts
```diff
@@ -206,7 +206,58 @@ describe(`edit, mode=${AppMode.Typed}`, () => {
                 );
             });
         });
+
+    });
+});
+
+describe.only(`edit headers, mode=${AppMode.Typed}`, () => {
```
Marc-Andre-Rivet: I think a file for header tests is warranted.
Don't forget to remove the `.only` :)

src/dash-table/utils/actions.js
alexcjohnson: `R.merge` is deprecated - `mergeRight` is its more explicit replacement.

src/dash-table/utils/actions.js
```diff
@@ -53,23 +53,31 @@ export const clearSelection = {
...
+    const columnIndex = columns.findIndex(col => col.id === column.id);
+    transformedColumns[columnIndex] = cloneColumn;
     const { groupIndexFirst, groupIndexLast } = getGroupedColumnIndices(
-        column, columns, headerRowIndex
+        column, columns, headerRowIndex, mergeDuplicateHeaders
     );
-    /* eslint no-alert: 0 */
-    const newColumnName = window.prompt('Enter a new column name');
-    let newColumns = R.clone(columns);
+    let newColumns = R.clone(transformedColumns);
```
Marc-Andre-Rivet: I don't think the deep clone (https://ramdajs.com/docs/#clone) is necessary here. It will actually make multiple `memoizeOne` functions fail their test on their `column` argument, forcing a significant portion of the table to re-evaluate itself.

src/dash-table/utils/actions.js
```diff
@@ -53,23 +53,31 @@ export const clearSelection = {
     selected_cells: []
 };
 
-export function editColumnName(column, columns, headerRowIndex) {
+export function changeColumnHeader(column, columns, headerRowIndex, mergeDuplicateHeaders, newColumnName) {
+    let cloneColumn = R.mergeRight({}, column);
```
Marc-Andre-Rivet: There's no need to clone the column if it's not modified in the `if` below. `R.set` below (https://github.com/plotly/dash-table/pull/498/files#diff-4102a065e79334d16000051ec35189f7R75) will treat the entire data structure as immutable in any case.

src/dash-table/utils/actions.js
```diff
@@ -53,23 +53,31 @@ export const clearSelection = {
...
+    const maxLength = getHeaderRows(columns);
+    if (typeof column.name === 'string' && maxLength > 1) {
+            const newColumnNames = Array(maxLength).fill(column.name);
+            cloneColumn = R.mergeRight(column, {name: newColumnNames});
+    }
+    const transformedColumns =  columns.slice(0);
+    const columnIndex = columns.findIndex(col => col.id === column.id);
+    transformedColumns[columnIndex] = cloneColumn;
```
Marc-Andre-Rivet: Similarly, because of how `R.set` works, it's unnecessary to clone `columns` if the affected column above is not modified.

src/dash-table/utils/actions.js
Marc-Andre-Rivet: Interesting, that makes for **a lot** of `R.merge` usage to be updated in the table repo later :)

PR #502, opened 2019-07-15
add note about `fill_width` in 4.0.0 section

chriddyp:


chriddyp:
> Should we also update the migration docs page?

definitely

PR #507, opened 2019-07-18
Update dashTable to v4.0.2

rpkyle:
This PR proposes to update the `dashTable` package to v4.0.2, and adds examples to the R functions.


rpkyle:
> 💃 - Tricky tricky. Master is not 4.0.2 but rather 4.0.2 + `column.clearable`, when merging do not regenerate the code or you'll pick up non-4.0.2 stuff -- or do so and be misaligned!

As we discussed, I did check out the 4.0.2 tag and build from there, so this package should be ready for release. I'll test locally and then update the commit hash in the remotes for `plotly/dashR` and the hard version locking on `4.0.0` right after pulling the trigger.

Issue #490, opened 2019-07-04
Column is automatically merged when merge_duplicate_headers  = true even if the names are the same even if the data is different


alinastarkov:
![bug1](https://user-images.githubusercontent.com/23650639/60670674-3e7cb800-9e3f-11e9-8973-92e6c66b49f1.gif)

and the data in demo/data.ts is:
```
...
 {
        id: 'ccc',
        name: ['City', 'Canada', 'Toronto'],
        type: ColumnType.Numeric,
        data: gendata(i => i, rows)
    },

    {
        id: 'ddd',
        name: ['City', 'Canada', 'Toronto'],
        type: ColumnType.Numeric,
        data: gendata(i => i * 100, rows)
    },
...
```

alexcjohnson:
Nice find @alinastarkov!

Is there a use case for two columns with entirely merged headers? To simplify that, is there a use case for merging single-row headers? I guess I can imagine cases I might want to do it - say I have 2 columns for each year, and it's clear from the values in each column what they mean, like the first column is sales volume and the second is percent change from the previous year. It would preclude single-column operations like sorting, but that doesn't seem like a good reason to forbid it. This bug might be though, if we can't figure out a workaround.

PR #508, opened 2019-07-18
Issue 314 - Hidden Columns

Marc-Andre-Rivet:
Hidden columns feature #314 

Support to toggle columns visibility.

**API modifications**
- added `column.hideable: boolean | undefined`
- added `hidden_columns: string[] | undefined`

This proposed solution deviates from the one describe in the parent issue in a few ways.
- `hideable` only supports `boolean`: supporting `boolean[]` makes it hard to give a consistent experience when toggling visibility while supporting making non-hideable but hidden columns visible if necessary. It also requires more complex UI interactions that are not obvious to define correctly.. also deviates from more simple column visibility ideas usually associated with spreadsheets
- the on-hover UI that shows a dropdown / marker for hidden columns interacts badly with the existing actions inside the header cells

So,
- columns can only be hidden one by one
- with multiple header rows, the last row is the one with the `hide` action
- making columns visible again is done through a dropdown / toggle columns button / menu that closes when the user clicks on something else than the menu itself

While not ideal this solution should be flexible enough to make it possible for us to improve upon while not limiting our options.

-----------------------------------

On a totally different note, with the addition of multiple actions in the header cells, the use of unicode characters is starting to be problematic as they hardly ever align correctly or feel consistent. With that in mind, introducing `fortawesome`, a JS / es6-aware (read: tree shaking) wrapper around Font-Awesome free icon library to be used for the actions.

image

One of the goal of a previous PR was to make it easy for users to override the default icon with an icon of their choice. This is still possible with the proposed solution, but the selector(s) will be slightly different. As (a) the modified code w/ customization support has not been release yet and (b) this must be done through unofficial `css` props or custom stylesheet, this seems fine.


dash_table/index.html
```diff
@@ -5,9 +5,8 @@
     </head>
     <body>
         <div id='root'></div>
-        
-        <script crossorigin src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
-        <script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.production.min.js"></script>
+        <script src='https://unpkg.com/react@16.8.6/umd/react.production.min.js'></script>
+        <script src='https://unpkg.com/react-dom@16.8.6/umd/react-dom.production.min.js'></script>
```
Marc-Andre-Rivet: Slighty tighter versioning -- we know Dash targets a specific version of React, no need to add potential instability through testing with the latest.

demo/AppMode.ts
```diff
@@ -7,15 +7,14 @@ import {
...
     ColumnType,
     TableAction
 } from 'dash-table/components/Table/props';
 import { TooltipSyntax } from 'dash-table/tooltips/props';
 
 export enum AppMode {
-    Clearable = 'clearable',
-    ClearableMerged = 'clearableMerged',
+    Actionable = 'actionable',
+    ActionableMerged = 'actionableMerged',
```
Marc-Andre-Rivet: Modified `clearable` modes to display all actions instead

package.json
```diff
@@ -57,6 +57,10 @@
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
     "@cypress/webpack-preprocessor": "^4.1.0",
+    "@fortawesome/fontawesome-svg-core": "^1.2.19",
+    "@fortawesome/free-regular-svg-icons": "^5.9.0",
+    "@fortawesome/free-solid-svg-icons": "^5.9.0",
+    "@fortawesome/react-fontawesome": "^0.1.4",
```
Marc-Andre-Rivet: The core FA library for JS / tree-shaking, the free icons and the React wrapper.
The original implementation used CSS pseudo-elements instead of the React wrappers but the table / FA core interaction for actively checking pseudo-elements degraded the table's performance significantly (~300% increase in render time)

src/core/storage/Cookie.ts
```diff
@@ -28,7 +28,7 @@ export default class CookieStorage {
     }
 
     public static set(id: string, value: string, domain: string = '', path: string = '/') {
-        let expires = new Date((new Date().getTime() + __20years)).toUTCString();
+        let expires = new Date(Date.now() + __20years).toUTCString();
```
Marc-Andre-Rivet: Unrelated cleanup that was done while the issue still included persistence. Keeping this small change around.

src/dash-table/components/CellFactory.tsx
```diff
@@ -51,7 +50,8 @@ export default class CellFactory {
             style_cell_conditional,
             style_data,
             style_data_conditional,
-            virtualized
+            virtualized,
+            visibleColumns
```
Marc-Andre-Rivet: `columns` became `visibleColumns` pretty much everywhere under the sanitation layer. Updating usage / destructuring-ordering accordingly.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -183,8 +184,23 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+        const menu = this.menuRef;
+
+        if (this.props.activeMenu &&
+            menu &&
+            menu.current &&
+            !menu.current.contains(event.target as Node)
+        ) {
+            this.props.setState({
+                activeMenu: undefined
+            });
```
Marc-Andre-Rivet: Generalized the `click-outside` handler. Now checks for two different / unrelated cases. (a) table's lost of focus, (b) closing the columns menu

Marc-Andre-Rivet: Using the new way of `ref` in React 16

chriddyp:
can you share some screenshots of what this looks like or update the review app with an example we can play with?

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -777,6 +798,10 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
                 className='dash-table-tooltip'
                 tooltip={tableTooltip}
             />
+            <div className='dash-spreadsheet-menu'>
+                {this.renderMenu()}
+                <ExportButton {...buttonProps} />
+            </div>
```
Marc-Andre-Rivet: Nothing permanent but grouping the toggle and the export buttons.

src/dash-table/components/Table/Table.less
```diff
@@ -485,21 +517,10 @@
     }
 
     .dash-spreadsheet-inner {
-        .column-header--clear::before {
-            content: 'ø';
-        }
```
Marc-Andre-Rivet: Changed for actual icons

src/dash-table/components/Table/controlledPropsHelper.ts
```diff
@@ -0,0 +1,106 @@
+import * as R from 'ramda';
```
Marc-Andre-Rivet: Externalizing controlled props logic from the component.

src/dash-table/components/Table/derivedPropsHelper.ts
```diff
@@ -0,0 +1,107 @@
+import * as R from 'ramda';
```
Marc-Andre-Rivet: Externalizing derived props logic from the component.

src/dash-table/components/Table/style.ts
```diff
@@ -0,0 +1,13 @@
...
+
+library.add(
+    faEraser,
+    faEyeSlash,
+    faPencilAlt,
+    faSort,
+    faSortDown,
+    faSortUp,
+    faTrashAlt
+);
```
Marc-Andre-Rivet: Loads font-awesome icons in a way that allows for tree-shaking. The downside is that this file needs to be updated when new icons are needed.

src/dash-table/dash/Sanitizer.ts
```diff
@@ -62,27 +62,39 @@ const applyDefaultsToColumns = (defaultLocale: INumberLocale, defaultSort: SortA
...
+
         return R.merge(props, {
-            columns: this.applyDefaultsToColumns(locale_format, props.sort_as_null, props.columns, props.editable),
+            columns,
+            export_headers: headerFormat,
             fixed_columns: getFixedColumns(props.fixed_columns, props.row_deletable, props.row_selectable),
             fixed_rows: getFixedRows(props.fixed_rows, props.columns, props.filter_action),
             locale_format,
-            export_headers: headerFormat
+            visibleColumns
```
Marc-Andre-Rivet: New `visibleColumns` for the TS land -- hidden columns are removed. This is the prop used for most of everything in the table now. `columns` is still used but for specific scenarios.

src/dash-table/derived/header/content.tsx
```diff
@@ -134,42 +134,66 @@ function getter(
...
+                    const renamable = getColumnFlag(headerRowIndex, column.renamable);
 
                     return (<div>
                         {sort_action !== TableAction.None && isLastRow ?
                             (<span
                                 className='sort'
                                 onClick={doSort(column.id, sortBy, mode, setProps)}
                             >
-                                {getSortingIcon(column.id, sortBy)}
+                                <FontAwesomeIcon icon={getSortingIcon(column.id, sortBy)} />
```
Marc-Andre-Rivet: Usage for font-awesome icons with the React wrapper.

tests/visual/percy-storybook/DashTable.percy.tsx
```diff
@@ -200,38 +200,34 @@ const sparseData = (() => {
...
+const hiddenColumns = columnsA2J.map(c => c.id).filter((_, index) => index % 2 === 0);
 
 storiesOf('DashTable/Hidden Columns', module)
     .add('hides', () => (<DataTable
         setProps={setProps}
         id='table'
         data={dataA2J}
-        columns={hiddenColumns}
+        columns={columnsA2J}
+        hidden_columns={hiddenColumns}
```
Marc-Andre-Rivet: These changes were overlooked in the 4.0 work -- `hidden` wasn't supported anymore. Now changed to use `hidden_columns` instead

src/dash-table/components/CellFactory.tsx
Marc-Andre-Rivet: columns -> visibleColumns
IVisibleColumn -> IColumn
etc. on that theme

chriddyp:
> making columns visible again is done through a dropdown / toggle columns button / menu 

could you share a gif of what this interaction looks like?

src/dash-table/dash/DataTable.js
```diff
@@ -166,6 +166,8 @@ export const propTypes = {
          */
         editable: PropTypes.bool,
 
+        hideable: PropTypes.bool,
```
alinastarkov: missing props description

src/dash-table/dash/DataTable.js
```diff
@@ -460,6 +462,8 @@ export const propTypes = {
      */
     fill_width: PropTypes.bool,
 
+    hidden_columns: PropTypes.arrayOf(PropTypes.string),
```
alinastarkov: idem

Marc-Andre-Rivet:
toggle

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+[#314](https://github.com/plotly/dash-table/issues/314)
+- New `column.hideable` flag that displays a "eye" action icon in the column
```
alexcjohnson:
```suggestion
- New `column.hideable` flag that displays an "eye" action icon in the column
```

dash_table/index.html
alexcjohnson: Is there any (easy) way to tie this explicitly to the version in dash? Or perhaps just a test that they're the same, so if we change it in dash we'll be prompted to update it here?

src/dash-table/components/ControlledTable/index.tsx
alexcjohnson: neat, wasn't aware of `React.createRef` ✨

alexcjohnson: Do we want menu clicks to drop focus on the table? Looks like that's what this will do, but feels to me like the focus should remain.

alexcjohnson:
What happens if there are 100 columns? Does (or could) the menu start to scroll at some point?

src/dash-table/components/Table/props.ts
```diff
@@ -234,6 +230,7 @@ export interface IUSerInterfaceTooltip {
 }
 
 export interface IState {
+    activeMenu?: 'show/hide';
```
alexcjohnson: The idea I guess is we may later add other menus so this would become an enum?

src/dash-table/dash/DataTable.js
```diff
@@ -166,6 +166,18 @@ export const propTypes = {
...
+         * If True, then the column is hideable through the UI.
+         * If there are multiple column headers, only the last row of
+         * the column headers will dispay the `hide` action button.
+         *
+         * Hidden columns are listed by id in the `hidden_columns` prop.
+         *
+         * If `hidden_columns` is not empty, a toggle button is displayed
+         * above the table that allows toggling columns visibility.
+         */
+        hideable: PropTypes.bool,
```
alexcjohnson: Was briefly considering whether we want a global `hideable` or `columns_hideable` prop to set the default for all columns. I think perhaps not, as hiding *all* columns would be... bad. So we should encourage folks to think about which columns to allow hiding by not allowing the simple option of "hide them all."

Alternatively, if we have logic to prevent hiding the only remaining column (do we? 🙏) maybe a "hide them all" option wouldn't be so bad. What do you think?

alexcjohnson:
Question (to which I don't have a "right" answer in mind, just curious): if I'm merging headers, and have headers `a | b | a` and I hide column `b`, do the two `a`'s merge?

dash_table/index.html
Marc-Andre-Rivet: Interesting. Nothing I can think up right now. Opening a follow up story.

Marc-Andre-Rivet: https://github.com/plotly/dash-table/issues/522

src/dash-table/components/Table/props.ts
Marc-Andre-Rivet: Yes. This is all highly experimental and probably won't survive the coming round of UI rework for the column ops.. but, yes.

Marc-Andre-Rivet:
> What happens if there are 100 columns? Does (or could) the menu start to scroll at some point?

It would have been a very large menu. Limited the size and improved the menu's styling (on my machine at home the column names were wrapping up for some reason).

https://github.com/plotly/dash-table/pull/508/commits/342849fe92c6a5bbbf3867bf617f49b331b9d3f7

> Question (to which I don't have a "right" answer in mind, just curious): if I'm merging headers, and have headers a | b | a and I hide column b, do the two a's merge?

Interestingly enough this is something similar to what @alinastarkov came across a few weeks ago: https://github.com/plotly/dash-table/issues/490 -- as is, I'm certain the table will merge those together. I'd say it shouldn't, if only because `merge` is driven by columns but probably shouldn't be driven by `visible` columns. Same for other edge cases where say `[A, B]` and `[C, B]` would see the top row as unmerged and the second row merged. 

> Alternatively, if we have logic to prevent hiding the only remaining column (do we? pray) maybe a "hide them all" option wouldn't be so bad. What do you think?

Actually haven't considered that case. The entire table sort of disappears when there's no columns.. a special problem of its own. For now, I'll add logic (and a test) to disable hiding columns if there is only one left.

Marc-Andre-Rivet:
Also, it needs to be possible to hide/show merged columns instead of just the last row. Maybe keeping `column.hideable` + the table level `hideable` + some `hideable_row` which takes an integer would be acceptable. The menu would be populated with the names in that row. That last one with a `beta` flag of some sort. There has to be a better way to implement this.

src/dash-table/components/ControlledTable/index.tsx
Marc-Andre-Rivet: The focus remains. If you click to open the menu, you can still type / navigate the table and the `focused` cell highlight isn't lost. `onBlur` will happen though as you clicked outside the current cell.

src/dash-table/dash/DataTable.js
```diff
@@ -434,6 +441,26 @@ export const propTypes = {
...
+     */
+    hideable: PropTypes.bool,
+
+    /**
+     * 'beta'. Define the header row on which the hideable action will be
+     * visible.
+     *
+     * Defaults to the last header row.
+     */
+    hideable_row: PropTypes.number,
```
alexcjohnson: why not continue the pattern from `clearable`, `deletable`, and `renamable` of supporting an array of bools?

Marc-Andre-Rivet: The main issue I've been having is with the reverse operation (toggling the column back into the table) -- these other operations do not have this reverse.

With the toggle menu, I find that the menu would be messy if it had to show all combinations / options of merged cells based on various header rows that could be used to toggle them.

Instead, this prop forces the `hide` action to exist on a single header row. Making it possible to show items in the toggle menu that are all at the same level.

I'm aware that this prop may not represent the best solution in the long termand that's why I've taken the liberty to tag it as `beta`.

alexcjohnson: also: mention these new props in the changelog

alexcjohnson: Personally I don't see exact parity between the two toggle mechanisms as being important - or perhaps even desirable. I feel like the menu should list all columns and allow toggling each independently, even if the hideable button in the header itself is placed in a merged cell and hides all those columns together.

Also, not sure if we had this discussion before or not, but seems to me like a single `true` in any of these props would be better as "just the last row", and if someone really wants all rows to get the buttons they would need to provide `[true, true, true]`. Is that too much of a breaking change to discuss making now?

Marc-Andre-Rivet: No no. Let's discuss this.

alexcjohnson: OK, after offline discussion for context, we agreed (I think, correct me if I'm missing a piece @Marc-Andre-Rivet):
- `hideable` can be a bool or array of bools, plus (and all similar props get this option too) `'last'` to only apply to the last header row. That way it's easy to get this behavior without needing to create an array `[false, false, false, true]` but it won't break the existing `true` that turns on all rows.
- The hide/show menu gets a list corresponding to all possible hideable buttons. Each entry is the concatenation of all header rows above and including the one the button is in, ordered by column index then row index. This may look weird if you've shown buttons in multiple rows, but we're OK with that because the behavior is weird in that case too, and the user was explicit about enabling this weirdness.
- The behavior of this menu follows the button on screen: ie if you have a merged header and *any* of the merged columns is visible (even if you've hidden some of them using a lower down hide button) then the checkbox is checked, and unchecking it hides all remaining columns. After that, checking the box to re-show will reshow all merged columns, it will not attempt to invert to the partially hidden state.

Marc-Andre-Rivet:
> `hideable` can be a bool or array of bools, plus (and all similar props get this option too) 'last' to only apply to the last header row. 

Similar props are: clearable, deletable, hideable, renamable. All of them now have `last` and `first`.

- Staying consistent with the above, removed the table-level `hideable` prop as it does not have an equivalent.

- The toggle menu shows all available options that can be toggled on/off and displays them in order of column_index, then row_index.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -806,7 +837,67 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+                                />
+                                <label>{!column.name ?
+                                    column.id :
+                                    typeof column.name === 'string' ?
+                                        column.name :
+                                        column.name.slice(0, i + 1).filter(name => name.length !== 0).join(' | ')
+                                }</label>
+                            </div>)
+                        };
+                    }))).filter(i => !R.isNil(i)).sort((a, b) => a.j - b.j).sort((a, b) => a.i - b.i).map(a => a.component)}
```
Marc-Andre-Rivet: This is a bit messy -- can rework if seen as problematic.

Generates a sparse component[][] containing the list of inputs but keeping their associated index, remove nil items and sort in reverse order of condition/priority (columns are more important than rows -- ordering is like a radix sort), retrieve the component from the wrapper.

src/dash-table/components/Table/props.ts
```diff
@@ -275,6 +272,8 @@ export interface IProps {
     fill_width?: boolean;
     filter_query?: string;
     filter_action?: TableAction;
+    hideable?: boolean | boolean[];
```
Marc-Andre-Rivet: removing

src/dash-table/components/Table/props.ts
```diff
@@ -275,6 +272,8 @@ export interface IProps {
...
+    hidden_columns?: string[];
```
Marc-Andre-Rivet: removing

src/dash-table/components/Table/props.ts
```diff
@@ -321,6 +320,7 @@ interface IDefaultProps {
     fill_width: boolean;
     filter_query: string;
     filter_action: TableAction;
+    hideable: boolean | boolean[];
```
Marc-Andre-Rivet: removing

src/dash-table/derived/header/columnFlag.ts
```diff
@@ -0,0 +1,11 @@
...
+    i: number,
+    last: number,
+    flag?: boolean | boolean[] | 'first' | 'last'
+): boolean => flag === 'last' ?
+        i === last :
+        flag === 'first' ?
+            i === 0 :
+            typeof flag === 'boolean' ?
+                flag :
+                !!flag && flag[i];
```
Marc-Andre-Rivet: This piece of logic is needed in lots of places now

src/dash-table/components/ControlledTable/index.tsx
alexcjohnson: `.sort((a, b) => (a.i - b.i) || (a.j - b.j))` ?

Marc-Andre-Rivet: Right. That should work too. Checking / updating.

src/dash-table/dash/DataTable.js
```diff
@@ -168,34 +185,51 @@ export const propTypes = {
...
+         * If `last`, the `rename` action button will only appear on the last header
+         * row. If `first` it will only appear on the first header row. These
+         * are respectively shortcut equivalents to `[False, ..., True]` and
+         * `[True, ..., False]`.
+         *
+         * If there are merged, multi-header columns then you can choose
+         * which column header row to display the `rename` action button in by
+         * supplying an array of booleans.
+         * For example, `[true, false]` will display the `rename` action button
+         * on the first row, but not the second row.
```
alexcjohnson: Are we using Python (`True`) or JS (`true`) capitalization here? This inconsistency exists all sorts of places across our component repos, but IIRC we said we'd try to standardize on JS in the JS code, paving the way for language-dependent translation at some future time in the back-end codegen.

Also `[false, ..., true]` is ambiguous. Perhaps `[false, ..., false, true]`? That said I might just remove the shortcut sentence, as it's not a fixed shortcut (depends on # of headers) and I think it's already pretty clear without that.

Marc-Andre-Rivet: Will improve consistency for the modified props and go the JS route. Good point for the array examples.

PR #523, opened 2019-08-01
Copy paste headers

### LINKED ISSUE #318: ability to include column headers when copying-and-pasting
Currently, when you copy and paste contents from the DataTable into Excel, only the currently selected contents are included. This item would enable the copy-and-paste to include the column headers when pasting the data.
This setting would be set by the Dash developer through a property like “`include_headers_on_copy_and_paste=True`”.

Note that this would be independent of the rows or columns (#317) that have been selected

alinastarkov:
Closes #318


src/dash-table/utils/TableClipboardHelper.ts
```diff
@@ -5,15 +5,20 @@ import Clipboard from 'core/Clipboard';
...
 
 export default class TableClipboardHelper {
     private static lastLocalCopy: any[][] = [[]];
+    private static localCopyWithoutHeaders: any[][] = [[]];
 
     public static toClipboard(e: any, selectedCells: SelectedCells, columns: Columns, data: Data) {
         const selectedRows = R.uniq(R.pluck('row', selectedCells).sort((a, b) => a - b));
         const selectedCols: any = R.uniq(R.pluck('column', selectedCells).sort((a, b) => a - b));
 
+        const transposedHeaders = createHeadings(R.pluck('name', columns), getHeaderRows(columns));
```
Marc-Andre-Rivet: When merging in with the hidden columns PR #508, will need to pass both the columns and visibleColumns -- visibleColumns for processing the data / creating the copy/paste info and columns (containing the hidden ones) to determine the number of header rows correctly

PR #525, opened 2019-08-02
Fix sticky behavior in Safari

Marc-Andre-Rivet:
@alinastarkov found `-webkit-sticky`, just adding it and fixing a (would be) regression on fixed columns (when transparent)

### DIFF
--- a/src/dash-table/components/Table/Table.less
+++ b/src/dash-table/components/Table/Table.less
@@ -257,6 +257,7 @@
                 flex: 0 0 auto;
                 left: 0;
                 position: sticky;
+                position:-webkit-sticky;
                 z-index: 400;
             }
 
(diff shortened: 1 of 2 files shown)


src/dash-table/components/Table/Table.less
```diff
@@ -436,6 +437,10 @@
         background-color: rgb(250, 250, 250);
     }
 
+    .dash-spreadsheet-inner td {
+        background-color: white;
+    }
```
alexcjohnson: Does this have a visual impact?

Marc-Andre-Rivet: No. A style was applied on `tr` and was removed which caused issues with fixed columns (columns content was visible as it was scrolled under the fixed ones). Adding it back on `td` instead.. just because that way it's on cells.. and since everything is otherwise applied on cells, ever so slightly cleaner. There's no real difference between applying that style on tr or td in the end.

PR #534, opened 2019-08-06
Remove newlines and lists from docstrings

shammamah:
The newlines and lists in the docstrings were making it difficult to read in the Python console; they have been replaced with semicolons/periods.


alexcjohnson:
This looks good to me, if it'll let us proceed with https://github.com/plotly/dash-docs/pull/479. Will wait for @Marc-Andre-Rivet to chime in though with any concerns.

Issue #447, opened 2019-05-28
Error selecting cell in virtualized table


kottmast:
Clicking to select a cell that is "below the fold" throws: `TypeError: Cannot read property '1' of undefined`, where '1' is the column index of the cell that is clicked.

Simple example...

```python
dash_table.DataTable(
            id='table',
            data=df.to_dict('records'),
            columns=[{'id': c, 'name': c} for c in df.columns],
            editable=False,
            filtering=True,
            sorting=True,
            sorting_type="multi",
            pagination_mode=False,
            virtualization=True
)
```

When the data is loaded and the table is displayed, clicking a cell highlights that cell to indicate it is selected. If you then scroll down and click to select a new cell, the error is thrown. Full stack trace:

```
react-dom@16.8.6.min.js?v=0.24.0&m=1554834449:117 TypeError: Cannot read property '1' of undefined
    at e.value (type.ts:35)
    at e.value (type.ts:83)
    at data.ts:32
    at e.getDataEdges (memoizer.ts:7)
    at EdgeFactory.tsx:28
    at e.memoizedCreateEdges (memoizer.ts:7)
    at e.value (EdgeFactory.tsx:138)
    at index.tsx:18
    at t.value (index.tsx:458)
    at ce (react-dom@16.8.6.min.js?v=0.24.0&m=1554834449:98)
```

The property it tries to read appears to be the column index of the cell that was clicked, ie. if you click in the 5th column the error will be "Cannot read property '4' of undefined".

After that the table either disappears or becomes unresponsive. I discovered this while trying to add a callback to do something with the selected cells. There is no apparent server-side error and the callback executes successfully, so the error seems to happen after the callback is executed.

Versions:
dash 0.43.0
dash-table 3.7.0
Chrome 74.0

kottmast:
I just downgraded to dash_table==3.6.0 and do not see the error so it seems to be a regression introduced in 3.7.0.

Marc-Andre-Rivet:
@kottmast Thanks for reporting this. Investigating.

vivekvs1:
Having a similar error but for a different scenario. Assigning 'data' for a table based on selected_data of another table is throwing the same error. Like @kottmast has mentioned, it works fine when I downgraded to  dash_table==3.6.0

wvwhome:
Also error: ⛑️
Cannot read property '2' of undefined
with 3.7.0.  but I'm also using 'row_id' which is only in 3.7.0.  Stuck.

[...]

PR #535, opened 2019-08-06
Add test for selection of cells outside of the viewport.

### LINKED ISSUE #520: Virtualized table error when navigating past the lower edge of the viewport
Issue first described in https://github.com/plotly/dash-table/issues/447.

The bug seems fixed in `dash==1.0.2` but that can only have happened by random luck.

There's no actual fix to be done here, just adding a test to make sure navigating with the keyboard past the edge of the viewport (top or bottom) does not trigger a console error / behaves correctly.

shammamah:
Closes #520.

## About
* Added two tests to replicate the behaviour described in #447.
* Changed `data-dash-row` to point to the actual row in the data as opposed to the row number in the viewport.
* Added new functions to scroll the table all the way to the bottom/top.


src/dash-table/derived/cell/wrappers.tsx
```diff
@@ -28,7 +28,7 @@ class Wrappers {
             (column, columnIndex) => this.getWrapper(
                 false,
                 false,
-                rowIndex,
+                rowIndex + _offset.rows,
```
Marc-Andre-Rivet: You can rename `_offset` to `offset`. The underscore was there to prevent tslint from flagging it as an unused parameter since it was required for caching purposes but not when calculating.

tests/cypress/src/DashTable.ts
```diff
@@ -61,6 +61,22 @@ export default class DashTable {
         return cy.get(`#table tbody td.cell--selected`);
     }
 
+    static scrollFullTableHeight() {
+        cy.get(`.row.row-1`).scrollTo(0, 10000);
+    }
```
Marc-Andre-Rivet: Maybe scroll to top and scroll to bottom functions instead?

tests/cypress/tests/standalone/navigation_test.ts
```diff
@@ -83,6 +83,13 @@ Object.values(BasicModes).forEach(mode => {
                 DashTable.getCell(2, 1).should('have.class', 'focused');
                 DashTable.getCell(3, 1).should('not.have.class', 'focused');
             });
+
+            it('can move out of the viewport in virtualized mode', () => {
+                DashTable.toggleScroll(true);
+                for (let i = 0; i < 25; i++) {
+                    DOM.focused.type(Key.ArrowDown);
+                }
+            });
```
Marc-Andre-Rivet: Nothing specific is being validated here. If the down arrow causes an error and/or does not update the selection the test will not fail.

shammamah: I tested this with the version of the table that had this error, and the test did end up failing, since there is a console error (which I thought would be implicitly included in the test failure conditions). Is there another way to catch console errors?

tests/cypress/tests/standalone/navigation_test.ts
```diff
@@ -100,6 +107,14 @@ Object.values(BasicModes).forEach(mode => {
                 DashTable.getCell(4, 2).should('have.class', 'focused');
                 DashTable.getCell(3, 1).should('not.have.class', 'focused');
             });
+
+            it('can select a cell outside of the viewport', () => {
```
Marc-Andre-Rivet: I think the test name could be improved. It's not about selecting a cell outside of viewport but rather about selecting a cell and scrolling it out of the viewport.

Like for the test above, would be interested in checking if there are console errors triggered by these steps. Maybe something like: https://github.com/cypress-io/cypress/issues/300#issuecomment-513514236

tests/cypress/tests/standalone/navigation_test.ts
Marc-Andre-Rivet: Ah! You're right. Unhandled errors cause the tests to fail by default - did not realize that was the case.
https://docs.cypress.io/api/events/catalog-of-events.html#To-turn-off-all-uncaught-exception-handling

Marc-Andre-Rivet: Would still like the test do more validation. The implementation could change and the error be swallowed instead of being unhandled. The test would not fail in that case.

tests/cypress/tests/standalone/navigation_test.ts
Marc-Andre-Rivet: @shammamah From your comment in https://github.com/plotly/dash-table/pull/535/files#r315342988, this is moot.

tests/cypress/src/DashTable.ts
```diff
@@ -61,6 +61,30 @@ export default class DashTable {
         return cy.get(`#table tbody td.cell--selected`);
     }
 
+    static scrollToTop() {
+        cy.get(`.cell.cell-1-1.dash-fixed-content`).invoke(`outerHeight`).then(height => {
+            cy.scrollTo(0, -1 * height);
```
Marc-Andre-Rivet: Minor, but you can do this instead:

```suggestion
            cy.scrollTo(0, -height);
```

PR #538, opened 2019-08-08
Issue 317 - selectable columns

Marc-Andre-Rivet:
Adding selectable columns support to the table with API similar to that of selectable rows.

* Columns get new nested prop `selectable: 'last' | 'first' | boolean | boolean[]` like other column actions.
* Table gets new prop `column_selectable: 'single' | 'multi' | false`
* Table gets a new prop `selected_columns` with the selected columns (hidden or otherwise)
* Table gets a new prop `derived_viewport_selected_columns` with the selected columns (non-hidden)

Missing
* [x] docstrings
* [x] changelog entry


src/dash-table/utils/actions.js
```diff
@@ -1,13 +1,27 @@
...
+            const c = columns[i];
+
+            if (c.name && Array.isArray(c.name) && c.name.length > headerRowIndex && c.name[headerRowIndex] === column.name[headerRowIndex]) {
+                columnIndex = i;
+            } else {
+                break;
+            }
+        }
+    }
+
```
Marc-Andre-Rivet: Backward looking to pick up all columns hidden or not.

tests/cypress/tests/standalone/column_test.ts
```diff
@@ -83,6 +83,158 @@ describe(`column, mode=${AppMode.ActionableMerged}`, () => {
...
+        DashTable.getSelectColumnById(1, 'ccc').should('not.be.checked');
+        DashTable.getSelectColumnById(2, 'ccc').should('not.be.checked');
+        DashTable.getSelectColumnById(0, 'ddd').should('not.be.checked');
+        DashTable.getSelectColumnById(0, 'eee').should('not.be.checked');
+        DashTable.getSelectColumnById(1, 'eee').should('not.be.checked');
+        DashTable.getSelectColumnById(0, 'fff').should('not.be.checked');
+        DashTable.getSelectColumnById(0, 'ggg').should('not.be.checked');
+        DashTable.getSelectColumnById(1, 'ggg').should('not.be.checked');
+    });
+});
```
Marc-Andre-Rivet: A few tests for single select and multi select cases

demo/AppMode.ts
```diff
@@ -399,4 +347,20 @@ function getState() {
...
+    const flavorParam = Environment.searchParams.get('flavor');
+    const flavors = flavorParam ? flavorParam.split(';') : [];
+
+    let state = getModeState(mode);
+    flavors.forEach(flavor => {
+        const [key, valueString] = flavor.split('=');
+        const value = JSON.parse(valueString);
+
+        (state.tableProps as any)[key] = value;
+    });
```
alexcjohnson: oh that's cool, so now I can tweak arbitrary props in the demo just by adding them to the url as `&flavor=...`? 🎉

demo/AppMode.ts
```diff
@@ -71,16 +78,16 @@ function getBaseTableProps(mock: IDataMock) {
...
-            max_width: '1000px',
+            maxWidth: '1000px',
             width: '1000px'
         },
         style_cell_conditional: [
-            { max_width: 150, min_width: 150, width: 150 },
-            { if: { column_id: 'rows' }, max_width: 60, min_width: 60, width: 60 },
-            { if: { column_id: 'bbb' }, max_width: 200, min_width: 200, width: 200 },
-            { if: { column_id: 'bbb-readonly' }, max_width: 200, min_width: 200, width: 200 }
+            { if: { }, maxWidth: 150, minWidth: 150, width: 150 },
```
alexcjohnson: that looks weird, but I guess it's valid... is there a reason to put it here rather than `style_cell`?

demo/AppMode.ts
Marc-Andre-Rivet: Yes. I've been using a "mode" param for a while but as more tests get added there's a lot of cases where we just want the same configuration + merging or + another small diff in the setup. Hence the "flavor" The syntax is a bit tricky at times because of the params encoding and since the prop's rhs must be json parseable.

demo/AppMode.ts
Marc-Andre-Rivet: nope, moving it

.circleci/config.yml
```diff
@@ -109,6 +110,7 @@ jobs:
                     pip install -r dev-requirements.txt --quiet
                     git clone --depth 1 git@github.com:plotly/dash.git dash-main
                     pip install -e ./dash-main --quiet
+                    cd dash-main && pip install -r requires-ci.txt
```
alexcjohnson: does this not work as `pip install -e ./dash-main[ci]`?

CHANGELOG.md
```diff
@@ -1,12 +1,22 @@
...
+- New `selected_columns` prop that contains the list of visible and hidden columns that are currently selected
+- New `derived_viewport_selected_columns` that contains the list of visible columns that are currently selected
+    This prop is read-only. Use `selected_columns` in callbacks instead.
+
 ### Fixed
 [#533](https://github.com/plotly/dash-table/issues/533)
 - Fixed problem clearing one column shifting everything to the left and
 leaving the last column blank
-- Add merge_duplicate_headers prop to correct `export_format: display` behaviour.
+- Add merge_duplicate_headers prop to correct `export_format: display` behavior.
```
alexcjohnson: heh I'm sure we have British spelling all over the place... it's important not to use it in public API but I usually let it slide in text. I won't complain about removing it though :)

src/dash-table/dash/DataTable.js
```diff
@@ -128,7 +128,7 @@ export const propTypes = {
          * If `last`, the `clear` action button will only appear on the last header
          * row. If `first` it will only appear on the first header row. These
          * are respectively shortcut equivalents to `[false, ..., false, true]` and
-         * `[true, ..., true, false]`.
+         * `[true, false, ..., false]`.
```
alexcjohnson: 🙈 good catch!

CHANGELOG.md
Marc-Andre-Rivet: I have a spellchecker running with `en-US` as the default in vscode - it's not great but when it catches things like that I try and fix them.

.circleci/config.yml
Marc-Andre-Rivet: should work. updating

src/dash-table/dash/DataTable.js
```diff
@@ -1114,6 +1148,10 @@ export const propTypes = {
         PropTypes.oneOfType([PropTypes.string, PropTypes.number])
     ),
 
+    /**
+     * `derived_viewport_selected_columns` contains the ids of the
+     * `selected_columns` that are not currently hidden.
```
alexcjohnson: Looks a bit funny as a double negative... though I guess this way is technically the most correct because a column can be "not currently hidden" but also not visible if it's scrolled off-screen.

PR #568, opened 2019-09-04
Issue 567 - Add border-radius support

Marc-Andre-Rivet:


src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -817,7 +818,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
                 <div
                     ref='table'
                     className={innerClasses.join(' ')}
-                    style={tableStyle}
+                    style={R.omit(BORDER_PROPERTIES_AND_FRAGMENTS, tableStyle)}
```
Marc-Andre-Rivet: Omit the border stuff on the inner container and only apply to the outer container - otherwise the border gets doubled.

src/dash-table/derived/style/py2jsCssProperties.ts
```diff
@@ -557,6 +559,7 @@ export default new Map<string, string>([
     ['borderLeftStyle', 'borderLeftStyle'],
     ['borderLeftWidth', 'borderLeftWidth'],
     ['borderRight', 'borderRight'],
+    ['borderRadius', 'borderRadius'],
```
Marc-Andre-Rivet: Adding support for `border-radius` and the Python and React flavors

tests/visual/percy-storybook/Border.defaults.percy.tsx
```diff
@@ -144,4 +144,17 @@ storiesOf('DashTable/ListView style, Border (available space filled)', module)
...
+        setProps,
+        id: 'table',
+        data: data,
+        columns,
+        style_table: {
+            border: '1px solid red',
+            borderRadius: '15px',
+            overflow: 'hidden'
+        }
+    }} />));
```
Marc-Andre-Rivet: Basic test to make sure the radius is applied

src/dash-table/derived/style/py2jsCssProperties.ts
alexcjohnson: whew, exciting file... any reason this couldn't be rewritten DRYer? Like:
```js
const attrParts = [
    // each attr only once
    ['align', 'content'],
    ['align', 'items'],
    ['alignment', 'adjust'],
    ...
];

const makeMap = attrParts => {
   // generate the tripled list in all formats
}

export default new Map(makeMap(attrParts));
```
Wouldn't need to be done now, but seems more maintainable and efficient.

src/dash-table/derived/edges/type.ts
```diff
@@ -26,10 +26,19 @@ export const BORDER_PROPERTIES: BorderProp[] = [
...
+
 export const BORDER_PROPERTIES_AND_FRAGMENTS: string[] = R.uniq(
     R.filter(
         p => p.indexOf('border') === 0,
-        Array.from(py2jsCssProperties.values())
+        cssProperties
+    )
+);
+
+export const HEIGHT_AND_WIDTH_PROPERTIES: string[] = R.uniq(
```
alexcjohnson: heh could get rid of these `R.uniq` calls too if `py2jsCssProperties` was rewritten 🌴 and exposed the unique list too 😉

src/dash-table/derived/style/py2jsCssProperties.ts
Marc-Andre-Rivet: Actually this file is generated - forgot to push the generator update.

Goal of the generator is to prevent having to eval all of this at runtime and only have a table mapping for python/css/react casing. That said, your comment below (https://github.com/plotly/dash-table/pull/568#discussion_r320931811) is correct. Should generate a resolve array of these values instead of doing it on the fly/

Marc-Andre-Rivet:
Updated the generator based on @alexcjohnson comments. Without the regex involved this is indeed a bit smaller than it was before and the initial gen cost in the FE is negligible.

src/dash-table/derived/style/py2jsCssProperties.ts
```diff
@@ -1,766 +1,28 @@
...
+    camels.push(camel);
+
+    entries.push([camel, camel]);
+    entries.push([toKebabCase(prop), camel]);
+    entries.push([toSnakeCase(prop), camel]);
+});
+
+export default new Map<string, string>(entries);
+
+export const KnownCssProperties = camels;
```
alexcjohnson: heh 🐫 makes it 🌴 🎉

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -817,7 +822,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
                 <div
                     ref='table'
                     className={innerClasses.join(' ')}
-                    style={tableStyle}
+                    style={INNER_STYLE}
```
alexcjohnson: Ah that's way better 🏆

PR #571, opened 2019-09-05
fix row selection alignment

### LINKED ISSUE #18: improve row select radio button style
row select radio buttons are too big?
image

Marc-Andre-Rivet:
Fixes #18. Will trigger many visual test changes.

image

image


alexcjohnson:
What about @chriddyp's "too big" comment, is that still an issue?

Marc-Andre-Rivet:
I don't think so. The inputs are significantly smaller than the screenshot from June last year.

PR #573, opened 2019-09-05
Issue 188 - Simpler multiline / ellipsis

### LINKED ISSUE #188: A simpler interface for multi-line cells and ellipses?
Currently, multi-line cells requires some pretty complex selectors and rules:
```
        dash_table.DataTable(
            style_table={'width': '100%'},
            style_data={'whiteSpace': 'normal'},
            content_style='grow',
            css=[{
                'selector': '.dash-cell div.dash-cell-value',
                'rule': 'display: inline; white-space: inherit; overflow: inherit; text-overflow: inherit;'
            }],
            data=df_election.to_dict('rows'),
            columns=[{'id': c, 'name': c} for c in df_election.columns]
        ),
```

image

How can we make this simpler?

Marc-Andre-Rivet:
Fixes #188 

Requires a `dash-docs` usage update: https://github.com/plotly/dash-docs/pull/642

Up to now it was necessary to do:
```
    style_data={'whiteSpace': 'normal'},
    css=[{
        'selector': '.dash-cell div.dash-cell-value',
        'rule': 'display: inline; white-space: inherit; overflow: inherit; text-overflow: inherit;'
    }],
```
in order to allow multiline & ellipsis.

With this change, it is possible to do it with only:
```
    style_data={
        'height': 'auto',
        'whiteSpace': 'normal'
    },
```

It's also now possible to set `white-space`, `overflow` and `text-overflow` in `style_**` and have it applied (inherited) by the cell's content without extra css.

Alongside being simpler, it also fixes old/weird alignment issues we've been seeing in the documentation for a long time now:

Before
image

After
image

At some point, it might be interesting to know which css properties can safely be inherited by default by `div.dash-cell-value` and update https://github.com/plotly/dash-table/blob/188-multiline/src/dash-table/components/Table/Table.less#L367 with all those values. 

See follow up: https://github.com/plotly/dash-table/issues/574


src/core/Stylesheet.ts
```diff
@@ -55,19 +55,6 @@ class StylesheetFacade {
...
-        } else {
-            return `${dimension}${defaultUnit}`;
-        }
-    }
-
-    static hasUnit(dimension: any) {
-        return typeof dimension === 'string' &&
-            /^\d+(\.\d+)?(px|em|rem|%)$/.test(dimension);
-    }
-
```
Marc-Andre-Rivet: Stumbled upon this code. It isn't used anymore. Removing.

alexcjohnson:
Simpler is nice, but the remaining percy diffs look problematic and in the matching docs PR I don't see ellipsis https://dash-docs-pr-642.herokuapp.com/datatable/sizing

Marc-Andre-Rivet:
@alexcjohnson Sorry about that. I promoted this to `ready for review` when it really wasn't. There's some additional tweaking required.

Marc-Andre-Rivet:
@alexcjohnson I think this is ready for another round

PR #578, opened 2019-09-08
576 - column edits with hidden columns

### LINKED ISSUE #576: Bug editing column names while columns are hidden
Causes hidden columns to be dropped entirely

alexcjohnson:
Fixes #576 - editing column names or deleting (and clearing, which already worked correctly, but I made sure it still does) columns while some columns are hidden.

Also fixes a minor issue I encountered along the way: if you click to edit a column name then click `Cancel` at the prompt, it would clear the name. Fixed it so now it keeps the previous name.

I chose two slightly different behaviors for merged columns with only part of the merged set hidden, but I think it makes sense:
- For name edits, even the hidden pieces get renamed. The rationale being that if several columns have a merged header, that defines them as a category, and by editing that part of the name you're altering the category name, not altering what's *in* that category.
- For deleting / clearing, nothing hidden is removed. Here I felt it was more important, because this is an irreversible change, to only affect things you can see.

This creates a weird edge case with name editing:
- Make three columns (with different second row names so we don't hit #490 😏) with first-row names `A`, `B`, `A`
- Hide `B` -> now the first row headers merge into a single `A`
- Rename `A` to `C`
- Only the first column changes to `A`; the other one is still `A`, so they cease to be merged. This is because when renaming we consider all the columns, not just the visible ones, so in that context these cells actually aren't merged.

I'm torn about whether to consider this a bug or not... but there it is. It's pretty esoteric and I could see arguments either way


Marc-Andre-Rivet:
Would you mind documenting the edge case as a standalone issue.. let's say a bug + discussion needed so as to not lose track of it?

First reaction is that very little distinguishes a hidden column from a deleted column. I think it's more surprising to the user to see table 1 with deleted B go from `A A` -> `C A` and table 2 with hidden B go from `A A` -> `C C` then having both behave the same.

But you're right, this is not clear cut.

alexcjohnson:
@Marc-Andre-Rivet changelog added and new issue created for the edge case mentioned https://github.com/plotly/dash-table/issues/585 - any more comments?

PR #589, opened 2019-09-18
Issue 460 - datestartswith relational operator behavior on number expression

### LINKED ISSUE #460: datestartswith should cast values to strings automatically
filtering-datetime

`1975` gets processed as a number and the date normalization function expects a string

Marc-Andre-Rivet:
Fixes #460 

- improve datestartswith operator handling of types
- fix issue where equal was incorrectly used on default operators
- add appropriate unit tests

### DIFF
--- a/src/dash-table/syntax-tree/lexeme/relational.ts
+++ b/src/dash-table/syntax-tree/lexeme/relational.ts
@@ -50,12 +50,16 @@ const LEXEME_BASE = {
 };
 
 export const contains: IUnboundedLexeme = R.merge({
-    evaluate: relationalEvaluator(([op, exp]) =>
-        !R.isNil(exp) &&
-        !R.isNil(op) &&
-        (R.type(exp) === 'String' || R.type(op) === 'String') &&
-        op.toString().indexOf(exp.toString()) !== -1
-    ),
+    evaluate: relationalEvaluator(([op, exp]) => {
+        if (R.isNil(op) || R.isNil(exp)) {
+            return false;
+        }
+
+        op = typeof op === 'number' ? op.toString() : op;
+        exp = typeof exp === 'number' ? exp.toString() : exp;
+
+        return op.toString().indexOf(exp.toString()) !== -1;
+    }),
     subType: RelationalOperator.Contains,
     regexp: /^(contains)/i
 }, LEXEME_BASE);
@@ -88,6 +92,9 @@ const DATE_OPTIONS: IDateValidation = {
 
 export const dateStartsWith: IUnboundedLexeme = R.merge({
     evaluate: relationalEvaluator(([op, exp]) => {
+        op = typeof op === 'number' ? op.toString() : op;
+        exp = typeof exp === 'number' ? exp.toString() : exp;
+
         const normalizedOp = normalizeDate(op, DATE_OPTIONS);
         const normalizedExp = normalizeDate(exp, DATE_OPTIONS);
 
--- a/src/dash-table/syntax-tree/lexeme/relational.ts
+++ b/src/dash-table/syntax-tree/lexeme/relational.ts
@@ -50,16 +50,12 @@ const LEXEME_BASE = {
 };
 
 export const contains: IUnboundedLexeme = R.merge({
-    evaluate: relationalEvaluator(([op, exp]) => {
-        if (R.isNil(op) || R.isNil(exp)) {
-            return false;
-        }
-
-        op = typeof op === 'number' ? op.toString() : op;
-        exp = typeof exp === 'number' ? exp.toString() : exp;
-
-        return op.toString().indexOf(exp.toString()) !== -1;
-    }),
+    evaluate: relationalEvaluator(([op, exp]) =>
+        !R.isNil(exp) &&
+        !R.isNil(op) &&
+        (R.type(exp) === 'String' || R.type(op) === 'String') &&
+        op.toString().indexOf(exp.toString()) !== -1
+    ),
     subType: RelationalOperator.Contains,
     regexp: /^(contains)/i
 }, LEXEME_BASE);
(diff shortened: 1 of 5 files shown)


tests/cypress/tests/unit/single_column_syntactic_tree_test.ts
```diff
@@ -110,9 +117,119 @@ describe('Single Column Syntax Tree', () => {
...
+
+                expect(structure.right).to.not.equal(null);
+                if (structure.right) {
+                    expect(structure.right.type).to.equal(LexemeType.Expression);
+                    expect(structure.right.subType).to.equal('value');
+                    expect(structure.right.value).to.equal(1);
+                }
+            }
+        });
+    });
```
Marc-Andre-Rivet: Without the change to the single column syntax tree
https://circleci.com/gh/plotly/dash-table/13136, https://github.com/plotly/dash-table/pull/589/commits/04bbe1d41f30cfd1e634c6274405943db89e7fcf
and passing again afterwards
https://circleci.com/gh/plotly/dash-table/13143, https://github.com/plotly/dash-table/pull/589/commits/8087b40d7d96d6147ae45decd5f6a14f082efc67

tests/cypress/tests/unit/dash_table_queries_test.ts
```diff
@@ -74,8 +74,8 @@ describe('Dash Table Queries', () => {
 
                 describe('contains', () => {
                     processCases(c.syntaxer, [
-                        { name: 'cannot compare "11" to 1', query: `${c.hideOperand ? '' : '{a} '}contains 1`, target: { a: '11' }, valid: true, evaluate: true },
-                        { name: 'cannot compare 11 to 1', query: `${c.hideOperand ? '' : '{a} '}contains 1`, target: { a: 11 }, valid: true, evaluate: false },
+                        { name: 'compares "11" to 1', query: `${c.hideOperand ? '' : '{a} '}contains 1`, target: { a: '11' }, valid: true, evaluate: true },
```
Marc-Andre-Rivet: typo

tests/cypress/tests/unit/dash_table_queries_test.ts
```diff
@@ -74,8 +74,8 @@ describe('Dash Table Queries', () => {
...
+                        { name: 'compares 11 to 1', query: `${c.hideOperand ? '' : '{a} '}contains 1`, target: { a: 11 }, valid: true, evaluate: true },
```
Marc-Andre-Rivet: `contains` now does not distinguish between string 11 and number 11 in all (string|number, string|number) operand/expression combinations

tests/cypress/tests/unit/dash_table_queries_test.ts
```diff
@@ -84,8 +84,8 @@ describe('Dash Table Queries', () => {
                         { name: 'compares "abc" to "b"', query: `${c.hideOperand ? '' : '{a} '}contains "b"`, target: { a: 'abc' }, valid: true, evaluate: true },
                         { name: 'compares "abc" to " b"', query: `${c.hideOperand ? '' : '{a} '}contains " b"`, target: { a: 'abc' }, valid: true, evaluate: false },
                         { name: 'compares "abc" to "b "', query: `${c.hideOperand ? '' : '{a} '}contains "b "`, target: { a: 'abc' }, valid: true, evaluate: false },
-                        { name: 'compares "abc" to " b"', query: `${c.hideOperand ? '' : '{a} '}contains " b"`, target: { a: 'a bc' }, valid: true, evaluate: true },
-                        { name: 'compares "abc" to "b "', query: `${c.hideOperand ? '' : '{a} '}contains "b "`, target: { a: 'ab c' }, valid: true, evaluate: true }
+                        { name: 'compares "a bc" to " b"', query: `${c.hideOperand ? '' : '{a} '}contains " b"`, target: { a: 'a bc' }, valid: true, evaluate: true },
```
Marc-Andre-Rivet: typo

tests/cypress/tests/unit/dash_table_queries_test.ts
```diff
@@ -96,7 +96,7 @@ describe('Dash Table Queries', () => {
                         { name: '0yyy in "0yyy"', query: `${c.hideOperand ? '' : '{a} '}datestartswith "0987"`, target: { a: '0987' }, valid: true, evaluate: true },
                         { name: 'yyyy in "yyyy"', query: `${c.hideOperand ? '' : '{a} '}datestartswith "2006"`, target: { a: '2005' }, valid: true, evaluate: false },
                         { name: 'yyyy in "yyyy"', query: `${c.hideOperand ? '' : '{a} '}datestartswith "2005"`, target: { a: '2005' }, valid: true, evaluate: true },
-                        { name: 'yyyy in yyyy', query: `${c.hideOperand ? '' : '{a} '}datestartswith 2005`, target: { a: '2005' }, valid: true, evaluate: false },
+                        { name: 'yyyy in yyyy', query: `${c.hideOperand ? '' : '{a} '}datestartswith 2005`, target: { a: '2005' }, valid: true, evaluate: true },
```
Marc-Andre-Rivet: the relational operator will now cast number to string for `datestartswith`

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -50,12 +50,16 @@ const LEXEME_BASE = {
...
-    ),
+    evaluate: relationalEvaluator(([op, exp]) => {
+        if (R.isNil(op) || R.isNil(exp)) {
+            return false;
+        }
+
+        op = typeof op === 'number' ? op.toString() : op;
+        exp = typeof exp === 'number' ? exp.toString() : exp;
+
+        return op.toString().indexOf(exp.toString()) !== -1;
```
alexcjohnson: isn't this double `.toString`?

But also: should we be trying to use the column formatting here? What if you have prices formatted with 2-digit decimals and you want to match all even-dollar values, with `contains '.00'`?

Marc-Andre-Rivet: Getting access to formatting from here would be difficult. In a sense this would be no different from when we do a number comparison where the number could be presented in hex or the expression be an hex.

I think you're right, I got carried away by the number case for dates. Deviating too much from what was agreed upon when we last went through filtering. Will revert the `contains` changes and focus on (1) datestartswith and (2) the lexeme bug

alexcjohnson: OK - certainly understand that this would be tough, and it's a weird edge case so I'm happy to ignore for now. Let's make an issue for it though, if it doesn't already exist. If a user is filtering as a string, they would expect the string used in the filter to be the one they see.

Issue #313, opened 2018-12-19
Ability to export data as excel or csv


chriddyp:
- For Excel files, only XLSX (not XLS) will be supported
- Only the data will be exported, formatting will not be exported
- The export will include the data in the current view. For example, if columns are hidden, sorted, or filtered, then the exported file will display the current view.
- Export will not protect users from “CSV Injection” attacks (https://www.owasp.org/index.php/CSV_Injection)
Exporting Excel files may require a large 3rd party open source library. This library is too large to include in the default version of dash-table, so we will need to engineer dynamic JavaScript module loading as part of this requirement.

We will consider the UI for this feature in a separate issue (we need to design a UI needs encompasses all of the new features that we're adding).

Marc-Andre-Rivet:
XLSX: One option that seems well supported https://github.com/SheetJS/js-xlsx -- also has a pro/commercial tier with unknown conditions. The community version is under Apache License; the implications of that might need to be checked out to be safe.

CSV: Many options exist Including SheetJS above.

As for other optional operations on the table, implement this in the most isolated manner possible to keep it well isolated from the table's implementation. Internally should be its own component.

The feature's implementation must be purely front-end and will involve some JS magic to initiate what looks like a file download -- this is a known problem / been solved may times.

Lazy-loading would be great but is not needed for this feature to be implemented. Will increase the table's bundle size significantly though.

With a purely functional UI, this shouldn't take more than a few days to implement and test.

Marc-Andre-Rivet:
Just dropping this here, the pro version of xlsx-js offers styling support
https://sheetjs.com/pro

No idea about licensing.

Marc-Andre-Rivet:
@alinastarkov From our previous discussion, I'd say the CSV should export with the column id's and the XLSX should export with the column headers (multi-row / merged if required)

The idea being that if the user does a CSV export I would expect him to be somewhat aware of the data structure or planning on further processing the data in system x-y-z. Exporting headers would potentially cause the CSV to have multiple header rows which is not processed by default (in pandas, can do headers=[0,1] for rows 0 and 1 being used as headers) -- also, exporting the headers might result in duplicated names.

For XLSX, the user would arguably want to present or further process the data manually, not programmatically, and in this case the header names are more relevant.

@chriddyp are you ok with this?

chriddyp:
[...]

PR #592, opened 2019-09-19
Issue 546 - Visible columns and data misalignment in export

### LINKED ISSUE #546: v1.1.0 datatable hidden_columns/exporting loses hidden headers when export_header='names'
Hi there,

hidden_columns in 1.1.0 has really helped me a lot, however, here comes the question with exporting:

After using hidden_columns to hide some of the columns, I need to export data(no matter what is displayed after hiding several columns or the whole table).

It seems that the **data** of hidden columns will be exported anyway but the **headers** of them will not if I set **export_headers='names'.** It just leaves several data without headers.

I think the "ids" or the "names" should be same for the headers but the content, therefore, could I believe that there might be something wrong?

Here is a demo code to show that problem:

`import dash
import dash_table

app = dash.Dash(__name__)

    id='table',
    columns=[{"name": i, "id": i} for i in ['a', 'b', 'c', 'd']],
    data=[{'a':1, 'b':1, 'c':1, 'd':1}, {'a':1, 'b':1, 'c':1, 'd':1},{'a':1, 'b':1, 'c':1, 'd':1},{'a':1, 'b':1, 'c':1, 'd':1}],
    hidden_columns=['d'],
    export_format='csv',
    export_headers='names'
)

if __name__ == '__main__':
    app.run_server(debug=True)`

And the csv looks like:
image

Marc-Andre-Rivet:
Fixes #546 

- [ ] documentation update for the new prop and default export / columns behavior


tests/cypress/tests/unit/exportUtils_tests.ts
```diff
@@ -210,16 +210,20 @@ describe('export', () => {
...
+            ['rows', 'rows', 'rows']
+        ];
+
         const data = [
-                {col1: 1, col2: 2, col3: 3},
-                {col1: 2, col2: 3, col3: 4},
-                {col1: 1, col2: 2, col3: 3}
+            { col1: 1, col2: 2, col3: 'x', col4: 3 },
+            { col1: 2, col2: 3, col3: 'x', col4: 4 },
+            { col1: 1, col2: 2, col3: 'x', col4: 3 }
```
Marc-Andre-Rivet: Add one unused (hidden) column to the dataframe

tests/cypress/tests/unit/exportUtils_tests.ts
```diff
@@ -266,7 +270,7 @@ describe('export', () => {
                 B2: {t: 'n', v: 2},
                 B3: {t: 'n', v: 3},
                 B4: {t: 'n', v: 2},
-                C1: {t: 's', v: 'col3'},
+                C1: {t: 's', v: 'col4'},
```
Marc-Andre-Rivet: Changed col3 for col4 above

alexcjohnson:
The poster in #546 thought the correct behavior should be to show *all* columns, hidden or not. This PR results in the export correctly matching what's displayed, which sounds like a reasonable default, but perhaps to make everyone happy we need to support either option somehow?

Marc-Andre-Rivet:
I think I misunderstood the poster's meaning. I still think the correct **default** behavior is to ignore hidden columns.

(1) we already replicate the view’s content when it comes to sorting and filtering
(2) this is what we had in mind when implementing the feature, see https://github.com/plotly/dash-table/issues/313
(3) including hidden columns would  leave no alternative means of **not** including the columns if someone wanted to do the reverse -- they would be left with only the option of deleting the columns and the associated data. Here, the user can simply make the columns visible again to make the view match the desired export.
(4) interestingly default behavior is not uniform - LibreOffice ignores hidden columns by default on png/csv export and Google Spreadsheet does the opposite, can't test Excel

My suggestion is to expand on the `export_headers` prop like so:

```
    export_headers: PropType.oneOfType(
        PropTypes.oneOf(['none', 'ids', 'names', 'display']),
        PropTypes.exact({
            headers: PropTypes.oneOf(['none', 'ids', 'names', 'display']),
            columns: Proptypes.oneOf(['all', 'visible'])
        })
    ),
```

As we did elsewhere, leaving the basic usage simple but presenting a more complex prop for tweaking / advance usage.

alexcjohnson:
Whether to include all or just visible columns doesn't really fit under the prop `export_headers` IMO - I'd just make a new prop `export_columns: Proptypes.oneOf(['all', 'visible'])`

Marc-Andre-Rivet:
Fair enough. My attempts to minimize props were ill advised.

src/dash-table/components/Export/index.tsx
```diff
@@ -1,33 +1,37 @@
...
 }
 
 export default React.memo((props: IExportButtonProps) => {
 
-    const { columns, export_format, virtual_data, export_headers, visibleColumns, merge_duplicate_headers } = props;
-    const isFormatSupported = export_format === 'csv' || export_format === 'xlsx';
+    const { columns, export_columns, export_format, virtual_data, export_headers, visibleColumns, merge_duplicate_headers } = props;
+    const isFormatSupported = export_format === ExportFormat.Csv || export_format === ExportFormat.Xlsx;
+
+    const exportedColumns = export_columns === ExportColumns.Visible ? visibleColumns : columns;
```
Marc-Andre-Rivet: Use either visible columns or all columns depending on the prop setting

src/dash-table/components/Export/index.tsx
```diff
@@ -1,33 +1,37 @@
...
 import React from 'react';
-import { IDerivedData, Columns } from 'dash-table/components/Table/props';
+import { IDerivedData, Columns, ExportHeaders, ExportFormat, ExportColumns } from 'dash-table/components/Table/props';
 import { createWorkbook, createHeadings, createWorksheet } from './utils';
 import getHeaderRows from 'dash-table/derived/header/headerRows';
 
 interface IExportButtonProps {
     columns: Columns;
-    export_format: string;
+    export_columns: ExportColumns;
```
Marc-Andre-Rivet: Updated string usage to enums

src/dash-table/components/Export/utils.tsx
```diff
@@ -1,13 +1,13 @@
...
-import { Data } from 'dash-table/components/Table/props';
+import { Data, ExportHeaders } from 'dash-table/components/Table/props';
 
 interface IMergeObject {
     s: {r: number, c: number};
     e: {r: number, c: number};
 }
 
-export function transformMultDimArray(array: (string | string[])[], maxLength: number): string[][] {
+export function transformMultiDimArray(array: (string | string[])[], maxLength: number): string[][] {
```
Marc-Andre-Rivet: typo

tests/cypress/tests/unit/exportUtils_tests.ts
```diff
@@ -290,11 +295,11 @@ describe('export', () => {
         });
         it('create sheet with undefined column for clearable columns', () => {
             const newData = [
-                {col2: 2, col3: 3},
-                {col2: 3, col3: 4},
-                {col2: 2, col3: 3}
+                {col2: 2, col4: 3},
+                {col2: 3, col4: 4},
+                {col2: 2, col4: 3}
```
Marc-Andre-Rivet: col3 became col4 above

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 [#460](https://github.com/plotly/dash-table/issues/460)
 - The `datestartswith` relational operator now supports number comparison
 - Fixed a bug where the implicit operator for columns was `equal` instead of the expected default for the column type
+[#546](https://github.com/plotly/dash-table/issues/546)
+- Visible columns are used correctly for both header and data rows
+
+### Added
+[#546](https://github.com/plotly/dash-table/issues/546)
+- New prop `export_columns` that takes values `all` or `visible` (default). This prop controls the columns used during export
```
alexcjohnson: I'd prefer if `Added` came before `Fixed` in the changelog - new features >> 🐛 . (anyway as is it's splitting the `Fixed` section)

PR #593, opened 2019-09-19
Issue 539 - Booleans not showing

### LINKED ISSUE #539: Boolean values not showing
Like in the experimental version, cells with boolean values are empty, regardless of True or False.

https://github.com/plotly/dash-table-experiments/issues/10

dmt0:
Fixes #539

Note, that this does not add `bool` as a supported type. Filtering won't work, since trying to search for `true` or `false` will just search for those strings. Sorting puts `false` before number `0`, and `true` between numbers `0` and `1`.


.vscode/settings.json
```diff
@@ -11,13 +11,13 @@
...
-    ],
+      "'"
+  ],
     "cSpell.language": "en",
     "cSpell.diagnosticLevel": "Error",
     "cSpell.languageSettings": [
-        { "languageId": "*", "dictionaries": ["fonts", "css", "html", "npm", "typescript", "python"]}
-    ],
+      { "languageId": "*", "dictionaries": ["fonts", "css", "html", "npm", "typescript", "python"]}
+  ],
```
Marc-Andre-Rivet: Why? - is `]`'s indent changed?

.vscode/settings.json
```diff
@@ -40,5 +40,10 @@
         "yotta",
         "zepto",
         "zetta"
-    ]
+    ],
+    "editor.formatOnSave": false,
+    "javascript.format.enable": false,
+    "typescript.format.enable": false,
+    "json.format.enable": false,
```
Marc-Andre-Rivet: Most of these probably are better off in your user's settings than the project settings

dmt0: Actually on all other Plotly projects that I see (I've been in 5 so far, and it seems true for projects in the Dash land too), `.vscode` is added to `.gitignore`. And it makes sense. VSCode is awesome, but we can't assume that every developer will be using it, especially when we have an open source project. Here we seem to be using it mostly for cSpell, which can have its own config file AFAIKT.

We've also been using prettier as an auto-formatter everywhere (which is awesome BTW!).

Of course every developer that works on `dash-table` could just set the global config to whatever is above, and then disable auto-formatter for each project locally one by one. It's this project that's going against the grain though.

dmt0: Leaving this out possibly for another PR...

dmt0:
Looks like I'm not authorized to merge.

PR #594, opened 2019-09-19
Issue 591 - Row and column selection with multiple tables

### LINKED ISSUE #591: row/column selectable single applies to all tables on the page
Make 2 tables with row or column selectable=single - you can only select one row and one column on the page, should be one per table.

reported in https://community.plot.ly/t/multiple-dash-tables-selection-error/28802

```
import dash
import dash_table

keys = ['a', 'b', 'c']
data = [{'a': 'a', 'b': 'b', 'c': 'c'}, {'a': 'aa', 'b': 'bb', 'c': 'cc'}]

app = dash.Dash(__name__)
app.layout = html.Div([
    dash_table.DataTable(
        id='a',
        columns=[{"name": i, "id": i, "selectable": True} for i in keys],
        data=data,
        column_selectable='single',
        row_selectable='single'
    ),
    dash_table.DataTable(
        id='b',
        columns=[{"name": i, "id": i, "selectable": True} for i in keys],
        data=data,
        column_selectable='single',
        row_selectable='single'
    )
])

if __name__ == '__main__':
    app.run_server(debug=True)
```

Marc-Andre-Rivet:
Fixes #591


demo/App.tsx
```diff
@@ -50,14 +50,18 @@ class App extends Component<any, any> {
...
 
-            return (<DataTable
-                {...props}
-            />);
+            const baseId = this.state.tableProps2 && this.state.tableProps2.id;
+
+            return (this.state.tableProps2 ? <DataTable
+                {...this.state.tableProps2}
+                id={baseId ? 'table2' : baseId}
+            /> : null);
```
Marc-Andre-Rivet: Change the 2 tables test behavior so that
- the props are not shared between the tables 🙄 
- the 2nd table will have an `id` only if the 1st one has one

src/dash-table/derived/cell/operations.tsx
```diff
@@ -42,7 +42,7 @@ function rowSelectCell(idx: number, rowSelectable: Selection, selectedRows: numb
         <input
             type={rowSelectable === 'single' ? 'radio' : 'checkbox'}
             style={{ verticalAlign: 'middle' }}
-            name='row-select'
+            name={`row-select-${id}`}
```
Marc-Andre-Rivet: Use the table's id to make the group unique

src/dash-table/derived/header/content.tsx
```diff
@@ -230,7 +231,7 @@ function getter(
                                         column_selectable === 'single',
                                         !allSelected
                                     )}
-                                    name='column-header--select'
+                                    name={`column-select-${id}`}
```
Marc-Andre-Rivet: Use the table's id to make the group unique

tests/cypress/src/DashTable.ts
```diff
@@ -19,104 +19,200 @@ const getSelector = (state: State) => {
     }
 };
 
-export default class DashTable {
-    static getCell(row: number, column: number, editable: State = State.Ready) {
-        return cy.get(`#table ${getSelector(editable)} tbody tr td.column-${column}`).eq(row);
+export class DashTableHelper {
```
Marc-Andre-Rivet: Generalize the helper to work for multiple tables - Leaving the static implementation below for existing tests

tests/cypress/tests/standalone/select_row_column_test.ts
```diff
@@ -0,0 +1,79 @@
...
+                DashTable.getSelect(0).within(() => cy.get('input').click());
+                DashTable.getSelect(0).within(() => cy.get('input').should('be.checked'));
+                cy.get('tr th.column-0 .column-header--sort').last().click({ force: true }).click({ force: true });
+                DashTable.getSelect(0).within(() => cy.get('input').should('not.be.checked'));
+            });
+        });
+    });
+});
+
+describe('select rows & columns in multiple tables without id', () => {
```
Marc-Andre-Rivet: Renamed the file to take into consideration that there are now rows and columns tests.

Create two tables without ids and check that row and column selections are not shared between them.

tests/cypress/src/DashTable.ts
```diff
@@ -19,104 +19,200 @@ const getSelector = (state: State) => {
...
-    static toggleScroll(toggled: boolean) {
+    public toggleScroll(toggled: boolean) {
         cy.get('.row-1').then($el => {
             $el[0].style.overflow = toggled ? '' : 'unset';
         });
     }
 }
+
+export default class DashTable {
+    private static __helper = new DashTableHelper('table');
```
alexcjohnson: would it not work to just say `export default const DashTable = new DashTableHelper('table');`?

Marc-Andre-Rivet: 😞 very much so

Issue #598, opened 2019-09-20
quotes are required for filtering text entries with spaces in it


chriddyp:
Currently, you need to supply quotes around text when filtering. I think our default behavior should allow filtering strings without this as using quotes isn't discoverable.

![filtering with spaces](https://user-images.githubusercontent.com/1280389/65345233-9c8f8080-db8e-11e9-97ac-b7cf37683241.gif)

chriddyp:
Relatedly, I don't understand why `PRODUCT3` returns _all_ of the results?

Marc-Andre-Rivet:
The quotes being required is not a bug but working as intended.

Expected in `4.3.0`
- User enters `PRODUCT 3`
- The filter parsing to fail resulting in a `red` highlight on that cell
- No filter is applied

Should be possible to allow for a more permissive string selector for the single column case because the form of the filter is limited to few cases:
- `<unary>`
- `<relational> <value|expression>`
- `<value|expression>`

Issue #597, opened 2019-09-20
Filtering for blank values is akward at the moment


emmanuelle:
Triggerered by [this community post](https://community.plot.ly/t/editable-datatable-empty-cell-highligthing/28998). 

In the context of editable tables, filtering for blank cells makes a lot of sense.

In order to filter for blank (non-defined) values, at the moment one has to do something like `{col} is nil or {col} == ''`. It would be more readable to have a syntax like `is blank` (or something to be discussed).

PR #599, opened 2019-09-20
Issue 598 - Permissive value in filter

Marc-Andre-Rivet:
#598
Allow more permissive expression value (with whitespaces without escaping or quotes) in column filters.

A limitation is that if it starts with an operator's text, the operator will be picked up instead. e.g `le jour` is `<= jour`, not `"le jour"`.

Added a test case for permissive strings. 
Added changelog entry.

issue598

#597 
New `is blank` unary operator that returns true for `undefined`, `null` and `''`.

In both cases, the documentation will need to be updated.

### DIFF
--- a/src/dash-table/syntax-tree/lexeme/expression.ts
+++ b/src/dash-table/syntax-tree/lexeme/expression.ts
@@ -6,6 +6,7 @@ import { ISyntaxTree } from 'core/syntax-tree/syntaxer';
 const FIELD_REGEX = /^{(([^{}\\]|\\.)+)}/;
 const STRING_REGEX = /^(('([^'\\]|\\.)*')|("([^"\\]|\\.)*")|(`([^`\\]|\\.)*`))/;
 const VALUE_REGEX = /^(([^\s'"`{}()\\]|\\.)+)(?:[\s)]|$)/;
+const PERMISSIVE_VALUE_REGEX = /^(([^'"`{}()\\]|\\.)+)(?:$)/;
 
 const getField = (
     value: string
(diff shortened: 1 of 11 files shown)


src/dash-table/syntax-tree/lexeme/expression.ts
```diff
@@ -6,6 +6,7 @@ import { ISyntaxTree } from 'core/syntax-tree/syntaxer';
 const FIELD_REGEX = /^{(([^{}\\]|\\.)+)}/;
 const STRING_REGEX = /^(('([^'\\]|\\.)*')|("([^"\\]|\\.)*")|(`([^`\\]|\\.)*`))/;
 const VALUE_REGEX = /^(([^\s'"`{}()\\]|\\.)+)(?:[\s)]|$)/;
+const PERMISSIVE_VALUE_REGEX = /^(([^'"`{}()\\]|\\.)+)(?:$)/;
```
alexcjohnson:
```suggestion
const PERMISSIVE_VALUE_REGEX = /^(([^'"`{}()\\]|\\.)+)$/;
```
Minor, no point not capturing the `$` right?

src/dash-table/syntax-tree/lexeme/expression.ts
```diff
@@ -55,17 +46,41 @@ export const stringExpression: IUnboundedLexeme = {
...
+        type: LexemeType.Expression
+    };
+};
+
+export const valueExpression = valueExpressionFactory(VALUE_REGEX);
+export const permissiveValueExpression = valueExpressionFactory(
+    PERMISSIVE_VALUE_REGEX,
+    (v: any) => typeof v === 'string' && v.indexOf(' ') !== -1 ?
+        `"${v}"` :
+        v
```
alexcjohnson: Nice solution - esp. as it relates to the generated query string. ✨

tests/cypress/tests/unit/single_column_syntactic_tree_test.ts
```diff
@@ -97,6 +97,14 @@ describe('Single Column Syntax Tree', () => {
         expect(tree.toQueryString()).to.equal('{a} = 1');
     });
 
+    it.only('can be permissive value expression', () => {
+        const tree = new SingleColumnSyntaxTree('Hello world', COLUMN_TEXT);
+
+        expect(tree.isValid).to.equal(true);
+        expect(tree.evaluate({ a: 'Hello world' })).to.equal(true);
+        expect(tree.evaluate({ a: 'Helloworld' })).to.equal(false);
+    });
```
alexcjohnson: Can you include a `toQueryString` test for permissive as well?

Marc-Andre-Rivet: Sure

PR #603, opened 2019-09-24
Remove artifacts from `dev`

Marc-Andre-Rivet:
This one turned out to be tricky because there's a weird interaction between the generator, Dash and the rebuilt dependency, in this case the table when the dependency is install in the dev's venv with `-e .`


dash_table_base/__init__.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import print_function as _
```
Marc-Andre-Rivet: What follows is problematic for a dev environment where the developer uses `pip install -e ./dash-table` to load the table package into his venv.

So... the dash generator is having some trouble here. With the generated *.py files missing, the package can no longer load correctly. This was apparently already an issue when this was done for the renderer, but since **all** the files are being generated in that case, the empty folder is handled correctly when Python does the `install_requires` check for Dash.

These files are not generated, moving them to a baseline folder that gets copied into `dash_table` at the right moment. That's the best option I've found so far apart from making the generator independent from Dash to resolve the dependencies chicken & egg problem.

The generation itself had to be modified also (package.json) because the R generation code depends on the baseline artifacts being present, which can't be, because otherwise the generator does not run at all..

So we generate Python, copy baseline, generate again with R. Not efficient but it works, it's two steps forward, one step backward.

package.json
```diff
@@ -21,7 +21,7 @@
     "private::build:js-dev": "run-s \"private::build -- --mode development --config webpack.dev.config.js\"",
     "private::build:js-test": "run-s \"private::build -- --mode development --config webpack.test.config.js\"",
     "private::build:js-test-watch": "run-s \"private::build -- --mode development --config webpack.test.config.js --watch\"",
-    "private::build:py": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json --r-prefix 'dash'",
+    "private::build:py": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json --r-prefix 'dash'",
```
Marc-Andre-Rivet: Double tap

package.json
```diff
@@ -43,10 +43,11 @@
     "private::test.standalone": "cypress run --browser chrome --spec 'tests/cypress/tests/standalone/**/*'",
     "build.watch": "webpack-dev-server --content-base dash_table --mode development --config webpack.dev.config.js",
     "build": "run-s private::build:js private::build:py",
+    "format": "run-s \"private::lint:ts -- --fix\"",
```
Marc-Andre-Rivet: For some reason, never made the tslint autofix available as a npm command..

package.json
```diff
@@ -43,10 +43,11 @@
...
     "lint": "run-s private::lint:*",
     "test.server": "run-p --race private::host* private::test.server",
     "test.standalone": "run-p --race private::host_js private::test.standalone",
-    "test.unit": "run-s private::test.python private::test.unit",
+    "test.unit": "run-s build private::test.python private::test.unit",
```
Marc-Andre-Rivet: removing the baseline artifacts, turns out the Py unit tests were always running against the latest release.. oops!

alexcjohnson: but if you run this locally you've probably already built, so would be annoying to have the test command do it again. Could we do this in the CI config instead?

Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/603/commits/2adceab9c9fe8fd9803eb20b3ee9841fe152b372

PR #605, opened 2019-09-30
Ensure that only the visible columns are exported.

### LINKED ISSUE #553: Export option - data columns are in a different order to the header names
This is connected to an issue noted in the community forum - 

https://community.plot.ly/t/datatable-export-columns-not-in-the-same-order-as-the-header/27507

When using the export functionality the header comes out in the correct order and matches the displayed table but the columns of data are in a different order. The net result is that the header names do not match the content.

This can be seen by using the export button on https://dash.plot.ly/datatable/editable and examining the results.

shammamah:
Closes #553.


src/dash-table/components/Export/index.tsx
```diff
@@ -27,12 +42,12 @@ export default React.memo((props: IExportButtonProps) => {
 
         const maxLength = getHeaderRows(columns);
         const heading = (export_headers !== ExportHeaders.None) ? createHeadings(columnHeaders, maxLength) : [];
-        const ws = createWorksheet(heading, virtual_data.data, columnID, export_headers, merge_duplicate_headers);
+        const ws = createWorksheet(heading, filterData(virtual_data.data, columnID), columnID, export_headers, merge_duplicate_headers);
         const wb = createWorkbook(ws);
         if (export_format === ExportFormat.Xlsx) {
-            XLSX.writeFile(wb, 'Data.xlsx', {bookType: 'xlsx', type: 'buffer'});
+            XLSX.writeFile(wb, 'Data.xlsx', { bookType: 'xlsx', type: 'buffer' });
```
shammamah: These diffs come from the linter -- I can remove them if necessary.

Marc-Andre-Rivet: I think this PR is redundant with what's been done here https://github.com/plotly/dash-table/pull/592. See https://github.com/plotly/dash-table/blame/dev/src/dash-table/components/Export/utils.tsx#L59 for column filtering.

PR #606, opened 2019-09-30
Issue 596 - Add Markdown presentation

### LINKED ISSUE #596: Allow for Markdown formatting in table cells
Existing issues like #222, #383, and #248 can be resolved if we add Markdown to the table cells.

Marc-Andre-Rivet:
Fixes #596 

- [x] check virtualized performance
- [x] virtualized behavior (varying row height -- breaks assumptions, this will be dealt outside the scope of this PR as this behavior is not caused by MD)
- [x] varying row height with fixed columns (https://github.com/plotly/dash-table/issues/649)
- [x] **navigation through and from markdown cell**
- [x] define copy/paste behavior (?!)
- [x] define filter behavior -- for now filters will behave the same way for markdown cells as for any other text cells (the value filtered on is the actual value in the df) -- expect at least one visual/sanity test for the filter behavior (guard rail)
- [x] define sort behavior (supporting/using displayed values is out of scope) -- for now sort will behave the same way for markdown cells as for any other text cells (the value sorted on is the actual value in the df) -- expect at least one visual/sanity test for the sort behavior (guard rail)
- [x] **visual tests** -- cover as much of md syntax as possible
- [x] behavior on non-text input value + test (e.g. number, boolean values, how does ReMarkable behave)
- [x] markdown image resize (test, docs)
- [x] more tests
- [x] docs (will be done in https://github.com/plotly/dash-docs/pull/733)

- [x] Test: Default highlight.js included with table does not assign itself to `window.hljs`
- [x] Test: Loading external highlight.js will take precedence on the table's version


src/dash-table/components/CellMarkdown/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+        const {
+            className,
+            value
+        } = this.props;
+
+        return (<div
+            className={[className, 'cell-markdown'].join(' ')}
+            {...this.getMarkdown(value)}
+        />);
+    }
```
Marc-Andre-Rivet: Starting with a fairly naive implementation that will work with non-virtualized tables as well as the more complex global caches, just not sure how it will scale with virtualization. To be tested.

src/dash-table/components/CellMarkdown/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface IProps {
+    active: boolean;
+    applyFocus: boolean;
+    className: string;
+    value: any;
+}
+
+export default class CellMarkdown extends PureComponent<IProps> {
+    private static md: Remarkable = new Remarkable();
```
Marc-Andre-Rivet: I've been thinking that maybe we can include the configuration for highlight.js usage, externalize the resource and expect special code block users to include the highlight.js they want as a resource from https://highlightjs.org/download/ ?! Either that or (1) not support it at all or (2) asyncing the resource -- in which case `React.Suspense` won't do and we'll have to handle it through state on `CellMarkdown`. This is mostly about completeness.. we should not load JS by default as it will involve a very special use case.

```
    private static md: Remarkable = new Remarkable({
        highlight: (value: string, language: string) => {
            const hljs: any = (window as any).hljs;

            if (!hljs) {
                return '';
            }

            if (language && hljs.getLanguage(language)) {
                try {
                    return hljs.highlight(language, value).value;
                } catch (_) { }
            }

            try {
                return hljs.highlightAuto(value).value;
            } catch (_) { }

            return '';
        }
    });
```

dash_table_base/__init__.py
```diff
@@ -76,7 +76,24 @@
...
+        'namespace': package_name,
+        'async': True
+    },
+    {
+        'relative_package_path': 'async~highlight.js.map',
+        'external_url': (
+            'https://unpkg.com/dash-table@{}/dash_table/async~highlight.js.map'
+        ).format(__version__),
+        'namespace': package_name,
+        'async': True
```
Marc-Andre-Rivet: Could you update the table's init.py file so it uses the same pattern as other repos with multiple async entries? e.g. https://github.com/plotly/dash-core-components/blob/dev/dash_core_components_base/__init__.py#L42

package.json
```diff
@@ -58,7 +60,10 @@
   "author": "Chris Parmer <chris@plot.ly>",
   "maintainer": "Ryan Patrick Kyle <ryan@plot.ly>",
   "license": "MIT",
-  "dependencies": {},
+  "dependencies": {
+    "@types/highlight.js": "^9.12.3",
+    "highlight.js": "^9.17.0"
+  },
```
Marc-Andre-Rivet: make these `devDeps` to be consistent with other deps usage

src/dash-table/components/Highlight/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export default class Highlight {
+
+    static isReady: Promise<boolean> | true = new Promise<boolean>(resolve => {
+        Highlight.hljsResolve = resolve;
+    });
+
+    static mdHtml = (value: string) => {
+        return Highlight.md.render(value);
+    }
```
Marc-Andre-Rivet: Renaming to `render` might be clearer

src/dash-table/components/Highlight/index.tsx
```diff
@@ -0,0 +1,42 @@
+import { Remarkable } from 'remarkable';
+import LazyLoader from 'dash-table/LazyLoader';
+
+export default class Highlight {
```
Marc-Andre-Rivet: 1. This is an indirection for the `Remarkable` library to use `Highlighlight.js`, not the other way around so the file's & class name are misleading
2. This is not a component, this code should live in another folder

src/dash-table/dash/DataTable.js
```diff
@@ -310,9 +310,9 @@ export const propTypes = {
 
         /**
          * The `presentation` to use to display the value.
-         * Defaults to 'input' for ['numeric', 'text', 'any'].
+         * Defaults to 'input' for ['datetime', 'numeric', 'text', 'any'].
```
Marc-Andre-Rivet: 😸

@Types/modules.d.ts
```diff
@@ -16,14 +16,14 @@ declare module 'fast-isnumeric' {
 }
 
 declare class Remarkable {
-    constructor();
+    constructor(options?: any);
```
Marc-Andre-Rivet: Just realized this `@types` exists for the library, maybe we can do away with our custom wrapper.
https://www.npmjs.com/package/@types/remarkable

third-party/highlight.js
```diff
@@ -0,0 +1,32 @@
...
+highlightjs.registerLanguage('markdown', markdown);
+highlightjs.registerLanguage('python', python);
+highlightjs.registerLanguage('r', r);
+highlightjs.registerLanguage('ruby', ruby);
+highlightjs.registerLanguage('shell', shell);
+highlightjs.registerLanguage('sql', sql);
+highlightjs.registerLanguage('xml', xml);
+highlightjs.registerLanguage('yaml', yaml);
+
+export default highlightjs;
```
Marc-Andre-Rivet: Since this is now our own cherry-picked version of Highlight.js would rather see this somewhere inside of `/src`

tests/visual/percy-storybook/@Types/modules.d.ts
```diff
@@ -23,14 +23,14 @@ declare module 'fast-isnumeric' {
 }
 
 declare class Remarkable {
-    constructor();
+    constructor(options?: any);
```
Marc-Andre-Rivet: Same as https://github.com/plotly/dash-table/pull/606/files#r358530234

tests/visual/percy-storybook/@Types/modules.d.ts
```diff
@@ -23,14 +23,13 @@ declare module 'fast-isnumeric' {
 }
 
 declare class Remarkable {
-    constructor();
     render(value: string): any;
 }
 
 declare interface RemarkableCtor {
-    new(): Remarkable;
+    new(options?: any): Remarkable;
```
Marc-Andre-Rivet: I think that with `@types/Remarkable` added, all of this becomes redundant.

@Types/modules.d.ts
```diff
@@ -16,14 +16,13 @@ declare module 'fast-isnumeric' {
 }
 
 declare class Remarkable {
-    constructor();
     render(value: string): any;
 }
 
 declare interface RemarkableCtor {
-    new(): Remarkable;
+    new(options?: any): Remarkable;
```
Marc-Andre-Rivet: I think that with `@types/Remarkable` added, all of the remarkable definitions becomes redundant.

tslint.json
```diff
@@ -9,6 +9,7 @@
             "cypress/**",
             "inst/**",
             "node_modules/**",
+	    "third-party/**",
```
Marc-Andre-Rivet: This is now redundant

Marc-Andre-Rivet:
@shammamah Testing navigation with:

```
import dash
from dash_html_components import Div

app = dash.Dash(__name__)

data=[
    dict(a=1, b='''```javascript
return 20;
```
'''),
    dict(a=1, b='''```js
console.log(\'pouet\');
        '''),
    dict(a=1, b='```js console.log(\'pouet\'); ```'),
    dict(a=1, b='```js console.log(\'pouet\'); ```'),
    dict(a=1, b='```js console.log(\'pouet\'); ```')
]

app.layout = Div([
    DataTable(
        columns=[
            dict(name='a', id='a', editable=True),
            dict(name='b', id='b', type='text', presentation='markdown')
        ],
        data=data,
        editable=False
    ),
    DataTable(
        columns=[
            dict(name='a', id='a'),
            dict(name='b', id='b', type='text', presentation='markdown')
        ],
        data=data,
        editable=False
    ),
])

app.run_server(debug=True)
```

Notice that for both tables it's possible to click/select a markdown cell and keyboard/navigate within the markdown column as expected. The top table, with `editable=True`, does not behave correctly after keyboard/navigating from column `a` to `b`. Notice that it's now impossible to keyboard/navigate. Does not apply to the read-only table.

muzamalrana:
hi Can anyone help me here ? I am totally stuck. Trying to embedd image from link into dash table cell. Following is the code snippet.
            html.Div([dt.DataTable(id='my-data',
                 columns=[
            {"id": 'artist_image', "name": "artist_image", "presentation": "markdown"},
            {"id": 'song_artist', "name": "song_artist"},
            {"id": 'monthly_listeners', "name": "monthly_listeners"},
            {"id": 'artist_followers', "name": "artist_followers"},
            {"id": 'artist_popularity', "name": "artist_popularity"}]
                                   ,style_as_list_view=True,
                                           filter_action="native",
    sort_action="native",
    sort_mode="multi",
    column_selectable="single",
                                   style_data_conditional=[
                                                {
                                                    'if': {'row_index': 'odd'},
                                                    'backgroundColor': '#000000'
                                                }
                                            ],
                                   style_cell={'textAlign': 'center','padding': '5px','backgroundColor':'#00FF00',
                                              'color':'white'},
                                                style_header={
                                                    'backgroundColor': '#202020',
                                                    'fontWeight': 'bold','color':'white'
                                                })],className='ten columns')])],className='row')], style=tab_style, selected_style=tab_selected_style)
@app.callback(
dash.dependencies.Output('my-data','data'),
[dash.dependencies.Input('my-graph1', 'clickData'),
dash.dependencies.Input('my_dropdown1', 'value'),
dash.dependencies.Input('radio_buttons', 'value'),
dash.dependencies.Input('my_dropdown3', 'value')]
)
def update_graph(click_data,dropdown,radio_button,dropdown3):
try:
genre1=click_data['points'][0]['customdata'][1]
except:
genre1='pop'
artist_data1=sm.artist_dataFrame(songs,artist,click_data,dropdown,radio_button)
print(artist_data1.head(1))
artist_data1=artist_data1[['artist_image','song_artist','monthly_listeners','artist_followers','artist_popularity','date_x']]
artist_data1['date_x']=pd.to_datetime(artist_data1['date_x'])
artist_data1=artist_data1[artist_data1['date_x']==artist_data1['date_x'].max()]
artist_data1=artist_data1.sort_values(by=radio_button,ascending=False).reset_index(drop=True)
artist_data1=artist_data1.drop(['date_x'],axis=1)
if dropdown3==0 or dropdown3=='All':
artist_data1=artist_data1
else:
artist_data1=artist_data1.head(dropdown3)
artist_data1['artist_image']=[html.Img(src=i) for i in artist_data1['artist_image']]
return artist_data1.to_dict('records')

PR #612, opened 2019-10-02
Add spaces to regex for relational operators.

### LINKED ISSUE #563: Text Comparison Operators
Typing "lens" as a filter for a text column gets interpreted as "le ns", that is "<= ns".

There is a workaround of course: typing "= lens" or "eq lens".

I suppose there's some value in being able to type "le something" instead of "<= something", but this should _only_ happen if the "le" is followed by a space.

I'd still prefer to be able to completely disable the text comparison operators. For example, French speakers would be surprised by the results of searching for "le something" compared to "Le Something" - even if their table provided case-insensitive search.

shammamah:
Closes #563.


alexcjohnson:
I think *all* the letter-based operators need this space - for example I can type `containsb` right now and it's interpreted as `contains b`

I think you were right to exclude the symbol operators - so `<5` still works as `< 5` - let's add a test though to lock that down.

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -57,7 +57,7 @@ export const contains: IUnboundedLexeme = R.merge({
         op.toString().indexOf(exp.toString()) !== -1
     ),
     subType: RelationalOperator.Contains,
-    regexp: /^(contains)/i
+    regexp: /^(contains\s)/i
```
Marc-Andre-Rivet: An incomplete query would end with `contains` without any additional spaces, we want to match that too for proper error handling / user feedback

```suggestion
    regexp: /^(contains(\s|$))/i
```

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -100,23 +100,23 @@ export const dateStartsWith: IUnboundedLexeme = R.merge({
...
     evaluate: relationalEvaluator(([op, exp]) => op < exp),
     subType: RelationalOperator.LessThan,
-    regexp: /^(<|lt)/i
+    regexp: /^(<|lt\s)/i
 }, LEXEME_BASE);
 
 export const notEqual: IUnboundedLexeme = R.merge({
     evaluate: relationalEvaluator(([op, exp]) => op !== exp),
     subType: RelationalOperator.NotEqual,
     regexp: /^(!=|ne)/i
```
Marc-Andre-Rivet: `ne` should also have `\s`

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -67,19 +67,19 @@ export const equal: IUnboundedLexeme = R.merge({
             op === exp
     ),
     subType: RelationalOperator.Equal,
-    regexp: /^(=|eq)/i
+    regexp: /^(=|eq\s)/i
```
Marc-Andre-Rivet: And the other ones :)

Marc-Andre-Rivet:
Needs changelog entry

tests/cypress/tests/unit/query_syntactic_tree_test.ts
```diff
@@ -589,5 +589,17 @@ describe('Query Syntax Tree', () => {
             expect(tree.evaluate({ a: 'abc v' })).to.equal(true);
             expect(tree.evaluate({ a: 'abc w' })).to.equal(false);
         });
+
+        it('correctly interprets text-based with no spaces as invalid', () => {
+            const tree = new QuerySyntaxTree('{a} le5');
+            expect(tree.isValid).to.equal(false);
+        });
+
+        it.only('correctly interprets non-text-based with no spaces as valid', () => {
```
Marc-Andre-Rivet:
```suggestion
        it('correctly interprets non-text-based with no spaces as valid', () => {
```

shammamah: oops 😛

Marc-Andre-Rivet:
@shammamah The percy diffs are interesting.. looks like there’s an extra space in the filter string now. Possibly a side effect of the new selector? There’s code that destructures/restructures that request from the individual column fragments. I suspect the new space is added somewhere in those steps.

Marc-Andre-Rivet:
There’s a “transform” on the lexemes that might need to be updated on the relational operators to take the (double?) space into account.

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -57,7 +57,8 @@ export const contains: IUnboundedLexeme = R.merge({
         op.toString().indexOf(exp.toString()) !== -1
     ),
     subType: RelationalOperator.Contains,
-    regexp: /^(contains)/i
+    regexp: /^((contains)(?=\s|$))/i,
```
Marc-Andre-Rivet: I had some browser support concerns but turns out lookahead is widely supported and not an issue. Lookbehind is not.
https://caniuse.com/#feat=js-regexp-lookbehind

PR #617, opened 2019-10-08
also allow 'parentheses' and 'percent'

theavey:
'parantheses' is spelled incorrectly; 
also allowing 'percent' is simple enough and can save three characters!

### DIFF
--- a/dash_table/Format.py
+++ b/dash_table/Format.py
@@ -51,6 +51,7 @@ Scheme = get_named_tuple('scheme', {
     'decimal_si_prefix': 's',
     'exponent': 'e',
     'fixed': 'f',
+    'percent': '%',
     'percentage': '%',
     'percentage_rounded': 'p',
     'binary': 'b',
@@ -65,6 +66,7 @@ Sign = get_named_tuple('sign', {
     'negative': '-',
     'positive': '+',
     'parantheses': '(',
+    'parentheses': '(',
     'space': ' '
 })
 


theavey:
I can add a comment in the changelog, but this seems a little insignificant. Let me know if you think it's warranted.

If the old spelling is removed later, that would definitely be worth a comment in the changelog, and possibly a deprecation warning beforehand.

theavey:
I haven't looked into the testing in this package, but I could also work on some tests for these changes.

alexcjohnson:
Thanks @theavey - good catch on `parantheses`. @Marc-Andre-Rivet how would you like to go about removing the misspelling? Only way I can think of to produce a deprecation notice when someone uses the misspelled one is to give it a fake value in the `namedtuple` and then look for that fake value in the `scheme` method, emit a warning, and convert to the real value.

I like the shorter `percent`, but I'm a little hesitant to add synonyms. @Marc-Andre-Rivet do you have an opinion here?

We're using the `dev` branch now for active development, `master` is just for releases. Would you retarget your PR to `dev`? I'm guessing that's why the tests didn't run...

alexcjohnson:
... or maybe outside PRs won't trigger CI runs in this repo? I notice #614 also didn't trigger CI. @Marc-Andre-Rivet is that the case? Can we enable that?

theavey:
The "percent" was less important to me, but it was the word I intuitively put in before looking up the correct keyword. I understand hesitating to put in synonyms for clarity and to avoid bloat.

I retargeted the PR, but that seems to have generated the merge conflict because I started from the master branch.

Marc-Andre-Rivet:
> or maybe outside PRs won't trigger CI runs in this repo?

They didn't. That's changed now. The next commit should trigger CI.

---

parantheses / parentheses: Yes! Let's definitely fix this and tag the misspelled one for deprecation in `v2.x` + create an issue for follow up.

percent / percentage: Also hesitant to add synonyms or to rename. Since no one seems to feel strongly about it, would drop it from the PR. Could be tagged alongside the typo for renaming in `v2.x`, it does feel a bit more consistent with the `money` format.

Marc-Andre-Rivet:
@theavey Took the liberty of updating the branch and .gitignore. The build artifacts will need to be removed from the PR.

That said, thanks for sticking with it, this PR has been waiting for too long. 🙇

PR #620, opened 2019-10-09
Issue 618 - Keyboard navigation in sparsely readonly table

### LINKED ISSUE #618: Can't navigate with keyboard through non-editable cells
Using the standalone app (for simplicity, will be the same with callbacks).
Not sure when this regression appeared, probably `<=1.0.0`.

Modify `eee` and `fff` columns in `data.ts` with

```
editable: false
```

Navigating with the keyboard, upon reaching columns `eee` or `fff`, won't be able to move from the cell afterwards.

Marc-Andre-Rivet:
Fixes #618 

New test, failure prior to fix: https://circleci.com/gh/plotly/dash-table/14271
Passes after fix: https://circleci.com/gh/plotly/dash-table/14278

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "dash-table",
-  "version": "4.3.0",
+  "version": "4.4.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/dash-table/components/CellLabel/index.tsx
+++ b/src/dash-table/components/CellLabel/index.tsx
@@ -3,6 +3,7 @@ import React, {
 } from 'react';
 
 interface IProps {
+    active: boolean;
     className: string;
     value: any;
 }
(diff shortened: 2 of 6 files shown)


src/dash-table/components/CellLabel/index.tsx
```diff
@@ -15,11 +16,34 @@ export default class CellLabel extends PureComponent<IProps> {
         } = this.props;
 
         return (<div
+            ref='el'
             className={className}
+            tabIndex={-1}
```
Marc-Andre-Rivet: this makes the element focusable

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-table",
-  "version": "4.3.0",
+  "version": "4.4.0",
```
alexcjohnson: 🙈

src/dash-table/components/CellLabel/index.tsx
alexcjohnson: neat, I didn't know about that
https://webaim.org/techniques/keyboard/tabindex

> A tabindex="-1" value removes the element from the default navigation flow and also allows it to receive programmatic focus.

Seems like the right call (as opposed to 0, keeping it in the tab "navigation flow")

PR #621, opened 2019-10-09
Issue 611 - Hidden columns on copy/paste

### LINKED ISSUE #611: Hideable columns disappear after pasting data into table?
as per https://community.plot.ly/t/hideable-columns-disappear-after-pasting-data-into-table/29342

i haven't attempted to reproduce this.

Marc-Andre-Rivet:
Fixes #611 

Similarly to what's being done elsewhere, the clipboard needs to take into consideration both the visible columns and all the available columns when handling a copy/paste operation.

The last visible column is used to determine placement and whether new columns are required, all columns are used to determine the final set of columns if new ones are needed and their final placement based on visible ones.

Added two tests to the clipboard to check the behavior when there are hidden columns.

### DIFF
--- a/src/dash-table/utils/applyClipboardToData.ts
+++ b/src/dash-table/utils/applyClipboardToData.ts
@@ -9,7 +9,8 @@ export default (
     values: any[][],
     activeCell: ICellCoordinates,
     derived_viewport_indices: number[],
-    columns: Columns,
+    columns_: Columns,
+    visibleColumns: Columns,
     data: Data,
     overflowColumns: boolean = true,
     overflowRows: boolean = true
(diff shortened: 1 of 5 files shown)


src/dash-table/utils/applyClipboardToData.ts
```diff
@@ -23,29 +24,41 @@ export default (
...
             newData.forEach(row => (row[`Column ${i}`] = ''));
         }
+
+        newColumns = R.insertAll(
+            R.indexOf(R.last(visibleColumns), columns_) + 1,
+            _newColumns,
+            newColumns
+        );
+
+        newVisibleColumns = R.concat(newVisibleColumns, _newColumns);
```
Marc-Andre-Rivet: When new columns are created, they are appended after the last visible column. Both visible and "all" columns are updated to both (1) handle the copy/paste below and (2) the columns update correctly.

tests/cypress/tests/unit/clipboard_test.ts
```diff
@@ -3,13 +3,75 @@ import * as R from 'ramda';
...
+            );
+
+            expect(res).to.not.equal(undefined);
+
+            if (res) {
+                expect(res.data.length).to.equal(2);
+                expect(Object.entries(res.data[0]).length).to.equal(3);
+                expect(res.columns.length).to.equal(3);
+                expect(res.columns[0].id).to.equal('c1');
+                expect(res.columns[1].id).to.equal('c2');
```
Marc-Andre-Rivet: Check that all columns are present and in the right order after a copy/paste operation involving hidden columns

src/dash-table/utils/applyClipboardToData.ts
alexcjohnson: hmm OK - I might have added the new columns after trailing hidden columns as well, but it's probably impossible to declare one way objectively better than the other. This is fine.

PR #623, opened 2019-10-09
Issue 206 - Filter copy/paste

### LINKED ISSUE #206: copying and pasting strings into the `filter` cell doesn't work?

Marc-Andre-Rivet:
Fixes #206 

Allow copy/paste from/to filter columns inputs.

**Note**: Sadly there's no good way to test this through Cypress. The table cell copy could be done because of the way the data is copied and pasted, could be shimmed with some tweaking. There's no clear path here.

### DIFF
--- a/src/dash-table/components/Filter/Column.tsx
+++ b/src/dash-table/components/Filter/Column.tsx
@@ -3,6 +3,7 @@ import React, { CSSProperties, PureComponent } from 'react';
 import IsolatedInput from 'core/components/IsolatedInput';
 
 import { ColumnId } from 'dash-table/components/Table/props';
+import TableClipboardHelper from 'dash-table/utils/TableClipboardHelper';
 
 type SetFilter = (ev: any) => void;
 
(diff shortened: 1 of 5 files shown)


src/dash-table/utils/TableClipboardHelper.ts
```diff
@@ -42,6 +42,12 @@ export default class TableClipboardHelper {
         Clipboard.set(e, value);
     }
 
+    public static clearClipboard() {
+        Clipboard.clear();
+        TableClipboardHelper.lastLocalCopy = [];
+        TableClipboardHelper.localCopyWithoutHeaders = [];
+    }
```
Marc-Andre-Rivet: Deal with the special table copy/paste code -- reset the previsouly set copies -- will pick up the last copy event normally when doing a paste from filter columns.

src/dash-table/components/Filter/Column.tsx
```diff
@@ -51,6 +52,13 @@ export default class ColumnFilter extends PureComponent<IColumnFilterProps, ISta
             style={style}
         >
             <IsolatedInput
+                onCopy={(e: any) => {
+                    e.stopPropagation();
+                    TableClipboardHelper.clearClipboard();
+                }}
+                onPaste={(e: any) => {
+                    e.stopPropagation();
```
Marc-Andre-Rivet: prevent copy/paste from being handled by the global table handler

PR #632, opened 2019-10-24
Async table component

Marc-Andre-Rivet:
Follow up to https://github.com/plotly/dash/pull/973 to be completed and `@plotly/webpack-dash-dynamic-import` to be updated / published in order to be completed.

Makes the entire table component async. As the table has side-effects (like the dcc.Graph), it needs to inform the renderer by using the `@plotly/dash-component-plugins` decorator.

### DIFF
--- a/dash_table_base/__init__.py
+++ b/dash_table_base/__init__.py
@@ -60,6 +60,22 @@ _js_dist = [
         ).format(__version__),
         'namespace': package_name,
         'dynamic': True
+    },
+    {
+        'relative_package_path': 'async~table.js',
+        'external_url': (
+            'https://unpkg.com/dash-table@{}/dash_table/async~table.js'
+        ).format(__version__),
+        'namespace': package_name,
+        'async': True
+    },
+    {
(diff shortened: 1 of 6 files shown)


dash_table_base/__init__.py
```diff
@@ -60,6 +60,22 @@
         ).format(__version__),
         'namespace': package_name,
         'dynamic': True
+    },
+    {
+        'relative_package_path': 'async~table.js',
+        'external_url': (
+            'https://unpkg.com/dash-table@{}/dash_table/async~table.js'
```
alexcjohnson: Oh... is external serving going to work with fingerprints?

alexcjohnson:
Seems straightforward... but I thought earlier you decided against doing this, like it wasn't going to help that much. What changed your mind?

Marc-Andre-Rivet:
@alexcjohnson It doesn't help much for a small app but I came to realize that for a large modular app like the docs, doing this across all core component packages greatly reduces the qty of JS that needs to be fetched in order for the app to start working.

PR #637, opened 2019-11-08
IE11 compatibility

Marc-Andre-Rivet:
Sibling of https://github.com/plotly/dash/pull/1006

Also fixes https://github.com/plotly/dash-table/issues/641


.circleci/config.yml
```diff
@@ -74,12 +74,15 @@ jobs:
 
             - run:
                 name: Run tests
-                command: npm run test.standalone
+                command: |
+                    rm -rf node_modules/cypress
+                    npm i cypress@3.4.1
+                    npm run test.standalone
```
Marc-Andre-Rivet: Updating Chrome from `71` to `78` and Cypress from `3.4.1` to 3.6.1` here seems to cause
- test performance degredation that accumulates leading to tests timing out eventually
- a different behavior that I can explain causing ~15% of tests to fail

While unit and server test suites are now running against latest, leaving the standalone tests as-is in its "legacy" state for now. I can't tell what's going on atm.

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ jobs:
                 key: deps1-{{ .Branch }}-{{ checksum "package-lock.json" }}-{{ checksum "package.json" }}-{{ checksum ".circleci/config.yml" }}
             - run:
                 name: Install npm packages
-                command: npm install
+                command: npm ci
```
Marc-Andre-Rivet: As good a place to start using `npm ci` as any. Provided that the lock file and package.json are aligned, will use the versions from the lock file. Will otherwise fail the build.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 jobs:
     "server-test":
         docker:
-            - image: circleci/python:3.6.7-node-browsers
+            - image: circleci/python:3.6-node-browsers
```
Marc-Andre-Rivet: Unlocking to use latest.

babel.config.js
```diff
@@ -1,22 +1,10 @@
...
-                'last 2 Firefox versions',
-                'last 2 Safari versions',
-                'last 2 Edge versions',
-                'Explorer 11'
-            ]
-        },
-        useBuiltIns: 'usage',
-        corejs: 3
-    }],
+    '@babel/preset-env',
```
Marc-Andre-Rivet: Making it consistent with the rest of Dash

src/dash-table/components/Table/index.tsx
```diff
@@ -49,24 +50,45 @@ export default class Table extends Component<SanitizedAndDerivedProps, Standalon
...
+                let applyFocus = false;
+
+                const activeElement = document.activeElement as HTMLElement;
+                if (activeElement) {
+                    applyFocus = !!DOM.getParentById(activeElement, this.props.id);
+                }
+
+                if (applyFocus !== currentApplyFocus) {
+                    nextState.applyFocus = applyFocus;
+                }
```
Marc-Andre-Rivet: State will be updated when next render is run.

Only foce refocus when (1) the active_cell has changed, (2) the focus is currently inside the table, otherwise this means the focus is elsewhere and there's either no point forcing focus or it will cause usability issues..

src/dash-table/components/Table/props.ts
```diff
@@ -410,6 +411,7 @@ export type SanitizedProps = Omit<Omit<
     Merge<PropsWithDefaults, {
         fixed_columns: number;
         fixed_rows: number;
+        loading_state: boolean;
```
Marc-Andre-Rivet: Synthetize the `loading_state` object to a boolean representing whether `data` is being worked on. Prevents a bunch of useless re-renders.

tests/cypress/tests/server/select_props_test.ts
```diff
@@ -115,7 +115,6 @@ describe('select row', () => {
             expectCellSelection([1, 2, 3], [3001, 3002, 3003], [0, 1], [0, 1], [2, 1], [2, 1]);
 
             // shrink the selection
-            DOM.focused.type(Key.Shift, { release: false });
```
Marc-Andre-Rivet: Cypress `3.6.x` API changed in subtle ways, doing two consecutive non-release shift key presses is actually two presses, needs to be removed for the selection to behave as expected.

tests/cypress/support/index.js
```diff
@@ -14,6 +14,7 @@
 // ***********************************************************
 
 // Import commands.js using ES2015 syntax:
+import '@babel/polyfill';
```
Marc-Andre-Rivet: The server build gets polyfills from the renderer, the standalone build adds it to the entrypoint. Unit tests get it from cypress, here.

.circleci/config.yml
alexcjohnson: Sure, I'm OK with this as long as we make an issue and come back to it in not too long - perhaps around Cypress 3.7?

src/dash-table/components/Table/index.tsx
```diff
@@ -49,24 +50,45 @@ export default class Table extends Component<SanitizedAndDerivedProps, Standalon
...
+                    );
+
+                    if (map !== currentMap) {
+                        nextState.workFilter = { map, value };
+                    }
+                }
+            }
+
+            // state for applying focus
+            if (nextProps.active_cell !== this.props.active_cell) {
```
alexcjohnson: what if `active_cell` was cleared? Maybe this is OK because there's nothing that `applyFocus` would actually focus *on*?

src/dash-table/components/Table/index.tsx
```diff
@@ -49,24 +50,45 @@ export default class Table extends Component<SanitizedAndDerivedProps, Standalon
...
+                if (!currentApplyFocus) {
+                    nextState.applyFocus = true;
+                }
+            } else if (nextProps.loading_state !== this.props.loading_state) {
```
alexcjohnson: you're going to have to explain this one to me... why does entering *or* exiting loading trigger this check?

tests/cypress/tests/unit/exportUtils_tests.ts
```diff
@@ -49,92 +50,92 @@ describe('export', () => {
     describe('getMergeRanges', () => {
         it('no merge', () => {
             const testedArray = [['a', 'b', 'c', 'd'],
-                                 ['a', 'b', 'c', 'd'],
-                                 ['a', 'b', 'c', 'd']];
+            ['a', 'b', 'c', 'd'],
+            ['a', 'b', 'c', 'd']];
```
alexcjohnson: NBD but did a linter do that? Seems markedly less readable than some flavor of aligned.

.circleci/config.yml
Marc-Andre-Rivet: https://github.com/plotly/dash-table/issues/642

src/dash-table/components/Table/index.tsx
Marc-Andre-Rivet: Exactly. No `active_cell`, nothing to focus.

src/dash-table/components/Table/index.tsx
Marc-Andre-Rivet: If a callback modifying `data` is triggered and the focus is on a table cell, the table needs to be re-rendered once the callback returns to make the cell writeable again.

tests/cypress/tests/unit/exportUtils_tests.ts
Marc-Andre-Rivet: I have VSCode auto-formatting turned on, must have done that when I added the import https://github.com/plotly/dash-table/pull/637/files/28c73dd0f967b7a5ed23092a95a4f7509567d307#diff-d17cae001a0125bd53a0f8ff682e933eR4

Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/637/commits/9e46f0df6faf9e51e4373793a8400ec927549af8

tests/cypress/tests/server/loading_uneditable_test.ts
```diff
@@ -22,34 +22,56 @@ describe('loading states uneditable', () => {
...
+        DashTable.getCell(0, 0, State.Ready);
+
+        DOM.focused.type(`Hello`);
         DashTable
             .getCell(0, 0)
-            .click()
-            .find('.dash-input-cell-value-container > input').should('have.length', 1);
+            .within(() => cy.get('.dash-cell-value')
+                .should('have.value', 'Hello'));
+    });
```
Marc-Andre-Rivet: Made the test a bit tighter by skipping Enter / Click, gives the table less chance to render multiple times and hide issues for renders with a smaller footprint (less props changed)

alexcjohnson: I'm not excited about the `cy.wait(5000)` statements that are proliferating here... I won't block on this, given all the cypress headaches already in this PR, but it would be great if at some point we could figure out a condition to wait for rather than a time.

Marc-Andre-Rivet: Reverting. I ran this in `3.4.1` but in `3.6.1` I'm encountering weird timing issues.

Marc-Andre-Rivet: About the `cy.wait(5000)` the callback waits for 5 seconds and the maximum default timeout for a selector to succeed in Cypress is 4000ms. It is a bit yucky.

PR #654, opened 2019-11-27
Fix mismatched heights on row_selectable/row_deletable columns. 

### LINKED ISSUE #649: document that using DataTable `row_selectable` and `row_deletable` with fixed columns results in mismatched row heights
Discovered while writing tests for #606.

Running this example: 
```python
import dash
import dash_table
import pandas as pd
import random

df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder2007.csv')

app = dash.Dash(__name__)

data = df.to_dict('records')
for d in data:
    d['country'] = (d['country']+'\n') * random.randint(1, 10)

app.layout = html.Div([
    dash_table.DataTable(
        id='datatable-interactivity',
        columns=[
            {"name": i, "id": i, "deletable": True, "selectable": True} for i in df.columns
        ],
        data=data,
        fixed_rows={'headers': True},
        fixed_columns={'headers': True},
        column_selectable="single",
        row_selectable="multi",
        row_deletable=True,
        page_action="native",
        page_current=0,
        page_size=100,
        style_data={
            'whiteSpace': 'pre',
            'height': 'auto'
        }
    )
])

app.run_server(debug=True)
```

Results in the following: 

2019-11-21 12 47 24 localhost e342bec8b2a8

It appears as though the height of all rows in the `row_selectable` and `row_deletable` columns is set by the first cell in the first row that is not in a fixed column.

shammamah:
Closes #649.

Before: 
Screen Shot 2019-11-27 at 10 52 37 AM
After: 
Screen Shot 2019-11-27 at 10 51 16 AM

### DIFF
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -679,6 +679,17 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
             });
         }
 
+	// Adjust the height of rows in the row_selectable and row_deletable columns
+	if (fixed_columns || (row_selectable || row_deletable)) {
+	    Array.from(r1c1.querySelectorAll(`tr`)).forEach((tr, index) => {
+		const style = getComputedStyle(tr);
+		const height = style.height;
+
+		this.stylesheet.setRule(`.dash-fixed-column:not(.dash-fixed-row) tr:nth-of-type(${index + 1})`,
+					`height: ${height} !important;`);
+	    });
+	}
+
         // Adjust the width of the fixed row / fixed columns header
         if (fixed_columns && fixed_rows) {
             Array.from(r1c0.querySelectorAll('tr:first-of-type td, tr:first-of-type th')).forEach((td, index) => {
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -680,7 +680,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         }
 
         // Adjust the height of rows in the row_selectable and row_deletable columns
-        if (fixed_columns || (row_selectable || row_deletable)) {
+        if (fixed_columns || row_selectable || row_deletable) {
             Array.from(r1c1.querySelectorAll(`tr`)).forEach((tr, index) => {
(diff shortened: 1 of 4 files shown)


src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -679,6 +679,17 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+        if (fixed_columns || (row_selectable || row_deletable)) {
+            Array.from(r1c1.querySelectorAll(`tr`)).forEach((tr, index) => {
+                const style = getComputedStyle(tr);
+                const height = style.height;
+
+                this.stylesheet.setRule(`.dash-fixed-column:not(.dash-fixed-row) tr:nth-of-type(${index + 1})`,
+                    `height: ${height} !important;`);
+            });
+        }
+
```
Marc-Andre-Rivet: Sadly that's probably the only we can do. Concerns on my side are mostly about performance.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -679,6 +679,17 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
             });
         }
 
+        // Adjust the height of rows in the row_selectable and row_deletable columns
+        if (fixed_columns || (row_selectable || row_deletable)) {
```
Marc-Andre-Rivet: Good thinking, escaping processing for no fixed_columns. No point wrapping row_selectable || row_deletable.

src/dash-table/components/ControlledTable/index.tsx
Marc-Andre-Rivet: Tried the issue example with 
```
data = data * 4
```

For 568 rows, initial table rendering took ~25 seconds 😢 mostly in rendering / relayout

shammamah:
Closing this PR as it will most likely require a more involved process to deal with the performance issues that this introduces (I suspect that the main culprit is the multiple calls to `this.stylesheet.setRule`, since it sets only one rule at a time. 

What was changed/added in this PR: 
* rows that belonged to the "fixed columns" sub-table were given indices like the non-fixed-column sub-table 
* similar logic was used to adjust the row heights as was used to adjust column widths for fixed rows

PR #664, opened 2019-12-12
Update Werkzeug and add pyyaml

wbrgss:
When running `npm run build` in this repo, within a virtual environment:

- Without upgrading Werkzeug to `0.15.6` I got 

```sh
  from .wrappers import Request
  File "/home/plotly/Workspace/dash-table/ve/lib/python3.7/site-packages/flask/wrappers.py", line 14, in <module>
  from werkzeug.wrappers.json import JSONMixin as _JSONMixin
ModuleNotFoundError: No module named 'werkzeug.wrappers.json'; 'werkzeug.wrappers' is not a package
```

Maybe we could also use a different version of Flask. In the end it might be a Flask bug (see [here](https://stackoverflow.com/questions/56933523/apache-airflow-airflow-initdb-throws-modulenotfounderror-no-module-named-wer))

- Without installing `pyyaml` from pip:

```sh
Traceback (most recent call last):
  File "/home/plotly/Workspace/dash-table/ve/bin/dash-generate-components", line 5, in <module>
    from dash.development.component_generator import cli
  File "/home/plotly/Workspace/dash-table/ve/lib/python3.7/site-packages/dash/development/component_generator.py", line 14, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

I'm using PopOS (a flavour of Ubuntu 19.04). @Marc-Andre-Rivet can you check if bumping the Werkzeug version breaks anything in your env and/or whether `pyyaml` is already in your env?


requirements.txt
```diff
@@ -49,6 +50,6 @@ traitlets==4.3.2
 typed-ast==1.1.0
 urllib3==1.23
 wcwidth==0.1.7
-Werkzeug==0.14.1
+Werkzeug==0.15.6
 wrapt==1.10.11
-xlrd>= 1.0.0
\ No newline at end of file
+xlrd>= 1.0.0
```
wbrgss: Not sure what happened here; whitespace strip?

alexcjohnson: Your editor probably added a trailing newline, which @Marc-Andre-Rivet's files often don't have. No worries, I do that all the time too 😏

alexcjohnson:
@Marc-Andre-Rivet What is this `requirements.txt` file? The only place I see it mentioned is in the "local server review apps" section of `CONTRIBUTING.md`. For running `npm run build` I'd think you want to be in an environment that has `dash[dev]` installed, and that'll bring along `PyYAML` with it (and then after that reinstall the table repo `-e`).

Re: Werkzeug - I guess that should have been updated with Flask in https://github.com/plotly/dash-table/pull/536, but I don't know why this `requirements.txt` file is locking down things we don't depend on directly, that seems like a recipe for all sorts of problems, unless this file is autogenerated from some other clean install process

Marc-Andre-Rivet:
@alexcjohnson I think this file is used for the Heroku deployment from GH:
image

Not sure if `dev-requirements` and `python-requirements` have any use at this point though.

shammamah:
@wbrgss `pyyaml` should be installed automatically with `pip install dash[dev]`. We might have to update the contributing guidelines instead to specify that that package should be installed.

Marc-Andre-Rivet:
Removing the `pyyaml` entry. As @shammamah pointed out this is included with `dash[dev]`.

PR #665, opened 2019-12-12
Fix short Firefox cell height for dropdown cells

### LINKED ISSUE #663: Firefox doesn't allow as much height for cells, leading to overflow
In my version of Firefox, a dropdown within a cell **with a new value selected** (i.e. not on initial render) will overflow its cell container. The bottom grey border is overlapped:

**Firefox**:
dropdown-overflow

**Chrome**
dropdown-overflow-chrome

wbrgss:
Fixes #663 (issue described in detail there; TLDR Firefox table cell height is off)

~Solution from https://stackoverflow.com/questions/3215553/make-a-div-fill-an-entire-table-cell#comment40719119_7600749~

I don't think this has any side effects with Chrome; i.e. it only helps parity between Chrome and FF rendering.


wbrgss:
**Before**
tab

**After**
fixed-table-height

wbrgss:
Looking at Percy, this seems to fix what I wanted on Firefox (and there are 11 more Firefox changes in the diff), but it also seems to shorten rows without dropdowns (?), [including in Chrome](https://percy.io/plotly/dash-table/builds/3467297/view/212249404/1280?mode=diff&browser=chrome&snapshot=212249404).

wbrgss:
OK, I changed the style rules and got it down to 11 Firefox diffs and 0 Chrome diffs, which is what I want. Please review when you have a chance @Marc-Andre-Rivet

Marc-Andre-Rivet:
@wbrgss Updated the branch and added a changelog file

PR #677, opened 2020-01-11
feat: disable cookie access under restricted sandboxes

josegonzalez:
When dash is embedded into an iframe with a sandbox attribute that only has allow-scripts, cookie access is disabled and dash-table fails to load. As such, we need to restrict our cookie usage by disabling functionality.

This patch removes the disabled functionality in a graceful manner, allowing dash-table to load in very restricted iframes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - [#670](https://github.com/plotly/dash-table/pull/670) Fix a bug where `derived_filter_query_structure` was not getting updated properly
+- [#677](https://github.com/plotly/dash-table/pull/677) Fix a bug where the table fails to load when used inside an iframe with a sandbox attribute that only has allow-scripts
 
 ## [4.5.1] - 2019-11-14
 ### Fixed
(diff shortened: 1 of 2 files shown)


src/core/storage/Cookie.ts
```diff
@@ -1,8 +1,31 @@
...
+            // Create cookie
+            document.cookie = 'cookietest=1';
+            let ret = document.cookie.indexOf('cookietest=') !== -1;
+            // Delete cookie
+            document.cookie = 'cookietest=1; expires=Thu, 01-Jan-1970 00:00:01 GMT';
+            return ret;
+        } catch (e) {
+            return false;
+        }
+    });
```
Marc-Andre-Rivet: @josegonzalez As I'm nearing a release and would want to get this in, took the liberty of updating this code so that the `CookieStorage.enbled` runs only once.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - [#670](https://github.com/plotly/dash-table/pull/670) Fix a bug where `derived_filter_query_structure` was not getting updated properly
+- [#677](https://github.com/plotly/dash-table/pull/677) Fix a bug where the table fails to load when used inside an iframe with a sandbox attribute that only has allow-scripts
```
Marc-Andre-Rivet: @josegonzalez Added a changelog entry

PR #696, opened 2020-02-18
Issue 681 - Clean up server-side tests

### LINKED ISSUE #582: Rewrite Cypress test servers using gunicorn
The table currently uses a set of independent Dash applications to run its end-to-end tests against.

https://github.com/plotly/dash-table/tree/master/tests/cypress/dash
https://github.com/plotly/dash-table/blob/master/package.json#L25

Use gunicorn to load all the test apps into a single app with routes instead.

### LINKED ISSUE #681: Clean up server-side tests
Server-side tests are built using n-apps for different flavours of the table (e.g. https://github.com/plotly/dash-table/blob/dev/package.json#L25) -- instead, use a single app with routing to simplify writing & running server-side tests for the table.

Alternatively, investigate rewriting the tests with Selenium. We've been having significant issues with Cypress for a while now and those issues don't look like they are going away. Work done in the last year in DCC also greatly improved our Selenium testing environment to the point that maintaining two different sets of testing environments might be moot.

Marc-Andre-Rivet:
Closes #582 
Closes #681 

While #582 was about moving to gunicorn, what proposed here is instead to put in place a new test infra using Selenium (1) to be consistent with other core projects, (2) use the testing infra work that was done over the last year with Selenium in Dash.

- Add DataTable mixin
- Expose selectors for cells / columns / etc. - to be completed when more tests will be added or transitioned
- Rewrite a subset of tests


package.json
```diff
@@ -23,22 +23,18 @@
     "private::build:js-test-watch": "run-s \"private::build -- --mode development --config webpack.test.config.js --watch\"",
     "private::build:py": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json --r-prefix 'dash'",
     "private::host_dash8081": "python tests/cypress/dash/v_be_page.py",
-    "private::host_dash8082": "python tests/cypress/dash/v_copy_paste.py",
     "private::host_dash8083": "python tests/cypress/dash/v_fe_page.py",
     "private::host_dash8084": "python tests/cypress/dash/v_data_loading.py",
     "private::host_dash8085": "python tests/cypress/dash/v_default.py",
     "private::host_dash8086": "python tests/cypress/dash/v_pagination.py",
-    "private::host_dash8087": "python tests/cypress/dash/v_markdown.py",
```
Marc-Andre-Rivet: These tests are converted to Selenium tests as part of this PR

tests/selenium/conftest.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # start server with app and pass Dash arguments
+        self.server(app, **kwargs)
+
+        # set the default server_url, it implicitly call wait_for_page
+        self.server_url = self.server.url
+
+
+@pytest.fixture
+def test(request, dash_thread_server, tmpdir):
```
Marc-Andre-Rivet: This is the `dash_duo`, `dash_dcc` equivalent for the table. As testing the table is mostly about interacting with discreet items inside the table component itself, this mixin aims to expose in an easy to use way the various parts of the table. To begin: cell and column headers.

It puts into application some lessons learned from the Cypress/JS helper so as to minimize finicking. For example, when one passes an integer vs a string for a cell column name the helper will figure out what the right selector is. Also, a lot of validations have to do with getting the text content of the cells, this is also streamlined so that it's no longer necessary to click the cell / know that the cell is editable or readonly, etc.

tests/selenium/conftest.py
```diff
@@ -0,0 +1,196 @@
...
+from selenium.webdriver.support.wait import WebDriverWait
+
+_validate_col = lambda col: (isinstance(col, str) and len(col) > 0) or (isinstance(col, int) and col >= 0)
+_validate_col_id = lambda col_id: isinstance(col_id, str) and len(col_id) > 0
+_validate_id = lambda id: isinstance(id, str) and len(id) > 0
+_validate_key = lambda key: isinstance(key, str) and len(key) == 1
+_validate_row = lambda row: isinstance(row, int) and row >= 0
+_validate_state = lambda state: state in [_READY, _LOADING, _ANY]
+_validate_target = lambda target: isinstance(target, DataTableFacade)
+_validate_mixin = lambda mixin: isinstance(mixin, DataTableMixin)
```
Marc-Andre-Rivet: Use preconditions to validate the inputs. The param name in the lambda needs to match the param name in the function, hence all keys should be `key`, etc.

tests/selenium/conftest.py
```diff
@@ -0,0 +1,196 @@
...
+
+_READY = '.dash-spreadsheet:not(.dash-loading)'
+_LOADING = '.dash-spreadsheet.dash-loading'
+_ANY = '.dash-spreadsheet'
```
Marc-Andre-Rivet: Since the operations are async, it sometimes matters what the state of the table is. These are extra selectors that are used to make sure the table is ready (or not). Typically `state = _READY` is used by default.

tests/selenium/conftest.py
```diff
@@ -0,0 +1,196 @@
...
+        self.target = target
+
+    def __enter__(self):
+        return self.target
+
+    def __exit__(self, type, value, traceback):
+        return
+
+
+class HoldKeyContext:
```
Marc-Andre-Rivet: Allows using contexts for holding keys, making it structurally clearer what's impacted by the "hold"

```
with test.hold(Keys.SHIFT):
  // do things

// SHIFT is released here
```

tests/selenium/conftest.py
```diff
@@ -0,0 +1,196 @@
...
+class DataTableContext:
+    @preconditions(_validate_target)
+    def __init__ (self, target):
+        self.target = target
+
+    def __enter__(self):
+        return self.target
+
+    def __exit__(self, type, value, traceback):
+        return
```
Marc-Andre-Rivet: Create a context to handle a specific table. Some tests require interactions between tables.

tests/selenium/conftest.py
```diff
@@ -0,0 +1,341 @@
...
+_validate_state = lambda state: state in [_READY, _LOADING, _ANY]
+_validate_target = lambda target: isinstance(target, DataTableFacade)
+
+_READY = '.dash-spreadsheet:not(.dash-loading)'
+_LOADING = '.dash-spreadsheet.dash-loading'
+_ANY = '.dash-spreadsheet'
+_TIMEOUT = 10
+
+class DataTableContext:
+    @preconditions(_validate_target)
```
alexcjohnson: Neat, I didn't know about the `preconditions` package, that's really nice! Given that these functions need to match argument names, I'd rather their definitions live closer to their usage (or in fact just declare them inline if they don't get repeated). Though I suppose the naming convention `_validate_<NAME>` for arg `<NAME>` makes it pretty clear what's going on once you know how `preconditions` works...

tests/selenium/conftest.py
```diff
@@ -0,0 +1,341 @@
...
+        ActionChains(self.driver).key_down(Keys.CONTROL).send_keys('c').key_up(
+            Keys.CONTROL
+        ).perform()
+
+    def paste(self):
+        ActionChains(self.driver).key_down(Keys.CONTROL).send_keys('v').key_up(
+            Keys.CONTROL
+        ).perform()
+
+    @preconditions(_validate_key)
```
alexcjohnson: redundant validation?

tests/selenium/conftest.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def copy(self):
+        ActionChains(self.driver).key_down(Keys.CONTROL).send_keys('c').key_up(
+            Keys.CONTROL
+        ).perform()
+
+    def paste(self):
+        ActionChains(self.driver).key_down(Keys.CONTROL).send_keys('v').key_up(
+            Keys.CONTROL
+        ).perform()
```
alexcjohnson: Can this use your nice hold context too? Like
```
with self.hold(Keys.CONTROL):
    self.driver.send_keys('v')
```

tests/selenium/test_basic_operations.py
```diff
@@ -0,0 +1,323 @@
...
+    with test.table('table') as target:
+        target.cell.click(0, 1)
+        test.send_keys(Keys.BACKSPACE)
+        test.send_keys(Keys.ENTER)
+
+        assert target.cell.get_text(0, 1) == ''
+
+
+# # https://github.com/plotly/dash-table/issues/700
+# def test_tbst017_delete_cell_updates_while_selected(test):
```
alexcjohnson: Are you intending to fix that bug in this PR, or just putting the test code here for re-enabling whenever that bug gets fixed? You can also do
```py
@pytest.mark.skip(reason="https://github.com/plotly/dash-table/issues/700")
def test_tbst017_delete_cell_updates_while_selected(test):
    ...
```

tests/selenium/conftest.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+class DataTableCellFacade(object):
+    @preconditions(_validate_id, _validate_mixin)
+    def __init__(self, id, mixin):
+        self.id = id
+        self.mixin = mixin
+
+    @preconditions(_validate_row, _validate_col, _validate_state)
+    def _get_cell_value(self, row, col, selector, state = _ANY):
```
alexcjohnson: If we ever add linting to our tests it'll tell you default args get no spaces `state=_ANY`.

Now might not be a bad time to do that actually, since there isn't that much existing Python test code that would need cleaning up.

tests/selenium/conftest.py
```diff
@@ -0,0 +1,341 @@
...
+
+    @preconditions(_validate_row, _validate_col, _validate_state)
+    def click(self, row, col, state = _ANY):
+        return self.get(row, col, state).click()
+
+    @preconditions(_validate_row, _validate_col, _validate_state)
+    def double_click(self, row, col, state = _ANY):
+        ac = ActionChains(self.mixin.driver)
+        ac.move_to_element(self._get_cell_value(row, col, state))
+        ac.pause(1) # sometimes experiencing incorrect behavior on scroll otherwise
```
alexcjohnson: That's 1 sec, right? That's an awfully long delay to just toss in... isn't there some condition we can wait for? I guess double clicks don't happen all that often in our tests, but still...

tests/selenium/test_basic_copy_paste.py
```diff
@@ -0,0 +1,212 @@
...
+            raise PreventUpdate
+
+    return app
+
+
+def test_tbcp001_copy_paste_callback(test):
+    test.start_server(get_app())
+
+    with test.table('table') as target:
+        target.cell.click(0, 0)
```
alexcjohnson: This usage is a bit funny grammatically, makes it look like you have on cell `target.cell` and you're clicking it at `(0, 0)` whatever that means. I suppose it would be a little clearer to say `target.cells.click(0, 0)` but what if we refactor it to something like `target.cell(0, 0).click()`? Same with `.row` and `.column`. But `.paging` is clear as is.

Marc-Andre-Rivet: I like it

tests/selenium/conftest.py
Marc-Andre-Rivet: Yes. 1 second. My thinking was that at least it's isolated. I agree it's terrible, I'll see if I can find some other way to make sure the UI is ready (when getting the last row, it takes a small amount of time to scroll to the desired cell).

tests/selenium/conftest.py
Marc-Andre-Rivet: Agreed

tests/selenium/test_basic_operations.py
Marc-Andre-Rivet: Nice, will update to use the decorator

tests/selenium/conftest.py
Marc-Andre-Rivet: Yeah, these validations are extremely redundant, that's why I created the lambdas with standardized naming. The disadvantage as you mention is that now the validation itself lives further away.

tests/selenium/test_basic_copy_paste.py
```diff
@@ -0,0 +1,212 @@
...
+        else:
+            raise PreventUpdate
+
+    return app
+
+
+def test_tbcp001_copy_paste_callback(test):
+    test.start_server(get_app())
+
+    with test.table('table') as target:
```
alexcjohnson: I know we talked about this already but now that I see it used I'm not sure the context manager is really the best way to expose individual tables - (nearly?) all the tests just create that context up front and never do anything outside of it, so it feels like it's just adding a level of indentation for little reason. What if we just turn
```
with test.table('table') as source, test.table('table2') as target:
```
into
```
source, target = test.table('table'), test.table('table2')
```
?

Marc-Andre-Rivet: It is fairly idiosyncratic. I can 🔪 it

tests/selenium/test_basic_operations.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/696/commits/ea38a8604e6ab7540b8e13eb2e8c5e5401a0f059

tests/selenium/conftest.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/696/commits/3d9cf3701730dd216cf0af1c41a4ff67ffba4153

tests/selenium/conftest.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/696/commits/854276e6f807c967ae411af8e5c86d5fd1e7b2f4

tests/selenium/conftest.py
Marc-Andre-Rivet: Nyes? It just happens to be implemented by a context that also knows about preconditions, but if the implementation changes it may not.

tests/selenium/test_basic_copy_paste.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/696/commits/6c36f1d77f935d78004f595dada5b425d3d9652b

tests/selenium/test_basic_copy_paste.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/696/commits/73092c7b929f4f2b2a3f997dd4bd47e4cc43f699

Marc-Andre-Rivet:
Lots of noise due to the update for flake and black + applying the format to everything.

PR #704, opened 2020-02-27
Issue 703 - Rename async modules

### LINKED ISSUE #703: Rename async modules
Same as https://github.com/plotly/dash-core-components/issues/745 but for the table.

Marc-Andre-Rivet:
Closes #703

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,9 @@ jobs:
                     . venv/bin/activate
                     npm run private::build:js-test
                     npm run private::build:py
-                    pip install -e .
+                    python setup.py sdist
+                    cd dist
+                    find . -name "*.gz" | xargs pip install --no-cache-dir --ignore-installed && cd ..
 
             - run:
                 name: Run tests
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -40,7 +40,9 @@ jobs:
                     . venv/bin/activate
                     npm run private::build:js-test
                     npm run private::build:py
-                    pip install -e .
+                    python setup.py sdist
+                    cd dist
+                    find . -name "*.gz" | xargs pip install --no-cache-dir --ignore-installed && cd ..
```
Marc-Andre-Rivet: Using the opportunity to install from tarball instead of the local folder `-e .`. This change helps identify issues when resource paths change and the MANIFEST.in file isn't synced.

alexcjohnson: Yeah that seems important... Is this an issue in any of our other repos? Perhaps a followup issue to check on this as well?

ManojSN:
nice!

PR #709, opened 2020-03-02
remove '#' from stylesheet id

alexcjohnson:
Just cleaning up a funny look - I'm assuming the `#` in the id wasn't intentional?
Screen Shot 2020-03-02 at 6 23 47 PM


alexcjohnson:
meh not so simple after all and not worth fighting with right now...

PR #720, opened 2020-03-13
Git hooks husky

Marc-Andre-Rivet:
Companion to https://github.com/plotly/dash/pull/1151

Husky
`Husky` is a JS package that automagically runs commands on git hooks. For this repo it will run the `lint` script and expect no errors before allowing a commit to go through.

For the hooks to be present, the developer will need to run `npm ci` or `npm install` locally when cloning the repo.


package.json
```diff
@@ -22,9 +22,12 @@
     "private::build:js-test-standalone": "run-s \"private::build -- --mode development --config webpack.test.standalone.config.js\"",
     "private::build:js-test-watch": "run-s \"private::build -- --mode development --config webpack.test.config.js --watch\"",
     "private::build:py": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json --r-prefix 'dash'",
+    "private::format.ts": "npm run private::lint.ts -- --fix",
+    "private::format.black": "black --exclude dash_table .",
     "private::host_js": "http-server ./dash_table -c-1 --silent",
-    "private::lint:ts": "tslint --project tsconfig.json --config tslint.json",
-    "private::lint:py": "flake8 --exclude=DataTable.py,__init__.py,_imports_.py dash_table tests && black --check --exclude dash_table .",
+    "private::lint.ts": "tslint --project tsconfig.json --config tslint.json",
+    "private::lint.flake": "flake8 --exclude=dash_table*,node_modules,venv",
```
alexcjohnson: so excluding `dash_table_base` too? we can't lint that?

Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/720/commits/cdd830886f5229362089f1fbedc7fd6bb4a35e51

.flake8
```diff
@@ -1,5 +1,6 @@
 [flake8]
 ignore = C901, E203, E266, E501, E731, W503
-max-line-length = 88
-max-complexity = 18
-select = B,C,E,F,W,T4
\ No newline at end of file
+select = B,C,E,F,W,T4
+per-file-ignores =
+    tests/*: E722, F811
+    dash_table_base/__init__.py: F401, F403
```
alexcjohnson: I'm fine with different rules for tests vs source files, but putting single-file overrides in the config feels confusing to me, as well as potentially overly broad  - can we just override these rules on the specific lines that need them?

Marc-Andre-Rivet: Sure, my thought was that by making it part of the configuration it would be harder to sneak an exception in. I'm fine either way. Updating.

PR #722, opened 2020-03-17
Issue 649 - Fix mismatched row height for fixed columns

### LINKED ISSUE #649: document that using DataTable `row_selectable` and `row_deletable` with fixed columns results in mismatched row heights

Marc-Andre-Rivet:
Closes #649 

This PR proposes to fix the misaligned rows issue by adding invisible cells in both the fixed columns and unfixed columns sections of the table. With some tweaks, this allows both fixed and unfixed cells to calculate their size based on the entire table's content, making sure that both size have the same size/height.

Afterwards the fixed rows can be coerced into having the correct width. With some extra tweaks for Firefox.

Percy browser versions have been updated for both `dash-table` and `dash-table-python-v0` in order for the snapshots to match local behavior.

A lot of the changes in the PR is fixing the test logic to handle the added `phantom-cell`.

--- 

Current implementation slows down rendering for fixed table significantly (about 2.5x slower) but is significantly better than previous attempts (about 10-20x slower). Still investigating what can be done to reduce the overhead.

Possible secondary strategies:
* Changing the default `page_size` from 250 to 100 would provide significant performance mitigation for default usage while not impacting usability significantly.


demo/data.ts
```diff
@@ -176,9 +176,15 @@ export const generateMixedMarkdownMockData = (rows: number) => unpackIntoColumns
...
+        data: gendata(i => [
             '```javascript',
-            'console.warn("this is a markdown cell")',
+            ...(i % 2 === 0 ?
+                ['console.warn("this is a markdown cell")'] :
+                [
+                    'console.log("logging things")',
+                    'console.warn("this is a markdown cell")'
+                ]
+            ),
```
Marc-Andre-Rivet: Changing the mock to have variable height Markdown rows

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -48,6 +48,9 @@ const INNER_STYLE = {
     minWidth: '100%'
 };
 
+const WIDTH_EPSILON = 0.5;
+const MAX_WIDTH_ITERATIONS = 30;
```
Marc-Andre-Rivet: 30 is a very high number but most of the time it takes 2-6 iterations for the table's column width to settle.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -146,7 +149,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
 
     componentDidUpdate() {
         this.updateStylesheet();
-        this.applyStyle();
```
Marc-Andre-Rivet: `applyStyle` is removed / replaced by new code to handle the fixed rows

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -261,13 +266,73 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+                const firstVisibleTd = r1c1.querySelector(`tr:first-of-type > *:nth-of-type(${fixed_columns + 1})`);
+                if (firstVisibleTd) {
+                    const r1c1FragmentBounds = r1c1.getBoundingClientRect();
+                    const firstTdBounds = firstVisibleTd.getBoundingClientRect();
+
+                    const width2 = firstTdBounds.left - r1c1FragmentBounds.left;
+                    r0c1.style.marginLeft = `-${width2}`;
+                    r0c1.style.marginRight = `${width2}`;
+                    r1c1.style.marginLeft = `-${width2}`;
+                    r1c1.style.marginRight = `${width2}`;
```
Marc-Andre-Rivet: If there are fixed columns, align the cell's table in such a way as to make the first visible column left-aligned with the cell.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -718,7 +700,10 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         const { r0c1 } = this.refs as { [key: string]: HTMLElement };
 
         Logger.trace(`ControlledTable fragment scrolled to (left,top)=(${ev.target.scrollLeft},${ev.target.scrollTop})`);
-        r0c1.style.marginLeft = `${-ev.target.scrollLeft}px`;
+
+        const margin = parseFloat(ev.target.scrollLeft) + parseFloat(getComputedStyle(r0c1).marginRight);
```
Marc-Andre-Rivet: `margin-left` and `margin-right` match in dimension, if not sign. Scroll adds to the initial offset.

src/dash-table/components/Filter/Column.tsx
```diff
@@ -39,15 +39,15 @@ export default class ColumnFilter extends PureComponent<IColumnFilterProps, ISta
...
+            className,
             columnId,
             isValid,
             style,
             value
         } = this.props;
 
         return (<th
-            className={classes + (isValid ? '' : ' invalid')}
+            className={className + (isValid ? '' : ' invalid')}
```
Marc-Andre-Rivet: Standardizing prop name across all components simplifies manipulation elsewhere.

src/dash-table/components/Table/Table.less
```diff
@@ -141,337 +142,353 @@
...
-	    margin-right: 5px;
-
-	    &:hover {
-		color: hotpink;
-		
-		&:disabled {
-		    color: graytext
-		}
-	    }
+        display: inline-block;
```
Marc-Andre-Rivet: Indentation here was all weirded out - fixed. Lots of noise.

src/dash-table/derived/table/fragments.tsx
```diff
@@ -19,6 +20,26 @@ function renderFragment(cells: any[][] | null, offset: number = 0) {
...
+
+const getLastOfType = (cell: JSX.Element) => React.cloneElement(cell, {
+    ...cell.props,
+    className: cell.props.className ? `${cell.props.className} last-of-type` : 'last-of-type'
+});
+
+const getFirstPhantom = (cell: JSX.Element) => React.cloneElement(cell, {
+    ...cell.props,
+    style: { ...cell.props.style, borderLeft: 'transparent' }
+});
```
Marc-Andre-Rivet: Various component cloning / manipulations.

`getHiddenCell` makes a phantom version of a cell - will be invisible
`getFirstPhantom` solves a doubled border issue

tests/cypress/src/DashTable.ts
```diff
@@ -25,27 +25,35 @@ export class DashTableHelper {
     }
 
     public getCell(row: number, column: number, editable: State = State.Ready) {
-        return cy.get(`#${this.id} ${getSelector(editable)} tbody tr td.column-${column}`).eq(row);
+        return cy.get(`#${this.id} ${getSelector(editable)} tbody tr td.column-${column}:not(.phantom-cell)`).eq(row);
```
Marc-Andre-Rivet: Update test helper to ignore phantom cells

tests/visual/percy-storybook/MarkdownCells.percy.tsx
```diff
@@ -3,100 +3,134 @@ import { storiesOf } from '@storybook/react';
...
-                        '--- | ---',
-                        'This page has two tables | yes',
-                        'This table has two rows | no',
-                        'This is an example of tableception | yes'].join('\n'),
-                    b: 'simple two-column table'
-                }
-            ]}
-            columns={columnFormats}
-        />))
+const variants: ITest[] = [
```
Marc-Andre-Rivet: Markdown visual tests with both fixed an no fixed columns

tests/visual/percy-storybook/Sizing.percy.tsx
```diff
@@ -69,13 +69,13 @@ storiesOf('DashTable/Sizing', module)
...
     .add('widths smaller than content (forced)', () => (<DataTable
         {...props}
-        css={[{
-            selector: '.dash-cell[data-dash-column=Region]',
-            rule: 'white-space: normal'
-        }]}
         style_data_conditional={[{
+            if: { column_id: 'Region' },
+            whiteSpace: 'normal'
+        }]}
```
Marc-Andre-Rivet: Use the standard prop to do this

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -261,12 +267,81 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+                r0c0.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
+            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i]));
+
+            Array.from<HTMLElement>(
+                r0c1.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
+            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+
+            Array.from<HTMLElement>(
+                r0c1.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
+            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
```
Marc-Andre-Rivet, line 344: All this replaces and adds to the previous `applyStyle` method

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -261,12 +267,81 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-            this.stylesheet.setRule('.dash-fixed-column tr', `height: ${getComputedStyle(contentTr).height};`);
+                it++;
+
+                if (
+                    (Math.abs(currentWidth - lastWidth) < WIDTH_EPSILON && it % 2 === 0) ||
+                    it >= MAX_WIDTH_ITERATIONS
+                ) {
+                    break;
+                }
+            } while (true);
```
alexcjohnson, line 318: Not a big deal, but given that ultimately there's a known limit (`MAX_WIDTH_ITERATIONS`) I'd have probably written this as a `for` loop, with only the width test down here breaking the loop early.

src/dash-table/components/Table/Table.less
alexcjohnson: Looks like some got fixed and some other parts got mucked up. Mostly lines 507+ but there are a few before that that look weird as well. Is there an auto-formatter for `.less`?

src/dash-table/derived/table/fragments.tsx
```diff
@@ -1,6 +1,7 @@
 import * as R from 'ramda';
 import React from 'react';
 import { memoizeOneFactory } from 'core/memoizer';
+import isNil from 'ramda/es/isNil';
```
alexcjohnson, line 4: Why not just the one in `import * as R from 'ramda';` above? In fact you use both copies...

src/dash-table/derived/table/fragments.tsx
```diff
@@ -43,13 +59,32 @@ export default memoizeOneFactory((
...
+                acc => acc.count < fixedColumns,
+                (acc, cell) => {
+                    acc.cells++;
+                    acc.count += (cell.props.colSpan || 1);
+
+                    return acc;
+                },
+                { cells: 0, count: 0 },
+                row as any
+            ).cells;
```
alexcjohnson, line 84: 🌴`const pivot = ...` looks identical to above?

src/dash-table/components/Table/Table.less
alexcjohnson: OIC you have some hard tabs mixed in with spaces here - this file looks funny in #728 as well as a result.

src/dash-table/derived/table/fragments.tsx
Marc-Andre-Rivet: Weird.. fixed with https://github.com/plotly/dash-table/pull/722/commits/33db6f3df51dc5fa6dcefdaa9752eda7f9ff54d4

src/dash-table/derived/table/fragments.tsx
Marc-Andre-Rivet: 🌴 https://github.com/plotly/dash-table/pull/722/commits/3d0b8c87f5828a7c6395b2209281a1571e6e36d6

src/dash-table/components/Table/Table.less
Marc-Andre-Rivet: I'll double check to make sure this file has been cleaned up.. but yes, at some point it became a mix of tabs/spaces

src/dash-table/components/ControlledTable/index.tsx
Marc-Andre-Rivet: Want to keep both exit conditions together, that's why I opted for a `while` instead of a `for` loop. Did some clean up so the exit condition is now in the `while` condition instead of the body. https://github.com/plotly/dash-table/pull/722/commits/207c90ab62e2c558a2e9e84d4284532f58d27db1

Marc-Andre-Rivet: Also removed the `it % 2 == 0` check which is no longer necessary

alexcjohnson: OK, makes sense. Glad the `it % 2 === 0` can be avoided!

PR #725, opened 2020-03-23
Add fallback copy handler.

### LINKED ISSUE #526: Copy/paste does not work in certain situations
Clicking in cell in first column with the standalone demo code (npm run build.watch + localhost:8080) does not trigger a `copy`. Clicking on cell in the 2nd column triggers a `copy`. Navigating with mouse to first column from 2nd column cell, `copy` can now be triggered exactly once.

There is a global `onCopy` handler in the controlled table. Seems like in certain situations the cell's input element may be capturing that event instead.

This may be a regression. Unsure how long this has been a problem.

shammamah:
## About 
Closes #526

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - [#713](https://github.com/plotly/dash-table/pull/713) Update from React 16.8.6 to 16.13.0
 
+### Fixed
+- [#725](https://github.com/plotly/dash-table/pull/725) Fix bug in which copy events were not triggered except upon selection of a different cell.
+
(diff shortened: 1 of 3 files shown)


shammamah:
>We've discussed this IRL but can you explain here the difficulties creating a test case to lock this behavior?

This behaviour wasn't picked up on earlier because our existing copy/paste tests have all passed. The current tests all have to do with copying/pasting from the table to the table itself. This error has shown up upon copy from the table to paste in an external spreadsheet.

Marc-Andre-Rivet:
> This error has shown up upon copy from the table to paste in an external spreadsheet.

In that case, might be possible to test this by copying from the table to an input in the same page?

Marc-Andre-Rivet:
Superseded by https://github.com/plotly/dash-table/pull/728

PR #729, opened 2020-04-06
Issue 256, 257, 608 - Add conditional styling capabilities

### LINKED ISSUE #256: Ability to change the background color when a cell is selected
The default color is red on chrome,but white on Edge and IE

### LINKED ISSUE #257: Override data cell styling when focused / selected
Feature follow up to the bug reported in #256 

Adding ```isActive: boolean``` and ```isSelected: boolean``` to style_data_conditional entries would allow overriding the styling applied when a data cell is selected -- and allow greater visual customization.

### LINKED ISSUE #608: style_data_condtional - "row_index" won't accept vector of values
I am using `dashTable`  4.0.2  and below is an example of highlighting multiple `dashDataTable` rows in Dash for R.

```
library(dash)
library(dashCoreComponents)
library(dashHtmlComponents)

# Load data
data(mtcars)

# Create DashTable
carsDashDT <- dashDataTable(
  id = "cars-table",
  columns = lapply(colnames(mtcars),
                   function(colName){
                     list(
                       id = colName,
                       name = colName
                     )
                   }),
  data = df_to_list(mtcars),
  row_selectable = "multi",
)

# App Start ------------------------------

# Initate Application
app <- Dash$new()

# Create Layout ------------------------------

app$layout(
  htmlDiv(list(
    carsDashDT
  ))
)

# Callback Start ------------------------------

# Highlight dashTable rows
app$callback(
  output(id = "cars-table", property = "style_data_conditional"),
params = list(
  input(id = "cars-table", property = "selected_rows")
),

  function(rows) {
    style_data_conditional <- NULL

        for (i in 1:length(rows)) {
          rowIndex <- rows[[i]]
          style_data_conditional_append = list(list(
            "if" =  list("row_index" = rowIndex),
            "backgroundColor" = "#B0BED9"
          ))
          style_data_conditional <- c(style_data_conditional,
                                      style_data_conditional_append)
        }

    return(style_data_conditional)
  }
)

app$run_server()
```

The example provided at https://dashr.plot.ly/datatable/style under _Conditional Formatting - Highlighting Certain Rows_ demonstrates how to highlight a single row with a numeric input.

Based on the example provided in the above link, I experimented with multiple different syntax but couldn't achieve highlighting multiple rows without using a `loop`.

The documentation for _dashDataTable_ > _style_data_conditional_ mentions:

_row_index (numeric | a value equal to: 'odd', 'even'; optional)_ 

It seems like `row_index` is currently limited to accepting single numeric input. It would be very nice to be able to apply conditional styles to multiple rows by feeding a vector/list of values. Rather than the `loop` in the `callback`, something similar to:

```
 style_data_conditional_append = list(list(
    "if" =  list("row_index" = rows)
))
```

In case I am missing it, I would appreciate if you can provide the correct syntax. Thanks.

Marc-Andre-Rivet:
Closes #256 
Closes #257
Closes #608 

- Replaces use of `var` for selected/active cells background and edge colors (IE11 fix)
- Adds support for `row_index`, `header_index`, `column_id` array of values in `style_**_conditional` props
- Adds support for `state:'active'|'selected'` condition on `style_data_conditional` prop

https://github.com/plotly/dash-table/issues/752 for follow up `selected_row` support.


src/dash-table/conditional/index.ts
```diff
@@ -4,6 +4,11 @@ import { ColumnId, Datum, ColumnType, IColumn } from 'dash-table/components/Tabl
 import { QuerySyntaxTree } from 'dash-table/syntax-tree';
 import { IConvertedStyle } from 'dash-table/derived/style';
 
+export interface IActiveElement {
+    is_active?: boolean;
+    is_selected?: boolean;
+}
```
Marc-Andre-Rivet, line 10: New styling conditions

src/dash-table/conditional/index.ts
```diff
@@ -13,11 +18,11 @@ export interface IIndexedHeaderElement {
 }
 
 export interface IIndexedRowElement {
-    row_index?: number | 'odd' | 'even';
+    row_index?: number[] | number | 'odd' | 'even';
```
Marc-Andre-Rivet, line 21: Support array of values

src/dash-table/conditional/index.ts
```diff
@@ -13,11 +18,11 @@ export interface IIndexedHeaderElement {
...
 }
 
 export interface INamedElement {
-    column_id?: ColumnId;
+    column_id?: ColumnId[] | ColumnId;
```
Marc-Andre-Rivet, line 25: Support array of values

src/dash-table/conditional/index.ts
```diff
@@ -56,9 +75,11 @@ export function ifRowIndex(condition: IIndexedRowElement | undefined, rowIndex:
     }
 
     const rowCondition = condition.row_index;
-    return typeof rowCondition === 'number' ?
-        rowIndex === rowCondition :
-        rowCondition === 'odd' ? rowIndex % 2 === 1 : rowIndex % 2 === 0;
+    return typeof rowCondition === 'string' ?
+        rowCondition === 'odd' ? rowIndex % 2 === 1 : rowIndex % 2 === 0 :
+        Array.isArray(rowCondition) ?
```
Marc-Andre-Rivet, line 80: Check for array

src/dash-table/conditional/index.ts
```diff
@@ -68,9 +89,11 @@ export function ifHeaderIndex(condition: IIndexedHeaderElement | undefined, head
     }
 
     const headerCondition = condition.header_index;
-    return typeof headerCondition === 'number' ?
-        headerIndex === headerCondition :
-        headerCondition === 'odd' ? headerIndex % 2 === 1 : headerIndex % 2 === 0;
+    return typeof headerCondition === 'string' ?
+        headerCondition === 'odd' ? headerIndex % 2 === 1 : headerIndex % 2 === 0 :
+        Array.isArray(headerCondition) ?
```
Marc-Andre-Rivet, line 94: Check for array

src/dash-table/derived/cell/wrapperStyles.ts
```diff
@@ -16,28 +18,39 @@ const partialGetter = (
 ) => traverseMap2(
     data,
     columns,
-    (datum, column, i) => getDataCellStyle(datum, i + offset.rows, column)(styles)
+    (datum, column, i) => getDataCellStyle(datum, i + offset.rows, column, false, false)(styles)
```
Marc-Andre-Rivet, line 21: Partial getter now assume all cells are active=False, selected=False

src/dash-table/derived/cell/wrapperStyles.ts
```diff
@@ -16,28 +18,39 @@ const partialGetter = (
...
             return;
         }
 
-        styles[i][j] = R.merge(styles[i][j], SELECTED_CELL_STYLE);
+        const active = isActiveCell(activeCell, i, j);
+
+        const style = {
+            ...SELECTED_CELL_STYLE,
+            ...getDataCellStyle(data[i], i + offset.rows, columns[j], active, true)(styles)
+        };
```
Marc-Andre-Rivet, line 48: Combine default selected cell styling with calculated style

src/dash-table/derived/edges/data.ts
```diff
@@ -30,17 +34,57 @@ export default memoizeOneFactory((
     traverse2(
         data,
         columns,
-        (datum, column, i, j) => edges.setEdges(i, j, getDataCellEdges(datum, i + offset.rows, column)(styles))
+        (datum, column, i, j) => edges.setEdges(i, j, getDataCellEdges(
+            datum,
+            i + offset.rows,
+            column,
+            false,
+            false
```
Marc-Andre-Rivet, line 42: Since border calculations are now done on active cells too - need a partial calculation that assumes active=False, selected=False for all cells

src/dash-table/derived/edges/data.ts
```diff
@@ -30,17 +34,57 @@ export default memoizeOneFactory((
...
+) => {
+    if (!baseline) {
+        return baseline;
     }
 
+    const edges = baseline.clone();
+
+    R.forEach(({ row: i, column: j }) => {
+        const active = isActiveCell(activeCell, i, j);
+        const priority = active ? Number.MAX_SAFE_INTEGER : Number.MAX_SAFE_INTEGER - 1;
```
Marc-Andre-Rivet, line 66: `active` beats `selected`, so it gets a higher priority

src/dash-table/derived/edges/index.ts
```diff
@@ -4,7 +4,7 @@ import { Datum, IColumn } from 'dash-table/components/Table/props';
 import { matchesDataCell, matchesDataOpCell, matchesFilterCell, getFilterOpStyles, matchesHeaderCell, getHeaderOpStyles } from 'dash-table/conditional';
 import { traverse2 } from 'core/math/matrixZipMap';
 
-function resolveEdges(styles: IConvertedStyle[]): BorderStyle {
+function resolveEdges(styles: IConvertedStyle[], priority?: number): BorderStyle {
```
Marc-Andre-Rivet, line 7: Now allow for priority override - useful for the active/selected borders

src/dash-table/derived/edges/index.ts
```diff
@@ -14,15 +14,22 @@ function resolveEdges(styles: IConvertedStyle[]): BorderStyle {
             const border = s.style[p] || s.style.border;
 
             if (border) {
-                res[p] = [border, i];
+                res[p] = [border, priority ?? i];
```
Marc-Andre-Rivet, line 17: null-coalescing operator, equivalent to `isNil(priority) ? i : priority`

tests/visual/percy-storybook/Style.percy.tsx
```diff
@@ -374,4 +388,158 @@ storiesOf('DashTable/Style type condition', module)
...
+            if: {
+                is_selected: false
+            },
+            backgroundColor: 'lightblue',
+            border: '1px solid blue'
+        }]}
+        active_cell={{ row: 1, column: 1 }}
+        selected_cells={[
+            { row: 1, column: 1, column_id: 'b' }
+        ]}
```
Marc-Andre-Rivet, line 541: Bunch of new visual tests. There's no need for new unit tests on the EdgesMatrix as only usage changed, no additional logic/functionality was added.

src/dash-table/derived/edges/index.ts
alexcjohnson: Whoa, I'd never seen that before 🥇 That could simplify our lives in a million places! Can you confirm that it's properly transpiled for [IE and the other browsers that don't support it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)?

Marc-Andre-Rivet: Yup - the null-coalescing operator is a new addition to ESx, and supported by TypeScript since 3.7
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing.

This transpiles to (yes, old style `void 0` for undefined!):
```
if (border) {
    res[p] = [border, priority !== null && priority !== void 0 ? priority : i];
}
```

src/dash-table/derived/cell/wrapperStyles.ts
```diff
@@ -16,28 +18,47 @@ const partialGetter = (
...
         }
 
-        styles[i][j] = R.merge(styles[i][j], SELECTED_CELL_STYLE);
-    }, selectedCells);
+        const active = isActiveCell(activeCell, i, j);
+
+        const style = {
+            ...getDataCellStyle(data[i], i + offset.rows, columns[j], active, true)(inactiveStyles),
+            ...SELECTED_CELL_STYLE,
+            ...getDataCellStyle(data[i], i + offset.rows, columns[j], active, true)(activeStyles)
```
Marc-Andre-Rivet, line 55: Make an exception for active/selected, apply !active/!selected styles first, apply the baseline, then apply active/selected styles

src/dash-table/derived/edges/data.ts
```diff
@@ -30,17 +34,84 @@ export default memoizeOneFactory((
...
+            )(inactiveStyles),
+            ...baseStyle,
+            ...getDataCellEdges(
+                data[iWithOffset][column_id],
+                iWithOffset,
+                columns[j],
+                active,
+                true,
+                priority
+            )(activeStyles)
```
Marc-Andre-Rivet, line 107: Same here: apply !active/!selected edges, apply default active/selected edges, then apply active/selected

src/dash-table/conditional/index.ts
```diff
@@ -29,17 +34,31 @@ export interface IEditableElement {
...
+
 export function ifColumnId(condition: INamedElement | undefined, columnId: ColumnId) {
-    return !condition ||
-        condition.column_id === undefined ||
+    if (!condition || condition.column_id === undefined) {
+        return true;
+    }
+
+    return Array.isArray(condition.column_id) ?
+        R.contains(columnId, condition.column_id) :
```
alexcjohnson, line 61:
```suggestion
        R.includes(columnId, condition.column_id) :
```

`contains` is deprecated (other instances too)

src/dash-table/conditional/index.ts
```diff
@@ -29,17 +34,31 @@ export interface IEditableElement {
...
 function ifAstFilter(ast: QuerySyntaxTree, datum: Datum) {
     return ast.isValid && ast.evaluate(datum);
 }
 
+export function ifColumnActive(condition: IActiveElement | undefined, active: boolean, selected: boolean) {
+    if (!condition) {
+        return true;
+    }
+
+    // Apply active if active, apply selected if selected AND not active (active and selected styling don't overlap)
```
alexcjohnson, line 50: It is pretty common to want some commonality - for example selection gets a certain background color, active gets that same background color plus a border. How hard would it be to allow this? Like, first you look for `is_active: true` conditions, then any attribute that's not already set can be set by an `is_selected: true` condition? Not terrible if we can't make that work, but it does seem like it would be more intuitive that way.

src/dash-table/conditional/index.ts
```diff
@@ -68,9 +89,11 @@ export function ifHeaderIndex(condition: IIndexedHeaderElement | undefined, head
     }
 
     const headerCondition = condition.header_index;
-    return typeof headerCondition === 'number' ?
-        headerIndex === headerCondition :
-        headerCondition === 'odd' ? headerIndex % 2 === 1 : headerIndex % 2 === 0;
+    return typeof headerCondition === 'string' ?
+        headerCondition === 'odd' ? headerIndex % 2 === 1 : headerIndex % 2 === 0 :
```
alexcjohnson, line 93:
```suggestion
        headerIndex % 2 === (headerCondition === 'odd' ? 1 : 0) :
```

src/dash-table/derived/edges/data.ts
```diff
@@ -30,17 +34,84 @@ export default memoizeOneFactory((
...
     }
 
+    const edges = baseline.clone();
+
+    const cells = selectedCells.length ?
+        selectedCells :
+        activeCell ? [activeCell] : [];
+
+    const activeStyles = styles.filter(style => style.checksActive());
+    const inactiveStyles = styles.filter(style => !style.checksActive());
```
alexcjohnson, line 69: https://ramdajs.com/docs/#partition

chriddyp:
@wbrgss - Could you take a look as well and make sure that we'll be able to customize some of this stuff in the Dash Enterprise Design Kit?

wbrgss:
@chriddyp this looks good to me w.r.t that project. Using these new `style_data_conditional` props will override the theme, which IMO is what we want. After there's an npm release I'll use that to update the generate the `style_data_conditional` `PropTypes`.

wbrgss:
@Marc-Andre-Rivet I'm noticing what I think is an odd regression in this PR; it could also be an unreleased regression from other changes on `dev`. I'm using an example you provided in the [Community Forum](https://community.plotly.com/t/dash-datatable-style-data-conditional-row-vice/21587/7).

On `dash-table` 4.6.2 this example works fine. With the built version of this PR, after clicking around to rapidly select different cells, I can't select cells anymore and I get this console error:

image

I can reproduce consistently but I'm not sure of the exact steps to get this state. Clicking more slowly seems to work OK. Please let me know if you have trouble reproducing and I can show you.

```py
# -*- coding: utf-8 -*-
import dash

import pandas as pd

url = 'https://github.com/plotly/datasets/raw/master/26k-consumer-complaints.csv'

rawDf = pd.read_csv(url)
df = rawDf.to_dict("rows"),

columns=[{"name": i, "id": i, "type": "text", 'format': {'locale': {'group': '.', 'decimal': ','}}} for i in rawDf.columns]

app = dash.Dash()
app.scripts.config.serve_locally = True

app.layout = DataTable(
    id='datatable',
    data=rawDf[0:10].to_dict("rows"),
    columns=columns,
    style_data_conditional=[
        {
            'if': {
                'filter_query': '{Sub-product} eq "Payday loan"',
            },
            'backgroundColor': 'pink'
        },
        {
            'if': {
                'filter_query': '{Sub-product} eq "Medical"',
            },
            'backgroundColor': 'lightblue'
        },
    ]
)

if __name__ == "__main__":
    app.run_server(port=8053)
```

src/dash-table/conditional/index.ts
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/729/commits/2af76ce8767edb59be885b108ee0dfab513adf54

src/dash-table/conditional/index.ts
Marc-Andre-Rivet: Reworked this more deeply. Instead of `is_active` and `is_selected` using a single `state:'active'|'selected'` nested prop for both. Makes it impossible to make invalid combinations like `active=True, selected=False`. https://github.com/plotly/dash-table/pull/729/commits/43c00fbe4bd71a45c3d24db8fde7cb2881968d17

src/dash-table/derived/edges/data.ts
Marc-Andre-Rivet: With active/selected/default there's now three buckets of styles, partition won't work.

Marc-Andre-Rivet:
@wbrgss The regression is in the PR, thanks for pointing it out, it would have been a pretty nasty bug. Fixed by https://github.com/plotly/dash-table/pull/729/commits/ce95cfc55841d6ab82575c26b898216de62a3811

wbrgss:
@Marc-Andre-Rivet

ce95cfc verified to have fixed the issue I mentioned above (https://github.com/plotly/dash-table/pull/729#issuecomment-613569260), nice!


254fb2c is going to break theming for a certain project. I could use `state:'active'|'selected'` in the wrapper, but then it's difficult to override the theme for users. I think the best solution for me would be to override the new CSS rule in that commit in order to preserve the theming and free up `state:'active'|'selected'` for user overrides. But right now, that project overrides the CSS rule that uses the IE11-incompatible `var(--`. So this might slow down this getting released from my perspective unless I do a maintenance release :slightly_frowning_face:

Marc-Andre-Rivet:
@wbrgss https://github.com/plotly/dash-table/pull/729/commits/345918ab54d26cd51d037f5c474833ccac8e29b1 should allow you to continue using the variables for existing overrides and will allow IE11 to work with the defaults (won't be able to override through the var, but that could never work anyway)

src/dash-table/components/Table/Table.less
```diff
@@ -521,6 +521,7 @@
             --background-color-ellipses: rgb(253, 253, 253);
             --faded-text: rgb(250, 250, 250);
             --faded-text-header: rgb(180, 180, 180);
+            --selected-background: rgba(255, 65, 54, 0.2)
```
wbrgss, line 524: needs a semicolon ;)

src/core/environment/index.ts
```diff
@@ -6,14 +6,14 @@ import { Edge } from 'dash-table/derived/edges/type';
...
 
-const DEFAULT_EDGE: Edge = '1px solid #d3d3d3';
-const ACTIVE_EDGE: Edge = '1px solid var(--accent)';
-
 interface ISearchParams {
     get: (key: string) => string | null;
 }
 
 export default class Environment {
+    private static readonly _supportsCssVariables = Boolean(window.CSS?.supports?.('.some-selector', 'var(--some-var)'));
```
alexcjohnson, line 14: Whoa, the optional chaining operator `?.` was new to me as well 🤯 🤩 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

tests/visual/percy-storybook/Style.percy.tsx
```diff
@@ -374,4 +388,124 @@ storiesOf('DashTable/Style type condition', module)
...
+                border: '1px solid blue'
+            },
+            {
+                if: {
+                    state: 'active'
+                },
+                backgroundColor: 'black',
+                color: 'white',
+                border: '1px solid white'
+            }
```
alexcjohnson, line 500: Can we add one where `active` _partially_ overrides `selected`? Like:
```js
{
  if: {state: 'selected'},
  backgroundColor: 'lightblue',
  border: '1px solid blue'
},
{
  if: {state: 'active'},
  color: 'MediumPurple',
  border: '2px solid black'
}
```
In that case, the active cell would still have a light blue background, right?

src/core/environment/index.ts
Marc-Andre-Rivet: Yup. Using all the new things! 😀

tests/visual/percy-storybook/Style.percy.tsx
Marc-Andre-Rivet: Yes, that's the expected behavior. https://github.com/plotly/dash-table/pull/729/commits/8085950b3ffbc3c618db19df52a314640387a4ce

alexcjohnson:
> @wbrgss The regression is in the PR, thanks for pointing it out, it would have been a pretty nasty bug. Fixed by [ce95cfc](https://github.com/plotly/dash-table/commit/ce95cfc55841d6ab82575c26b898216de62a3811)

> We want the datum, not the prop

Is there a test covering this nasty bug?

Marc-Andre-Rivet:
@alexcjohnson 
> Is there a test covering this nasty bug?

Short answer: no.
Long answer: Improved typing in https://github.com/plotly/dash-table/pull/729/commits/8350c64684358f436a1bc6112db4fa0e631ae978, should be impossible to try and use a row field as an object now.

The buggy code will now trigger a TypeScript build error:
image

rallyemax:
```
private static readonly _supportsCssVariables = Boolean(window.CSS?.supports?.('.some-selector', 'var(--some-var)'));

private static readonly _activeEdge: Edge = Environment._supportsCssVariables ? '1px solid var(--accent)' : '1px solid hotpink';
```
This test returns false in Chromium/Chrome 81.0.4044.129 (Linux, Win10) and Firefox 75.0. The result is that the accent color reverts to a hardcoded hotpink, preventing it from being overridden generally in CSS. (Unless I'm missing something, a general override is needed to prevent the selected cell from having accented borders, since vertically adjacent cells have adjacent borders set to this color as well and those don't have any unique class name that can be targeted with CSS.)

This works and appears to be slightly more idiomatic: 
```
window.CSS.supports('color', 'var(--fake-var)')
``` 
That also works with other valid attributes, like the ones at issue here (`border` and its directional siblings).

In a more general sense, it would be nice if active edge highlighting could be toggled on and off from the Python side. If that can't be implemented, then I would kindly request that the name of the variable be unique to active edge highlighting so that it can be changed or disabled in CSS without affecting other elements (like the sort arrows).

Is this really only to support IE? I think it's worth asking the question of whether it should be done at all. If supporting a deprecated browser requires the kind of gymnastics that even _might_ break functionality in non-deprecated browsers...well, what's the point?

bhofmei:
This PR really messed with my dash apps. I have numerous dash apps built on top of flask and most of the dash apps currently use data table. To keep a consistent theme across all dash apps, I was able to change the selected ground color by just creating a custom CSS and having that CSS used in all the dash apps.

With this update, it now **always** uses the default pink set as element style so I can't override the CSS. To keep the previous styling, I will now need to add this conditional to EVERY table. I currently have 20+ tables and many more coming.

Peter-Pasta:
I am also very interested how this could be implemented in a separate css stylesheet. I found a potential solution in the community forum but it leads to an error in my case: https://community.plotly.com/t/remove-or-change-hotpink-selection-from-datatable/26168/12

Bildschirmfoto 2020-08-18 um 17 10 14
Bildschirmfoto 2020-08-18 um 17 20 03

PR #731, opened 2020-04-07
Issue 384 - Handling of `None` data and columns props

### LINKED ISSUE #384: Validate `data` attribute
Currently, when the `data` attribute is set to `None`, Dash will happily accept as a valid value but this will result in the dreaded "Error loading dependencies" in the app.

This came up over in plotly/dash#616, where it resulted in a bug that was hard to pin down. 

It would be nice if the validation could extend to values passed in by the callback, rather than just on the initial contents of the layout (although I don't know if we've solved this problem for validation of components elsewhere) as this was how the linked example manifested.

Marc-Andre-Rivet:
Closes #384 

Changes table validation/sanitation to handle `data=None` and `columns=None` without throwing an error.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+testpaths = tests/
+addopts = -rsxX -vv
+log_format = %(asctime)s | %(levelname)s | %(name)s:%(lineno)d | %(message)s
+log_cli_level = ERROR
--- a/tests/selenium/test_empty.py
+++ b/tests/selenium/test_empty.py
@@ -1,0 +1,45 @@
+import dash
+from dash.dependencies import Input, Output
+from dash.exceptions import PreventUpdate
+
+from dash_table import DataTable
+from dash_html_components import Button, Div
+
+
+def get_app():
+    app = dash.Dash(__name__)
+
+    columns = [{"name": i, "id": i} for i in ["a", "b"]]
+    data = [dict(a=1, b=2), dict(a=11, b=22)]
+
+    app.layout = Div(
+        [
+            Button(id="clear-table", children=["Clear table"]),
+            DataTable(id="table", columns=columns, data=data),
+        ]
+    )
+
+    @app.callback(
+        [Output("table", "data"), Output("table", "columns")],
+        [Input("clear-table", "n_clicks")],
+    )
+    def clear_table(n_clicks):
+        if n_clicks is None:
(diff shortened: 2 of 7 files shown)


chriddyp:
What about if callbacks update an empty datatable dynamically? 

```python
app.layout = html.Div([
    html.Label('Select a dataset'),
    dcc.Dropdown(options=[...], value=..., id='dropdown'),
    dash_table.DataTable(id='table')
])

@app.callback(Output('table', 'data'), [Input('dropdown', va'lue')])
def ...
```

Marc-Andre-Rivet:
@chriddyp Good point. Will change the internal validation handling instead.

tests/selenium/test_empty.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_empt001_clear_(test):
+    test.start_server(get_app())
+
+    target = test.table("table")
+
+    assert target.is_ready()
+    test.driver.find_element_by_css_selector("#clear-table").click()
+    assert target.is_ready()
+    assert len(test.driver.find_elements_by_css_selector("tr")) == 0
```
Marc-Andre-Rivet, line 45: Once the callback returns, the table should (1) load successfully and get to the ready state, (2) should contain no rows.

TDD'ed locally, can undo redo the fix in the PR if we'd rather have the demonstration.

alexcjohnson: I'm fine with not seeing the test fail on CI - I assume we're all running our own tests with the fix disabled. But it would be nice to have this test check for the proper number of `<tr>` elements before the `clear-table`.

Also, given that you have the `is_ready()` method this may not be helpful, but in wildcards I added a [`wait_for_no_elements` method](https://github.com/plotly/dash/blob/3ec55b5a97fdd34ed8384085df884d86730b1a9c/dash/testing/browser.py#L223)

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+testpaths = tests/
+addopts = -rsxX -vv
+log_format = %(asctime)s | %(levelname)s | %(name)s:%(lineno)d | %(message)s
+log_cli_level = ERROR
```
Marc-Andre-Rivet, line 5: Same as for DCC - accelerates `-k` test runs
https://github.com/plotly/dash-core-components/pull/740

PR #734, opened 2020-04-09
Make R examples within help runnable for CRAN

### LINKED ISSUE #733: Set R examples to run for CRAN checks
As requested during the most recent CRAN submission, we should remove `\dontrun` from all R examples, and instead preface with

```
if(interactive()){
  ...
}
```

rpkyle:
Closes #733.


dash-info.yaml
```diff
@@ -9,70 +9,81 @@ pkg_help_description: >-
...
+                  row_deletable = TRUE
+                )
               )
-            )
 
-            app$run_server()
+              # block=FALSE also required to pass CRAN checks
+              app$run_server(block=FALSE)
+              app$server$stop()
+           }
```
Marc-Andre-Rivet, line 89:
```suggestion
           ) 
```

rpkyle:
@Marc-Andre-Rivet I suspect that these test failures are unrelated to my PR; should we merge in spite of them, or hold off until they're resolved?

PR #751, opened 2020-04-17
Issue 564 - Customizable column filtering logical operator

### LINKED ISSUE #564: [FEATURE] UI for selecting && or || between filters
That is, allow the user to pick whether to AND or to OR filters of multiple columns, or in less technical terminology, whether ALL OF or just ANY OF the column filters is required for a row to be shown.

Marc-Andre-Rivet:
Closes #564 

Expands on the `filter_action` prop to allow choosing the logical operator used between the column filters. The prop accepts the old values `native`, `custom`, `none` but now also allow

```
{
    type: 'native' | 'custom',
    operator: 'and' | 'or'
}
```

For backward compatibility purposes, if the operator is not defined or if the old values are used, the operator defaults to `and`.

This should allow for further extensions in the future, for example, if we add a column filter menu that allows more complex per-column conditions, this could be represented with something like:

```
{
    ...
    ui: 'input' | 'menu'
}
```


src/dash-table/derived/filter/map.ts
```diff
@@ -74,36 +74,42 @@ function updateMap(map: Map<string, SingleColumnSyntaxTree>, column: IColumn, va
...
+) {
     const asts = Array.from(map.values());
-    const globalFilter = getMultiColumnQueryString(asts);
+    const globalFilter = getMultiColumnQueryString(asts, operator);
 
     const rawGlobalFilter = R.map(
         ast => ast.query || '',
         R.filter<SingleColumnSyntaxTree>(ast => Boolean(ast), asts)
-    ).join(' && ');
+    ).join(` ${operator === FilterLogicalOperator.And ? '&&' : '||'} `);
```
alexcjohnson, line 88:
```suggestion
    ).join(operator === FilterLogicalOperator.And ? ' && ' : ' || ');
```
not a big deal, but the string interpolation feels less readable to me in this case - as well as a few chars longer 😅

Marc-Andre-Rivet: 👍

PR #766, opened 2020-04-27
Issue 765 - Sanitize Id

### LINKED ISSUE #765: [BUG] DashTable and Wildcard / Pattern Matching callbacks not working
**Describe your context**

I want to use a datatables 'data' attribute (or any other) in a callback but using the new pattern matching callbacks. 

This results in a seemingly unrelated JS error (parsing problem?)

```
Failed to execute 'addRule' on 'CSSStyleSheet': Failed to parse the rule '#{"action":"test","table":1} .dash-fixed-column tr { height: 30px; }'.
```

-  replace the result of `pip list | grep dash` below
```
dash                      1.11.0      
dash-bootstrap-components 0.9.2       
dash-canvas               0.1.0       
dash-core-components      1.9.1       
dash-cytoscape            0.1.1       
dash-daq                  0.3.3       
dash-extra-components     0.0.1       
dash-html-components      1.0.3       
dash-leaflet              0.0.3       
dash-planner              0.0.1       
dash-renderer             1.4.0       
dash-table                4.6.2    
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: OSX
    - Browser Any

**Describe the bug**

Here is a test setup to reproduce the bug:

```
import dash
import dash_table
import pandas as pd
from dash.dependencies import MATCH, Output, Input, ALL

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Div(id={'table': 1, 'action': 'output'}),
    dash_table.DataTable(
    id={'table': 1, 'action': 'test'},
    data=df.to_dict('records'),
)])

@app.callback(Output({'table': MATCH, 'action': 'output'}, 'children'),
              [Input({'table': MATCH, 'action': 'test'}, 'active_cell')])
def test(xy):
    return xy

if __name__ == '__main__':
    app.run_server(debug=True)

```

**Expected behavior**

- No Failed to execute 'addRule' on 'CSSStyleSheet' Error
- Working callback

**Screenshots**

image

image

Marc-Andre-Rivet:
Closes #765 

Escapes typical problematic characters that can be used for `id` and that need to be escaped for CSS selector.


tests/visual/percy-storybook/Css.percy.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import DataTable from 'dash-table/dash/DataTable';
+
+const setProps = () => { };
+
+storiesOf('DashTable/CSS override', module)
+    .add('css with escaped id', () => (<DataTable
+        setProps={setProps}
+        id={'`~!@#$%^&*()=+ \\|/.,:;\'"`?[]<>{}'}
```
Marc-Andre-Rivet, line 10: Without the fix, the table does not render and throws.

src/core/Stylesheet.ts
```diff
@@ -5,6 +5,8 @@ interface IRule {
     selectorText: string;
 }
 
+export const sanitizeId = (id: string) => id.replace(/([`~\!\@#\$%\^&\*\(\)=\+\\|\/,.'\";:\?><\[\]\{\}\s])/g, '\\$1');
```
alexcjohnson, line 8: There's a little more to it unfortunately... at least leading digits (`id="123"`) and given the complexity of discussions about this I suspect there's more. Maybe just pull in https://github.com/mathiasbynens/CSS.escape unless we already have `CSS.escape` in our polyfills?

Marc-Andre-Rivet: I was trying to deal with the most common offenders, but you're right, might as well just deal with all cases. `CSS.escape` is not included in the polyfill library we use. Since only the table requires it, I'll include it to the table and add tests for 2 typical cases (leading number, leading underscore not followed by a letter)

tests/visual/percy-storybook/Css.percy.tsx
alexcjohnson: leading to the developer screaming "~!@#$%^&*()=+ \\|/.,:;\'"`?[]<>{}"

tests/visual/percy-storybook/Css.percy.tsx
```diff
@@ -0,0 +1,38 @@
...
+        {...props}
+        id={'123'}
+    />))
+    .add('escaped characters', () => (<DataTable
+        {...props}
+        id={'`~!@#$%^&*()=+ \\|/.,:;\'"`?[]<>{}'}
+    />))
+    .add('stringified object as id', () => (<DataTable
+        {...props}
+        id={{ id: 3, group: 'A' }}
```
alexcjohnson, line 37: FWIW you're never going to actually get an object sent to the component by the renderer as an `id` - the renderer stringifies the id first, and does so as sorted JSON, so you'll get `'{"group":"A","id":3}'`

Does `DataTable` separately stringify ids? Because if I just call `CSS.escape` on an object it just gives `"\[object\ Object\]"`

Marc-Andre-Rivet: > Does DataTable separately stringify ids?

Oups.
https://github.com/plotly/dash-table/pull/766/commits/41981cbe86bb2006e7d14d3e363184f9a617154e

PR #785, opened 2020-05-27
Re-render on changes to undefined props

Marc-Andre-Rivet:
For callback chain testing purposes. Also fixes a bug that causes the table to not re-render when a prop goes from being undefined -> defined.

### DIFF
--- a/tests/cypress/tests/unit/table_update_test.ts
+++ b/tests/cypress/tests/unit/table_update_test.ts
@@ -1,0 +1,27 @@
+import shouldComponentUpdate from 'dash-table/components/Table/shouldComponentUpdate';
+
+describe('shouldComponentUpdate', () => {
+    it('should update on undefined -> defined props', () => {
+        assert(shouldComponentUpdate({}, { a: 0 }, {}, {}));
+    });
+
+    it('should update on undefined -> defined state', () => {
+        assert(shouldComponentUpdate({}, {}, {}, { a: 0 }));
+    });
+
+    it('should update on defined -> undefined props', () => {
+        assert(shouldComponentUpdate({ a: 0 }, {}, {}, {}));
+    });
+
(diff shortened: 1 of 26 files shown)


tests/integration/review_app/test_app_df_graph.py
```diff
@@ -152,5 +152,8 @@ def update_graph(rows, selected_rows):
     dash_duo.wait_for_element("#waitfor")
 
     dash_duo.wait_for_element("#{}".format(IDS["table"]))
+    dash_duo.wait_for_element("#pop svg")
+    dash_duo.wait_for_element("#lifeExp svg")
+    dash_duo.wait_for_element("#gdpPercap svg")
```
Marc-Andre-Rivet, line 157: Slight changes in timing / rendering require the test to be more specific and wait for the graph to be ready

Marc-Andre-Rivet:
@alexcjohnson With https://github.com/plotly/dash/pull/1254 merged, the fix is now necessary.

tests/cypress/tests/unit/table_update_test.ts
```diff
@@ -0,0 +1,27 @@
+import shouldComponentUpdate from 'dash-table/components/Table/shouldComponentUpdate';
+
+describe('shouldComponentUpdate', () => {
+    it('should update on undefined -> defined props', () => {
+        assert(shouldComponentUpdate({}, { a: 0 }, {}, {}));
+    });
+
+    it('should update on undefined -> defined state', () => {
+        assert(shouldComponentUpdate({}, {}, {}, { a: 0 }));
```
Marc-Andre-Rivet, line 9: This test fails without the added `...nextProps` https://github.com/plotly/dash-table/pull/785/commits/8849e85932ba6b4f3d739a3a6934fbb852d90115#diff-aa707fa377faf64638e062a12fb52d85R13

PR #787, opened 2020-05-29
Issue 710 - Markdown link click

### LINKED ISSUE #710: Clicking on a markdown link should open the link rather than selecting the cell
Users will use `editable=False` for read-only tables and therefore shouldn't select them. It feels awkward to have a cell become selected when the table otherwise looks and feels read-only. 

In read-only tables, users will be selecting cells to:
- Copy and paste text - where they'll expect the standard "click and drag" text selection to work
- Click on embedded markdown links

Both of these interactions feel awkward with the current cell selection behavior.

We could introduce a new property like "cell_selectable", but I believe that this property should be coerced to `False` if `editable=False`

Marc-Andre-Rivet:
Closes #710 

- Changes the behavior of markdown cells so that clicking on the cell both selects the cell and `click through` to the underlying element
- Changes the default markdown link target to `target=_blank` to open a new tab
- New `cell_selectable` prop (default: True) that turns on/off the ability to select and navigate cells in the table (also overrides value of `active_cell` and `selected_cells` during the sanitation phase)

- [x] Single click on markdown link opens it
- [x] "No selection" table mode

### DIFF
--- a/src/dash-table/components/EdgeFactory.tsx
+++ b/src/dash-table/components/EdgeFactory.tsx
@@ -186,7 +186,7 @@ export default class EdgeFactory {
     }
 
     private memoizedCreateEdges = memoizeOne((
-        active_cell: ICellCoordinates,
+        active_cell: ICellCoordinates | undefined,
         columns: Columns,
         visibleColumns: Columns,
         operations: number,
--- a/src/dash-table/components/Table/props.ts
+++ b/src/dash-table/components/Table/props.ts
@@ -301,6 +301,7 @@ export interface IProps {
     tooltip_conditional: ConditionalTooltip[];
 
     active_cell?: ICellCoordinates;
+    cell_selectable?: boolean;
     column_selectable?: Selection;
     columns?: Columns;
     dropdown?: StaticDropdowns;
@@ -351,35 +352,35 @@ export interface IProps {
 }
 
 interface IDefaultProps {
-    active_cell: ICellCoordinates;
+    cell_selectable: boolean;
     column_selectable: Selection;
+    css: IStylesheetRule[];
     dropdown: StaticDropdowns;
     dropdown_conditional: ConditionalDropdowns;
     dropdown_data: DataDropdowns;
-    css: IStylesheetRule[];
     editable: boolean;
+    end_cell: ICellCoordinates;
     export_columns: ExportColumns;
     export_format: ExportFormat;
     export_headers: ExportHeaders;
     fill_width: boolean;
     filter_query: string;
     filter_action: TableAction;
-    include_headers_on_copy_paste: boolean;
-    merge_duplicate_headers: boolean;
     fixed_columns: Fixed;
     fixed_rows: Fixed;
+    include_headers_on_copy_paste: boolean;
+    merge_duplicate_headers: boolean;
     row_deletable: boolean;
     row_selectable: Selection;
     selected_cells: SelectedCells;
     selected_columns: string[];
-    start_cell: ICellCoordinates;
-    end_cell: ICellCoordinates;
-    selected_rows: Indices;
     selected_row_ids: RowId[];
+    selected_rows: Indices;
     sort_action: TableAction;
     sort_by: SortBy;
     sort_mode: SortMode;
     sort_as_null: SortAsNull;
+    start_cell: ICellCoordinates;
     style_as_list_view: boolean;
     tooltip_data: DataTooltips;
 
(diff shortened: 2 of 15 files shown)


src/dash-table/components/EdgeFactory.tsx
```diff
@@ -186,7 +186,7 @@ export default class EdgeFactory {
     }
 
     private memoizedCreateEdges = memoizeOne((
-        active_cell: ICellCoordinates,
+        active_cell: ICellCoordinates | undefined,
```
Marc-Andre-Rivet, line 189: Typing information was inaccurate.

src/dash-table/components/Table/props.ts
```diff
@@ -301,6 +301,7 @@ export interface IProps {
     tooltip_conditional: ConditionalTooltip[];
 
     active_cell?: ICellCoordinates;
+    cell_selectable?: boolean;
```
Marc-Andre-Rivet, line 304: New optional prop. If false, user can't select cells and `active_cell` and `selected_cells` will always be treated as null/empty.

src/dash-table/dash/Sanitizer.ts
```diff
@@ -99,7 +100,16 @@ export default class Sanitizer {
             headerFormat = ExportHeaders.Ids;
         }
 
+        const active_cell = props.cell_selectable ?
+            props.active_cell :
+            undefined;
+
+        const selected_cells = props.cell_selectable ?
+            props.selected_cells :
+            defaultProps.selected_cells
```
Marc-Andre-Rivet, line 109: Make sure `active_cell` and `selected_cells` have default values if not selectable.

src/dash-table/handlers/cellEvents.ts
```diff
@@ -29,7 +30,14 @@ export const handleClick = (
         return;
     }
 
-    e.preventDefault();
+    const column = visibleColumns[col];
+    if (column.presentation !== Presentation.Markdown) {
+        e.preventDefault();
+    }
```
Marc-Andre-Rivet, line 36: Allow click-through to the most nested target element for Markdown cells.

src/dash-table/handlers/cellEvents.ts
```diff
@@ -29,7 +30,14 @@ export const handleClick = (
...
+
+    if (!cell_selectable) {
+        return;
+    }
```
Marc-Andre-Rivet, line 40: Opt out of selection processing if the table doesn't have selectable cells.

src/dash-table/utils/Markdown.ts
```diff
@@ -1,14 +1,14 @@
 import { Remarkable } from 'remarkable';
 import LazyLoader from 'dash-table/LazyLoader';
 
-export default class MarkdownHighlighter {
+export default class Markdown {
```
Marc-Andre-Rivet, line 4: Renamed as this helper doesn't only cover highlights now.

tests/selenium/test_markdown_link.py
```diff
@@ -0,0 +1,37 @@
...
+
+@pytest.mark.parametrize("cell_selectable", [True, False])
+def test_tmdl001_copy_markdown_to_text(test, cell_selectable):
+    test.start_server(get_app(cell_selectable))
+
+    target = test.table("table")
+
+    assert len(test.driver.window_handles) == 1
+    target.cell(0, "a").get().find_element_by_css_selector("a").click()
+    assert len(test.driver.window_handles) == 2
```
alexcjohnson, line 37: Only thing I might add here is (1) verify that the new window actually went to google, and (2) switch back to the first window and verify that the cell is selected iff cell_selectable.

Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/787/commits/617fa0e121062d76aba8061ad5998d9718f5c99b

FlorianRahe:
Nice feature, which I was looking for, but it is not working. My app crashes with the following error:
`The `dash_table.DataTable` component (version 4.7.0) with the ID "table" received an unexpected keyword argument: `cell_selectable`
Allowed arguments: active_cell,... `

alexcjohnson:
@FlorianRahe this feature has been merged but not yet released. It will be bundled with the next Dash release in a couple of weeks, or you can follow the steps in CONTRIBUTING.md to install in developer mode if you want to try it today.

bcliang:
@Marc-Andre-Rivet How does one override the new default `'linkTarget': '_blank'` to go back to the previous behavior (`'_self'`)?

PR #793, opened 2020-06-11
Issue 777 - Align column headers when using fixed rows 

### LINKED ISSUE #777: Headers aren't aligned with columns with fixed_rows
Related to https://github.com/plotly/dash-table/issues/775 and https://github.com/plotly/dash-table/issues/776

I can reproduce this after resizing my window, however I can't reproduce it on first draw. A user from the community can though:

https://community.plotly.com/t/dash-1-12-datatable-format-issues-when-using-fixed-header/39120

image

### LINKED ISSUE #780: Regression - header doesn't scroll horizontally with fixed_rows (As of dash-table 4.7.0 and dash 1.12)
As of dash-table 4.7.0 and dash 1.12, the header doesn't scroll horizontally *at all* when fixed_rows is true.

I guess the documentation pages are running these versions, since this bug can be seen at https://dash.plotly.com/datatable/height (last table in `Vertical Scroll With Fixed Headers`).

_Originally posted by @noisycomputation in https://github.com/plotly/dash-table/issues/746#issuecomment-628907194_

Marc-Andre-Rivet:
Fixes #777
Fixes #780 

Under certain circumstances, the table fragments were misaligned for scenarios involving `fixed_rows` and `fixed_columns`. Also, some visual tests were already showing that the behavior was incorrect but it was hard to spot unless zooming in. For that reason I believe the tests already appropriately cover the cases for this PR and that no additional tests are required, only more caution.

A few problems were encountered:
1. Certain browsers require both the `tr` and the `th` rows to have their size explictly set when forcing them through styles. This happens for fixed row/columns using the `data` nested field. For example, in Firefox, forcing a `tr` to have a width of `200px` can be overriden by a previous `th` requiring more space. The `th` "wins". The fix for this is to both set the `tr` and `th` cell size for the last row of each group.

2. The implementation was subtly wrong, using the size of the wrong fragment to coerce another. I think some of those is due to historical reasons, and were not reworked correctly when the table was reworked to allow variable row heights in https://github.com/plotly/dash-table/pull/722.

3. Certain browsers, for certain tables will not be able to converge on a single table width, but will instead oscillate between two very close values. As far as I know always distant by `1px`. The "iterations" code for the table now checks for cycles.

---

Also, a nice side-effect of this PR is that tables with fixed columns now behave more consistently vs. unfixed ones. Will need to check the documentation for any mention of incorrect width that will have to be removed.
image
now looks like this fixed and unfixed:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "private::test.python": "python -m unittest tests/unit/format_test.py",
     "private::test.unit": "cypress run --browser chrome --spec 'tests/cypress/tests/unit/**/*'",
     "private::test.standalone": "cypress run --browser chrome --spec 'tests/cypress/tests/standalone/**/*'",
-    "build.watch": "webpack-dev-server --content-base dash_table --mode development --config webpack.dev.config.js",
+    "build.watch": "webpack-dev-server --disable-host-check --content-base dash_table --mode development --config webpack.dev.config.js",
     "build": "run-s private::build:js private::build:py",
     "postbuild": "es-check es5 dash_table/*.js",
     "format": "run-s private::format.*",
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -334,7 +334,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         // If the table was resized, re-run `handleResize` except if the
         // new size is the starting size from the previous iteration
         if (
-            Math.abs(nextWidth - currentWidth) > 0.5 &&
+            Math.abs(nextWidth - currentWidth) > 0.1 &&
             nextWidth !== previousWidth
         ) {
             this.handleResize(true, currentWidth);
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -324,21 +324,22 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         }
 
         if (fixed_columns || fixed_rows) {
-            const r1c0CellWidths = Array.from(
-                r1c0.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
-            ).map(c => c.getBoundingClientRect().width);
-
             const r1c1CellWidths = Array.from(
                 r1c1.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
             ).map(c => c.getBoundingClientRect().width);
 
             Array.from<HTMLElement>(
                 r0c0.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i]));
+            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
 
             Array.from<HTMLElement>(
                 r0c1.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
             ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+
+            Array.from<HTMLElement>(
+                r1c0.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
+            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+
         }
     }
 
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -226,7 +226,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
     forceHandleResize = () => this.handleResize(true);
 
     resizeFragmentCells = (fragment: HTMLElement, forcedCellsWidths: string[] | number[]) => {
-        const lastRowOfCells = fragment.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *');
+        const lastRowOfCells = fragment.querySelectorAll<HTMLElement>('table.cell-table > tbody > tr:last-of-type > *');
         if (!lastRowOfCells.length) {
             return;
         }
@@ -247,8 +247,8 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
             return;
         }
 
-        Array.from(
-            lastTrOfThs.children
+        Array.from<HTMLElement>(
+            lastTrOfThs.children as any
         ).forEach((c, i) => this.setCellWidth(c, forcedCellsWidths[i]));
     }
 
@@ -942,7 +942,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         }
     }
 
-    private setCellWidth(cell: Element, width: string | number) {
+    private setCellWidth(cell: HTMLElement, width: string | number) {
         if (typeof width === 'number') {
             width = `${width}px`;
         }
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -258,11 +258,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         table.style.width = width;
     }
 
-    handleResize = (force: boolean = false, iteration: number = 0) => {
-        if (iteration >= 10) {
-            return;
-        }
-
+    handleResize = (force: boolean = false) => {
         const {
             fixed_columns,
             fixed_rows,
@@ -332,8 +328,11 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
 
         // It's possible that the previous steps resized the r1c1 fragment, if it did,
         // handleResize again until it stabilizes
-        if (tableWidth !== getComputedStyle(r1c1Table).width) {
-            this.handleResize(force, ++iteration);
+        const prevWidth = parseInt(tableWidth, 10);
+        const currentWidth = parseInt(getComputedStyle(r1c1Table).width, 10);
+
+        if (Math.abs(prevWidth - currentWidth) > 0.5) {
+            this.handleResize(true);
         }
     }
 
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -799,7 +799,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
 
         Logger.trace(`ControlledTable fragment scrolled to (left,top)=(${ev.target.scrollLeft},${ev.target.scrollTop})`);
 
-        const margin = parseFloat(ev.target.scrollLeft) + parseFloat(r0c0.style.width);
+        const margin = parseFloat(ev.target.scrollLeft) + (parseFloat(r0c0.style.width) || 0);
 
         r0c1.style.marginLeft = `${-margin}px`;
 
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -325,19 +325,19 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
 
             Array.from<HTMLElement>(
                 r0c0.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i]));
+            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i], !!fixed_columns));
 
             Array.from<HTMLElement>(
                 r0c0.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i]));
+            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i], !!fixed_columns));
 
             Array.from<HTMLElement>(
                 r0c1.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i], !!fixed_columns));
 
             Array.from<HTMLElement>(
                 r0c1.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i], !!fixed_columns));
         }
     }
 
@@ -945,22 +945,24 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         }
     }
 
-    private setCellWidth(cell: HTMLElement, width: number) {
+    private setCellWidth(cell: HTMLElement, width: number, applyDelta: boolean) {
         cell.style.width = `${width}px`;
         cell.style.minWidth = `${width}px`;
         cell.style.maxWidth = `${width}px`;
         cell.style.boxSizing = 'border-box';
 
-        /**
-         * Some browsers handle `th` and `td` size inconsistently.
-         * Checking the size delta and adjusting for it (different handling of padding and borders)
-         * allows the table to make sure all sections are correctly aligned.
-         */
-        const delta = cell.getBoundingClientRect().width - width;
-        if (delta) {
-            cell.style.width = `${width - delta}px`;
-            cell.style.minWidth = `${width - delta}px`;
-            cell.style.maxWidth = `${width - delta}px`;
+        if (applyDelta) {
+            /**
+             * Some browsers handle `th` and `td` size inconsistently.
+             * Checking the size delta and adjusting for it (different handling of padding and borders)
+             * allows the table to make sure all sections are correctly aligned.
+             */
+            const delta = cell.getBoundingClientRect().width - width;
+            if (delta) {
+                cell.style.width = `${width - delta}px`;
+                cell.style.minWidth = `${width - delta}px`;
+                cell.style.maxWidth = `${width - delta}px`;
+            }
         }
     }
 
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -333,7 +333,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
 
             Array.from<HTMLElement>(
                 r0c0.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i]));
+            ).forEach((c, i) => this.setCellWidth(c, r1c0CellWidths[i], true));
 
             Array.from<HTMLElement>(
                 r0c1.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
@@ -341,7 +341,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
 
             Array.from<HTMLElement>(
                 r0c1.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i], true));
         }
     }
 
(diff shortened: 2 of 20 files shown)


src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,68 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
 
         const { r0c0, r0c1, r1c0, r1c1 } = this.refs as { [key: string]: HTMLElement };
 
+        const r0c0Table = r0c0.querySelector('table');
+        const r0c1Table = r0c1.querySelector('table');
+        const r1c0Table = r1c0.querySelector('table');
+        const r1c1Table = r1c1.querySelector('table') as HTMLElement;
```
Marc-Andre-Rivet, line 286: Casting - this is the only fragment that is always present

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,68 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-            const tr2 = trs[index];
+        const currentTableWith = getComputedStyle(r1c1Table).width;
 
-            tr.style.height = `${tr2.clientHeight}px`;
-        });
+        if (!cycle) {
+            this.resizeFragmentTable(r0c0Table, currentTableWith);
+            this.resizeFragmentTable(r0c1Table, currentTableWith);
+            this.resizeFragmentTable(r1c0Table, currentTableWith);
+        }
```
Marc-Andre-Rivet, line 294: If not in a cycle, make sure all fragments are the same size as the "main" one (r1c1)

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,68 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+                r1c1.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
+            ).map(c => c.getBoundingClientRect().width);
 
-            tr.style.height = getComputedStyle(tr2).height;
-        });
+            if (!cycle) {
+                this.resizeFragmentCells(r0c0, widths);
+                this.resizeFragmentCells(r0c1, widths);
+                this.resizeFragmentCells(r1c0, widths);
+            }
```
Marc-Andre-Rivet, line 305: If not in a cycle, try and force all columns to be the same size as the columns from the main fragment. If in a cycle, don't do anything - the last calculated values will be used.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,68 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-                    const r1c0FragmentBounds = r1c0.getBoundingClientRect();
-                    const lastTdBounds = lastVisibleTd.getBoundingClientRect();
-                    currentWidth = lastTdBounds.right - r1c0FragmentBounds.left;
-
-                    const width = `${currentWidth}px`;
-
-                    r0c0.style.width = width;
-                    r1c0.style.width = width;
+                r0c0.style.width = `${lastFixedTdRight}px`;
+                r1c0.style.width = `${lastFixedTdRight}px`;
```
Marc-Andre-Rivet, line 316: Whether in a cycle or not, make sure the width allocated to the fixed columns is sufficient for the fixed columns.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,68 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+                    const firstVisibleTd = r1c1.querySelector(`tr:first-of-type > *:nth-of-type(${fixed_columns + 1})`);
+                    if (firstVisibleTd) {
+                        const r1c1FragmentBounds = r1c1.getBoundingClientRect();
+                        const firstTdBounds = firstVisibleTd.getBoundingClientRect();
+
+                        const width = firstTdBounds.left - r1c1FragmentBounds.left;
+                        r0c1.style.marginLeft = `-${width}px`;
+                        r0c1.style.marginRight = `${width}px`;
+                        r1c1.style.marginLeft = `-${width}px`;
+                        r1c1.style.marginRight = `${width}px`;
```
Marc-Andre-Rivet, line 329: If not fixed, re-evaluate the padding - this caused the main fragment to resize and is what forces the iteration process.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,68 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-                r0c1.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
-
-            Array.from<HTMLElement>(
-                r0c1.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+            // If the table was resized and isn't in a cycle, re-run `handleResize`.
+            // If the final size is the same as the starting size from the previous iteration, do not
+            // resize the main table, instead just use as is, otherwise it will oscillate.
+            if (nextWidth !== currentWidth) {
```
Marc-Andre-Rivet, line 342: The width can have a decimal part. Need to change this to use an epsilon instead of equality.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,68 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+                this.handleResize(true, currentWidth, nextWidth === previousWidth);
```
Marc-Andre-Rivet, line 343: Same here.

package.json
```diff
@@ -33,7 +33,7 @@
     "private::test.python": "python -m unittest tests/unit/format_test.py",
     "private::test.unit": "cypress run --browser chrome --spec 'tests/cypress/tests/unit/**/*'",
     "private::test.standalone": "cypress run --browser chrome --spec 'tests/cypress/tests/standalone/**/*'",
-    "build.watch": "webpack-dev-server --content-base dash_table --mode development --config webpack.dev.config.js",
+    "build.watch": "webpack-dev-server --disable-host-check --content-base dash_table --mode development --config webpack.dev.config.js",
```
Marc-Andre-Rivet, line 36: In Browserstack, when running on iOS devices, `localhost` is redirected to `bs-local` which webpack-dev-server doesn't like, this flag, as its name indicates, makes it not care

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,66 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-                const firstVisibleTd = r1c1.querySelector(`tr:first-of-type > *:nth-of-type(${fixed_columns + 1})`);
-                if (firstVisibleTd) {
-                    const r1c1FragmentBounds = r1c1.getBoundingClientRect();
-                    const firstTdBounds = firstVisibleTd.getBoundingClientRect();
-
-                    const width = firstTdBounds.left - r1c1FragmentBounds.left;
-                    r0c1.style.marginLeft = `-${width}px`;
-                    r0c1.style.marginRight = `${width}px`;
-                    r1c1.style.marginLeft = `-${width}px`;
-                    r1c1.style.marginRight = `${width}px`;
```
Marc-Andre-Rivet, line 307: `marginRight` caused the table to be smaller than expected with fixed columns. Removing it for both `r0c1` and `r1c1` fragments ensures the table is the expected size

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -249,95 +280,66 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-                r0c1.querySelectorAll('table.cell-table > tbody > tr:first-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
-
-            Array.from<HTMLElement>(
-                r0c1.querySelectorAll('table.cell-table > tbody > tr:last-of-type > *')
-            ).forEach((c, i) => this.setCellWidth(c, r1c1CellWidths[i]));
+            // If the table was resized and isn't in a cycle, re-run `handleResize`.
+            // If the final size is the same as the starting size from the previous iteration, do not
+            // resize the main table, instead just use as is, otherwise it will oscillate.
+            if (nextWidth !== currentWidth) {
```
Marc-Andre-Rivet, line 340: Tested these for scenarios involving decimal places and is seems to behave fine on all important targets, decimals or not.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -150,7 +147,33 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+            style_cell,
+            style_cell_conditional,
+            style_data,
+            style_data_conditional,
+            style_filter,
+            style_filter_conditional,
+            style_header,
+            style_header_conditional,
+            style_table
+        );
```
Marc-Andre-Rivet, line 175: Only resize through React pipeline if a style prop has changed. This behaves similarly to the 🔪 `forcedResizedOnly` state prop, but also allows style changes to trigger recalculation. For example `display:none;`, changes in column widths, etc.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -225,119 +248,183 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-        Array.from(r1c0.querySelectorAll('tr > th:first-of-type')).forEach((th, index) => {
-            const tr2 = trths[index].parentElement as HTMLElement;
-            const tr = th.parentElement as HTMLElement;
+        this.resetFragmentCells(r0c0);
+        this.resetFragmentCells(r0c1);
+        this.resetFragmentCells(r1c0);
 
-            tr.style.height = getComputedStyle(tr2).height;
-        });
+        this.handleResize();
```
Marc-Andre-Rivet, line 344: If executing this, the styles have changed, the above clears up all forced width before handling the resize, allowing for the table to resize itself correctly.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -225,119 +248,183 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-            const lastVisibleTd = r1c0.querySelector(`tr:first-of-type > *:nth-of-type(${fixed_columns})`);
+        const currentTableWidth = getComputedStyle(r1c1Table).width;
 
-            let it = 0;
-            let currentWidth = r1c0.getBoundingClientRect().width;
-            let lastWidth = currentWidth;
+        // Not pixels -> not displayed
+        if (!/\d+px/.test(currentTableWidth)) {
+            return;
+        }
```
Marc-Andre-Rivet, line 362: If not `px`, the table is not rendered, don't resize it. Lucky https://github.com/plotly/plotly.js/pull/4925 happened at the same time, otherwise it would have gone unoticed.

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
+import dash
+from dash.dependencies import Input, Output
+from dash.testing import wait
+from dash_table import DataTable
+from dash_html_components import Button, Div
+
+
+def cells_are_same_width(target, table):
+    wait.until(lambda: abs(target.size["width"] - table.size["width"]) <= 1, 3)
+
```
Marc-Andre-Rivet, line 10: Helper function that checks if two tables that are expected to have the same size / cell sizes are identical (or nearly identical) in size

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
...
+        dict(fixed_columns=dict(headers=True)),
+        dict(fixed_rows=dict(headers=True)),
+        dict(fixed_columns=dict(headers=True), fixed_rows=dict(headers=True)),
+        dict(fixed_columns=dict(headers=True, data=1)),
+        dict(fixed_rows=dict(headers=True, data=1)),
+        dict(
+            fixed_columns=dict(headers=True, data=1),
+            fixed_rows=dict(headers=True, data=1),
+        ),
+    ]
```
Marc-Andre-Rivet, line 91: Create all combinations of fixed columns/rows tables and hide them.

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
...
+                    width=500, minWidth=500, maxWidth=500, paddingBottom=10
+                )
+            ),
+            DataTable(
+                **base_props,
+                id="table750",
+                style_table=dict(
+                    width=750, minWidth=750, maxWidth=750, paddingBottom=10
+                )
+            ),
```
Marc-Andre-Rivet, line 128: These tables are used to make sure the test tables are behaving as expected when updated through the callback below

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
...
+                    display="none",
+                    paddingBottom=10,
+                )
+                for i in variations_range
+            ]
+        elif n_clicks:
+            return [
+                dict(width=350, minWidth=350, maxWidth=350, paddingBottom=10)
+                for i in variations_range
+            ]
```
Marc-Andre-Rivet, line 164: Change the table style to make them visible / hidden and/or resized

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
...
+    tables = [DataTable(**variation) for i, variation in enumerate(variations)]
+
+    app = dash.Dash(__name__)
+    app.layout = Div(tables)
+
+    test.start_server(app)
+
+    target = test.driver.find_element_by_css_selector("#table0")
+
+    for i in range(1, len(variations)):
```
Marc-Andre-Rivet, line 268: Test various configurations of fixed columns/rows and merge_duplicate_headers to make sure they all render the same with % based column widths.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -225,119 +248,183 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+        trOfThs.forEach(tr => {
+            const ths = Array.from<HTMLElement>(tr?.children as any);
+
+            if (!ths) {
+                return;
+            }
+
+            ths.forEach(this.clearCellWidth);
+        });
+    }
```
Marc-Andre-Rivet, line 282: Force all `th` rows with all cells to match the expected width.

src/dash-table/components/ControlledTable/index.tsx
alexcjohnson: OIC, `'100.1px'` will in fact pass this test as it's not a whole string match. Worried me at first but it should work.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -225,119 +248,183 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+        trOfThs.forEach(tr => {
+            const ths = Array.from<HTMLElement>(tr?.children as any);
+
+            if (!ths) {
+                return;
+            }
+            if (ths.length === widths.length) {
+                ths.forEach((c, i) => this.setCellWidth(c, widths[i]));
+            } else {
+                ths.forEach(c => this.setCellWidth(c, 0));
```
alexcjohnson, line 309: When do we hit this?

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -225,119 +248,183 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
+        if (r1c0Table) { r1c0Table.style.width = ''; }
 
-        // Adjust fixed columns headers to header's height
-        let trths = r1c1.querySelectorAll('tr > th:first-of-type');
-        Array.from(r1c0.querySelectorAll('tr > th:first-of-type')).forEach((th, index) => {
-            const tr2 = trths[index].parentElement as HTMLElement;
-            const tr = th.parentElement as HTMLElement;
+        this.resetFragmentCells(r0c0);
+        this.resetFragmentCells(r0c1);
+        this.resetFragmentCells(r1c0);
```
alexcjohnson, line 342: Minor - skip reset for a table that doesn't exist? Also seems like this could be 🌴
```
[r0c0, r0c1, r1c0].forEach(rc=> {
    rcTable=...;
    if (rcTable) { rcTable.style.width = ''; this.resetFragmentCells(rc); }
});
```

tests/selenium/test_sizing.py
alexcjohnson: This routine is testing the last row of cells; while we're here would it be worthwhile checking that all the rows of a single table are the same, ie rows aren't ragged? That's certainly been a problem in the past...

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
...
+    variations = []
+    i = 0
+    for style in styles:
+        for fix in fixes:
+            variations.append({**style, **fix, **base_props, "id": "table{}".format(i)})
+            i = i + 1
+
+    variations_range = range(0, len(variations))
+
+    tables = [DataTable(**variation) for i, variation in enumerate(variations)]
```
alexcjohnson, line 102:
```suggestion
    tables = [DataTable(**variation) for variation in variations]
```

tests/selenium/test_sizing.py
alexcjohnson: Beautiful test :)
Looks like you started to collect all these `style_table` dicts up above in `styles` then changed tacks and put them inline down here... could probably be simplified a little, also we don't really need n clones, could just be `return [dict(...)] * variations_range`... but 🤷 not important for a test, the functionality is great!

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
...
+                    "A-----------------VERY LONG",
+                    "B-----------------VERY LONG",
+                ],
+            },
+            {
+                "id": "c",
+                "name": [
+                    "A-----------------VERY LONG",
+                    "B-----------------VERY LONG",
+                    "C-----------------VERY LONG",
```
alexcjohnson, line 222: Would it make sense to have the last row of names all be different lengths, so we know it's not just the content match giving rise to the width match?

tests/selenium/test_sizing.py
```diff
@@ -0,0 +1,270 @@
...
+        ],
+        data=[
+            {"_": 0, "a": 85, "b": 601, "c": 891},
+            {"_": 0, "a": 967, "b": 189, "c": 514},
+            {"_": 0, "a": 398, "b": 262, "c": 743},
+            {"_": 0, "a": 89, "b": 560, "c": 582},
+            {"_": 0, "a": 809, "b": 591, "c": 511},
+        ],
+        style_table=dict(width=500, minWidth=500, maxWidth=500, paddingBottom=10),
+        style_cell=dict(width="20%", minWidth="20%", maxWidth="20%"),
```
alexcjohnson, line 234: I only see 4 columns - is it intentional to have <100% total width?

tests/selenium/test_sizing.py
Marc-Andre-Rivet: Jagged rows happen when we transition from one fragment to another of the DataTable. Checking the cells of any row of a fragment against the cells of any row of the `.cell-1-1` fragment should be sufficient if the table doesn't have a fixed portion.

What I realize from this comment is that for `szng002` I should also be testing the `target` against itself.

src/dash-table/components/ControlledTable/index.tsx
Marc-Andre-Rivet: When using `merge_duplicate_headers`

tests/selenium/test_sizing.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/793/commits/575aa2c4bc838474bcfa251c8596897be6c76d17#diff-cd5fbde6c058807ca1b16f17f7daa4d8R213

tests/selenium/test_sizing.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/793/commits/575aa2c4bc838474bcfa251c8596897be6c76d17#diff-cd5fbde6c058807ca1b16f17f7daa4d8R199

tests/selenium/test_sizing.py
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/793/commits/4dcc753d2ba65cec1dd14e51cc3e59b195e79f70 and https://github.com/plotly/dash-table/pull/793/commits/575aa2c4bc838474bcfa251c8596897be6c76d17#diff-cd5fbde6c058807ca1b16f17f7daa4d8R110

src/dash-table/components/ControlledTable/index.tsx
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/793/commits/575aa2c4bc838474bcfa251c8596897be6c76d17#diff-a6637ea69a08d602c311a574b4e4979aR340

tests/selenium/test_sizing.py
Marc-Andre-Rivet: The test did change a lot between this version and the first implementation. Cleaned it up. https://github.com/plotly/dash-table/pull/793/commits/575aa2c4bc838474bcfa251c8596897be6c76d17

Marc-Andre-Rivet:
@alexcjohnson Addressed all previous comments. Also added https://github.com/plotly/dash-table/pull/793/commits/b02a0226a00d2083caf8ec6e1e5fd8bc698ff927 to fix #780 and lock behavior with new tests.

PR #798, opened 2020-06-19
[WIP] Issue 797 - Misaligned rows after callback

### LINKED ISSUE #797: Misaligned table rows with `fixed_row` / `fixed_column` after callback
First mentioned here: https://community.plotly.com/t/dash-1-12-datatable-format-issues-when-using-fixed-header/39120/9. The issue still exists after #793 

I think the issue is the memoization / condition preventing resizing is too aggressive in preventing recalculations, not taking into account all the relevant props https://github.com/plotly/dash-table/blob/dev/src/dash-table/components/ControlledTable/index.tsx#L166

Marc-Andre-Rivet:
Fixes #797

### DIFF
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -150,7 +150,14 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         this.updateStylesheet();
         this.updateUiViewport();
 
-        this.handleResizeIf(...R.values(this.props));
+        const {
+            fixed_columns,
+            fixed_rows
+        } = this.props;
+
+        if (fixed_columns || fixed_rows) {
+            this.handleResizeIf(...R.values(this.props));
+        }
 
         this.handleDropdown();
         this.adjustTooltipPosition();
--- a/src/dash-table/components/ControlledTable/index.tsx
+++ b/src/dash-table/components/ControlledTable/index.tsx
@@ -150,31 +150,7 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
         this.updateStylesheet();
         this.updateUiViewport();
 
-        const {
-            style_as_list_view,
-            style_cell,
-            style_cell_conditional,
-            style_data,
-            style_data_conditional,
-            style_filter,
-            style_filter_conditional,
-            style_header,
-            style_header_conditional,
-            style_table
-        } = this.props;
-
-        this.handleResizeIf(
-            style_as_list_view,
-            style_cell,
-            style_cell_conditional,
-            style_data,
-            style_data_conditional,
-            style_filter,
-            style_filter_conditional,
-            style_header,
-            style_header_conditional,
-            style_table
-        );
+        this.handleResizeIf(...R.values(this.props));
 
         this.handleDropdown();
         this.adjustTooltipPosition();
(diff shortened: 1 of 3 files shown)


tests/selenium/test_sizing.py
```diff
@@ -248,3 +250,78 @@ def test_szng002_percentages_result_in_same_widths(test):
...
+                    "a": "SOME VERY LONG VALUE 2",
+                    "b": "SOME VERY LONG VALUE 3",
+                    "c": "SOME VERY LONG VALUE 4",
+                },
+                {"_": 0, "a": 89, "b": 560, "c": 582},
+                {"_": 0, "a": 809, "b": 591, "c": 511},
+            ]
+        ),
+        dict(merge_duplicate_headers=True),
+        # dict(fixed_rows=dict(headers=True)),
```
Marc-Andre-Rivet, line 294: Commenting out certain, more redundant, scenarios. With the above, this reduces the number of tests combinations from 126 to 40, which is more palatable in terms of CI time.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -151,30 +151,13 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
...
-            style_data,
-            style_data_conditional,
-            style_filter,
-            style_filter_conditional,
-            style_header,
-            style_header_conditional,
-            style_table
-        );
+        if (fixed_columns || fixed_rows) {
+            this.handleResizeIf(...R.values(this.props));
```
Marc-Andre-Rivet, line 159: Don't attempt resize if nothing is fixed. Otherwise, resize on any prop change. Most props can require a recalculation due to interactions with other props. This will have a performance impact as resizing will happen more often, but the bulk of the cost (relayout) would be happening anyway.

Better to be right slightly slower than to be wrong slightly faster..

tests/selenium/test_sizing.py
```diff
@@ -174,49 +215,10 @@ def update_styles(n_clicks):
...
-            {"_": 0, "a": 85, "b": 601, "c": 891},
-            {"_": 0, "a": 967, "b": 189, "c": 514},
-            {"_": 0, "a": 398, "b": 262, "c": 743},
-            {"_": 0, "a": 89, "b": 560, "c": 582},
-            {"_": 0, "a": 809, "b": 591, "c": 511},
-        ],
-        style_table=dict(width=500, minWidth=500, maxWidth=500, paddingBottom=10),
-        style_cell=dict(width="25%", minWidth="25%", maxWidth="25%"),
-    )
+    global base_props
```
alexcjohnson, line 218: you aren't modifying `base_props` right? Then you don't need this declaration.

tests/selenium/test_sizing.py
```diff
@@ -248,3 +250,78 @@ def test_szng002_percentages_result_in_same_widths(test):
...
+
+    props = {**base_props, **fixed_columns, **fixed_rows, **merge_duplicate_headers}
+
+    table = DataTable(**props, id="table")
+
+    app = dash.Dash(__name__)
+    app.layout = Div([Button(id="btn", children=["Update"]), table])
+
+    @app.callback(
+        [Output("table", key) for key in callback_props.keys()],
```
alexcjohnson, line 314: clever 🎉

tests/selenium/test_sizing.py
Marc-Andre-Rivet: Oh, nice. I thought it was required just to use it :)

Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/798/commits/c93050c95940ba127f784c9a731a4c8ed7ab44e6

PR #806, opened 2020-07-20
Issue 803 - Fix scroll alignment on scroll/hover, edit, navigate

### LINKED ISSUE #803: Misaligned table rows with `fixed_row` / `fixed_column` after mousing over table
This looks like the same error as #777 only it doesn't appear until I mouse back over the table after scrolling.
Peek 2020-07-14 10-10

Code:

```python
import dash
import dash_table
import pandas as pd

app = dash.Dash(__name__)

    columns=[{"name": i, "id": i} for i in list(df.columns) * 3],
    data=df.to_dict('records'),
    fixed_rows={'headers': True},
    style_cell={'minWidth': '180px', 'width': '180px', 'maxWidth': '180px'},
)

if __name__ == '__main__':
    app.run_server(debug=True, port=8022)
```

Marc-Andre-Rivet:
Fixes #803 

When testing this out, using the "two fingers" scroll does not cause misalignment on hover but using the scrollbar + hovering afterwards does. Editing / navigating also triggers the misalignment.

### DIFF
--- a/tests/selenium/test_scrolling.py
+++ b/tests/selenium/test_scrolling.py
@@ -4,8 +4,39 @@ from dash_table import DataTable
 
 import pandas as pd
 import pytest
+from selenium.webdriver.common.keys import Keys
 
 df = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/solar.csv")
+
+base_props = dict(
+    id="table",
+    columns=[{"name": i, "id": i} for i in df.columns],
+    row_selectable="single",
+    row_deletable=True,
+    data=df.to_dict("records"),
+    editable=True,
+    fixed_rows={"headers": True, "data": 1},
+    style_cell=dict(width=150),
+    style_table=dict(width=500),
+)
+
+
+def get_margin(test):
+    return test.driver.execute_script(
+        "return parseFloat(getComputedStyle(document.querySelector('#table .cell-0-1')).marginLeft);"
+    )
+
+
+def get_scroll(test):
+    return test.driver.execute_script(
+        "return document.querySelector('#table .row-1').scrollLeft;"
+    )
+
+
+def scroll_by(test, value):
+    return test.driver.execute_script(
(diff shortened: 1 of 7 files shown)


tests/selenium/test_scrolling.py
```diff
@@ -4,9 +4,40 @@
...
+
+
+def get_scroll(test):
+    return test.driver.execute_script(
+        "return document.querySelector('#table .row-1').scrollLeft;"
+    )
+
+
+def scroll_by(test, value):
+    return test.driver.execute_script(
```
alexcjohnson, line 37:
```suggestion
    test.driver.execute_script(
```

tests/selenium/test_scrolling.py
```diff
@@ -48,32 +68,64 @@ def test_scrol001_fixed_alignment(test, fixed_rows, fixed_columns, ops):
...
+        "return parseFloat(getComputedStyle(document.querySelector('#table .cell-0-0')).width) || 0;"
     )
 
+    scroll_by(test, 200)
+
+    # alignment is ok after editing a cell
+    target.cell(0, 3).click()
+    test.send_keys("abc" + Keys.ENTER)
+
+    wait.until(lambda: -get_margin(test) == fixed_width + get_scroll(test), 3)
```
alexcjohnson, line 123: This is presumably the same scroll position as the table had prior to the `send_keys` above, right? Is it necessarily the case that the table has finished rerendering after the `ENTER` before we execute this? Or do we need to look for something else that ensures rerendering is done? Same question re: the matching test below.

Marc-Andre-Rivet: I could add a wait for the `(0,4)` cell to be selected before checking the position to make 100% certain it re-rendered

Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/806/commits/315218f8bda067af01aae815fa8d1f0d7b588311

PR #808, opened 2020-07-22
Issue 807 - Customizable Markdown options

### LINKED ISSUE #807: Table markdown link opening new tab is problematic for certain workflows
Changed in https://github.com/plotly/dash-table/pull/787 this has come up as a workflow issue in the community forum https://community.plotly.com/t/dash-datatable-with-markdown-links-behaviour-since-dash-v1-13-0/41565 where application developers would like to be able to revert back to the old behavior (opening the link in the current tab)

Marc-Andre-Rivet:
Fixes #807 

Allows configuring markdown options at the table-level and will be compatible with future extensions if/when markdown support is increased.

Use the following value to open links in the current tab and get the behaviour as before #787 
```
markdown_options={"linkTarget": "_self"}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- [#808](https://github.com/plotly/dash-table/pull/808)Fix a regression introduced with [#787](https://github.com/plotly/dash-table/pull/787) making it impossible to open markdown links in the current tab.
+    - Adds a new `markdown_options` property that supports:
+        - `linkTarget` nested prop with values `_blank`, `_parent`, `_self`, `_top` (default: `_blank`)
+
 ### Fixed
 - [#806](https://github.com/plotly/dash-table/pull/806) Fix a bug where fixed rows a misaligned after navigating or editing cells [#803](https://github.com/plotly/dash-table/issues/803)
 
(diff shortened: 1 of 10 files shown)


src/dash-table/dash/DataTable.js
```diff
@@ -423,6 +427,15 @@ export const propTypes = {
         separate_4digits: PropTypes.bool
     }),
 
+    markdown_options: PropTypes.exact({
+        linkTarget: PropTypes.oneOf([
+            '_blank',
+            '_parent',
+            '_self',
+            '_top'
```
alexcjohnson, line 435: Can we just make this a string, to support named windows/frames too, and document the special values in the prop description?

Marc-Andre-Rivet: Sure. Even though it's redundant, I think I'll define the type as :
```
PropTypes.oneOfType([
            Proptypes.string,
            PropTypes.oneOf([
                '_blank',
                '_parent',
                '_self',
                '_top'
            ])
        ])
``` 
so the special values are documented eplicitly as part of the API.

chriddyp:
maybe link_target to keep our property naming consistent?

Marc-Andre-Rivet:
@chriddyp Will do. Requires mapping the API props to internal props for the Markdown library but no problem otherwise.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- [#808](https://github.com/plotly/dash-table/pull/808)Fix a regression introduced with [#787](https://github.com/plotly/dash-table/pull/787) making it impossible to open markdown links in the current tab.
+    - Adds a new `markdown_options` property that supports:
+        - `linkTarget` nested prop with values `_blank`, `_parent`, `_self`, `_top` (default: `_blank`)
```
alexcjohnson, line 9:
```suggestion
        - `link_target` nested prop with values `_blank`, `_parent`, `_self`, `_top` or an arbitrary string (default: `_blank`)
```

PR #814, opened 2020-08-13
Update table dependencies

Marc-Andre-Rivet:
Just some maintenance work -- lest we fall behind and it becomes harder to maintain it up to date.


.config/webpack/base.js
```diff
@@ -46,6 +46,14 @@ module.exports = (options = {}) => {
...
                     loader: 'raw-loader'
                 },
+                {
+                    test: /\.ts(x?)$/,
+                    include: /node_modules\/(highlight[.]js\/)/,
+                    use: [
+                        { loader: 'babel-loader', options: babel },
+                        { loader: 'ts-loader', options: ts },
+                    ]
+                },
```
Marc-Andre-Rivet, line 56: highlight.js is now released in es6 - we need to transpile further

.config/webpack/base.js
```diff
@@ -55,6 +63,13 @@ module.exports = (options = {}) => {
                         { loader: 'webpack-preprocessor', options: JSON.stringify(preprocessor) }
                     ]
                 },
+                {
+                    test: /\.js$/,
+                    include: /node_modules\/(highlight[.]js\/)/,
+                    use: [
+                        { loader: 'babel-loader', options: babel }
+                    ]
+                },
```
Marc-Andre-Rivet, line 72: highlight.js is now released in es6 - we need to transpile further

.storybook/main.js
```diff
@@ -0,0 +1,25 @@
...
+    webpackFinal: async (config, { configType }) => {
+        // jerry rig everything
+        config.resolve.alias.core = path.resolve(__dirname, './../src/core'),
+        config.resolve.alias['dash-table'] = path.resolve(__dirname, './../src/dash-table')
+
+        config.module = baseConfig.module;
+
+        return config;
+    }
+};
```
Marc-Andre-Rivet, line 25: Storybook 6 brings totally new configuration files and opinions. Making it work with our own needs..

tests/cypress/tests/standalone/markdown_test.ts
```diff
@@ -140,7 +140,7 @@ describe('markdown cells', () => {
         it('loads highlight.js and does not attach hljs to window', () => {
             cy.visit(`http://localhost:8080?mode=${AppMode.Markdown}`);
             // wait for highlight.js to highlight code
-            DashTable.getCellById(0, 'markdown-code-blocks').within(() => cy.get('code.language-python span.hljs-title'));
+            DashTable.getCellById(0, 'markdown-code-blocks').within(() => cy.get('code.language-python'));
```
Marc-Andre-Rivet, line 143: highlight.js 10 does not wrap the code in a span anymore, updating the test

.storybook/main.js
rpkyle: @Marc-Andre-Rivet For any package using Storybook >=6, I imagine we now have to add

```javascript
config.resolve.alias['dash-table'] = path.resolve(__dirname, './../src/dash-table')
```

but specific to the path/package name?

Marc-Andre-Rivet: Yes. But the table is the only project using storybook.

PR #815, opened 2020-08-14
[WIP] Smaller entrypoint bundle for async components

Marc-Andre-Rivet:
propTypes, defaultProps and persistenceTransforms represent the majority of the entrypoint bundle but are not truly needed to load the component. This attempts to reduce the entrypoint bundle while not affecting the Dash functionalities, component generation and props validation.

Reduces the entrypoint bundle size from 39kB to 13kB for this package.


package.json
```diff
@@ -21,7 +21,7 @@
     "private::build:js-test": "run-s \"private::build -- --mode development --config webpack.test.config.js\"",
     "private::build:js-test-standalone": "run-s \"private::build -- --mode development --config webpack.test.standalone.config.js\"",
     "private::build:js-test-watch": "run-s \"private::build -- --mode development --config webpack.test.config.js --watch\"",
-    "private::build:py": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json --r-prefix 'dash' --r-suggests 'dash'",
+    "private::build:py": "dash-generate-components src/dash-table/dash/fragments/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/fragments/DataTable.js dash_table -p package-info.json --r-prefix 'dash' --r-suggests 'dash'",
```
Marc-Andre-Rivet, line 24: Generate the docs against the "real" component instead of the wrapper - as long as the names are identical, should be fine

src/dash-table/dash/DataTable.js
```diff
@@ -19,1362 +12,22 @@ import LazyLoader from 'dash-table/LazyLoader';
...
-    persistence: PropTypes.oneOfType([
-        PropTypes.bool,
-        PropTypes.string,
-        PropTypes.number
-    ]),
+const DataTable = props => (
+    <Suspense fallback={null}>
+        <RealDataTable {...props} />
+    </Suspense>
+);
```
Marc-Andre-Rivet, line 19: Make it a function, because there's nothing happening here to justify a class

src/dash-table/dash/DataTable.js
```diff
@@ -19,1362 +12,22 @@ import LazyLoader from 'dash-table/LazyLoader';
...
-        name: {
-            extract: propValue => R.pluck('name', propValue),
-            apply: (storedValue, propValue) =>
-                R.zipWith(R.assoc('name'), storedValue, propValue)
-        }
-    }
-};
+        DataTable.defaultProps = realDataTable.defaultProps;
+        DataTable.propTypes = realDataTable.propTypes;
+        DataTable.persistenceTransforms = realDataTable.persistenceTransforms;
```
Marc-Andre-Rivet, line 29: When the real component is loaded, assign the props that matter - this step could be done in the decorator to ensure consistency for components.

src/dash-table/dash/DataTable.js
```diff
@@ -19,1362 +12,22 @@ import LazyLoader from 'dash-table/LazyLoader';
...
-     * Where persisted user changes will be stored:
-     * memory: only kept in memory, reset on page refresh.
-     * local: window.localStorage, data is kept after the browser quit.
-     * session: window.sessionStorage, data is cleared once the browser quit.
-     */
-    persistence_type: PropTypes.oneOf(['local', 'session', 'memory'])
-};
+const RealDataTable = asyncDecorator(DataTable, () =>
+    LazyLoader.table().then(res => {
+        const realDataTable = res.default;
```
Marc-Andre-Rivet, line 25: `import(...)` loads like an oldschool `require` -> need to get the default export on the `default` prop

src/dash-table/dash/DataTable.js
alexcjohnson: @Marc-Andre-Rivet this is certainly an interesting idea (and if we go ahead with it I agree doing it at the decorator level would be preferable)... but how do we know the renderer waits for this code before trying to [validate props](https://github.com/plotly/dash/blob/58eb07bf4f2f9d64054fdc65f9a1a17cb6fc355e/dash-renderer/src/TreeContainer.js#L41)?

Marc-Andre-Rivet: Good question. Made me realize this is not that generalizable as-is. This works for the table because of the derived props. Once the real component is loaded it is guaranteed to call `setProps` which will force the renderer to re-renderer the associated parent TreeContainer, making this work for the general case is possible but more involved and would involve both a decorator update and a tweak to the TreeContainer.

Marc-Andre-Rivet: One possible generalization: https://github.com/plotly/dash/pull/1372/files

PR #819, opened 2020-08-20
Issue 660 - Current page field is too small for large page number

### LINKED ISSUE #660: Current page number field not large enough for large tables
The changes implemented in relation to https://github.com/plotly/dash-table/issues/580 are great for large tables. However, the field displaying the current page number is not large enough for and does not adapt to large page numbers as in the example below:

image

Marc-Andre-Rivet:
Closes #660

### DIFF
--- a/tests/visual/percy-storybook/Style.percy.tsx
+++ b/tests/visual/percy-storybook/Style.percy.tsx
@@ -384,10 +384,29 @@ storiesOf('DashTable/Style type condition', module)
         style_data_conditional={[{
             if: { column_editable: true }, background_color: 'MediumPurple'
         }]}
-        pagination_settings={{
-            current_page: 0,
-            page_size: 10
+        page_current={0}
+        page_size={10}
+    />))
+    .add('large current page', () => (<DataTable
+        id='paging-large-current-page'
+        data={mock.data.slice(0, 10)}
+        columns={mock.columns.map((col: any) => R.merge(col, {
+            name: col.name,
+            deletable: true
+        }))}
+        style_table={{
+            width: '100%'
         }}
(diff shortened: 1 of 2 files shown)


tests/visual/percy-storybook/Style.percy.tsx
```diff
@@ -384,10 +384,29 @@ storiesOf('DashTable/Style type condition', module)
...
+        row_deletable={true}
+        row_selectable={true}
+        pagination_mode={'fe'}
+        style_data_conditional={[{
+            if: { column_editable: true }, background_color: 'MediumPurple'
+        }]}
+        page_action='custom'
+        page_count={999999}
+        page_current={987654}
+        page_size={10}
```
Marc-Andre-Rivet, line 409: A simple visual test to see that numbers with 6 digits are displayed correctly.

alexcjohnson:
This seems reasonable, but I wonder if these need to be fixed widths? Is it problematic to let them grow with their content?

Marc-Andre-Rivet:
Inputs are weird beast and can't typically made to automatically grow with content, unlike other elements.

Typically this can be worked around by using something like (1) `<div contenteditable />` as the div will allow resizing. I could also look into wrapping the input into a parent, position it absolutely + take all the available space, and use an identically styled span or div with the same content to force a resize. I don't have a problem spending more time tweaking this into a more "dynamic" solution if we feel like it.

alexcjohnson:
Ah right, number validation and other bits of `<input>` behavior are nice. What if we size both the input and the total page count to fixed widths but based on the number of digits in `page_count` instead of globally constant?

PR #820, opened 2020-08-24
Julia components

waralex:
Julia's component generation is added to the build script


CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ Use the review apps to verify callback functionality (note these examples are wr
 `circleci build --job test`
 
 ## Local Build
-`npm run build:js && npm run build:py`
+`npm run build:js && npm run build:backends`
```
alexcjohnson, line 44: Looks like this is out of date and should just be:
```suggestion
`npm run build`
```

PR #831, opened 2020-09-24
Issue 295 - Header tooltips

### LINKED ISSUE #295: Header tooltips
Naming columns can be easy, and can be difficult sometimes. Even when it is easy, we often need to explain a little more about the data in a certain column without having to clutter the page / table with too much text. 
A good solution to this might be to create the ability to add a clickable question mark (or some other symbol) which shows a tooltip explaining more about the column. 
Something like this: 

screen shot 2018-12-12 at 1 01 02 am

Or maybe just hovering over a column name shows the optional tooltip: 

screen shot 2018-12-12 at 1 08 11 am

Thanks!

Marc-Andre-Rivet:
Closes #295 

* Adds `use_with` nested prop to `tooltip` to allow default column tooltips to be applied to either or both `data` and `header` cells (both by default)
* Adds new `tooltip_header` prop that behaves similarly to `tooltip_data` but affects headers instead


src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -1155,15 +1157,21 @@ export default class ControlledTable extends PureComponent<
...
+            const cell = header
+                ? table.querySelector(
+                      `tr:nth-of-type(${
+                          row + 1
+                      }) th[data-dash-column="${id}"]:not(.phantom-cell)`
+                  )
+                : table.querySelector(
+                      `td[data-dash-column="${id}"][data-dash-row="${row}"]:not(.phantom-cell)`
+                  );
+
```
Marc-Andre-Rivet, line 1174: Still using (row, id) but now also deciding what element to select based on `header`

src/dash-table/dash/DataTable.js
```diff
@@ -858,6 +860,7 @@ export const propTypes = {
     tooltip: PropTypes.objectOf(
         PropTypes.oneOfType([
             PropTypes.exact({
+                use_with: PropTypes.oneOf(['data', 'header']),
```
Marc-Andre-Rivet, line 863: Backward compatibility addition of `use_with` to the `tooltip` prop -- app creator can use it to define whether the default tooltip will be visible on data, headers, or both.

src/dash-table/dash/DataTable.js
```diff
@@ -955,6 +958,43 @@ export const propTypes = {
...
+                PropTypes.string,
+                PropTypes.exact({
+                    delay: PropTypes.number,
+                    duration: PropTypes.number,
+                    type: PropTypes.oneOf(['text', 'markdown']),
+                    value: PropTypes.string.isRequired
+                })
+            ])
+        )
+    ),
```
Marc-Andre-Rivet, line 996: header equivalent of `tooltip_data`

src/dash-table/derived/cell/eventHandlerProps.ts
```diff
@@ -74,6 +85,13 @@ class EventHandler {
                         rowIndex,
                         columnIndex
                     );
+                case Handler.MoveHeader:
+                    return handleMoveHeader.bind(
+                        undefined,
+                        this.propsFn,
+                        rowIndex,
+                        columnIndex
+                    );
```
Marc-Andre-Rivet, line 94: New event handlers for header cells, equivalent but slightly different from the same handlers for the data cells

src/dash-table/derived/header/wrappers.tsx
```diff
@@ -1,47 +1,85 @@
...
-): JSX.Element[][] {
-    return R.map(([labels, indices]) => {
-        return R.addIndex<number, JSX.Element>(R.map)((columnIndex, index) => {
-            const column = columns[columnIndex];
+    get = (
+        columns: Columns,
+        labelsAndIndices: R.KeyValuePair<any[], number[]>[],
+        mergeHeaders: boolean
+    ): JSX.Element[][] =>
+        labelsAndIndices.map(([labels, indices], rowIndex) =>
```
Marc-Andre-Rivet, line 25: Clean up. TypeScript + R.map + R.addIndex don't make a great mix, requiring explicit typing of everything in most cases. The native implementation is easier to read and is also typed.

tests/selenium/test_tooltip.py
```diff
@@ -138,3 +139,77 @@ def test_ttip003_tooltip_disappears(test):
...
+            "1": "text1",
+            "2": {"use_with": "data", "value": "text2"},
+            "3": {"use_with": "header", "value": "text3"},
+        },
+        "tooltip_data": [{"1": "alt-text1"}],
+        "tooltip_header": [{}, {"1": "alt-header1-row1"}],
+    }
+
+    app = dash.Dash(__name__)
+    app.layout = DataTable(**props)
```
Marc-Andre-Rivet, line 167: Test combination of static tooltips with/without usage and data/header tooltips to make sure usage is applied and overrides override.

tests/selenium/conftest.py
```diff
@@ -129,16 +129,22 @@ def __init__(self, id, mixin, col_id, state=_ANY):
...
 
-        return self.mixin.find_elements(
-            '#{} {} tbody tr th.dash-header[data-dash-column="{}"]:not(.phantom-cell)'.format(
-                self.id, self.state, self.col_id
+        return self.mixin.find_element(
+            '#{} {} tbody tr:nth-of-type({}) th.dash-header[data-dash-column="{}"]:not(.phantom-cell)'.format(
+                self.id, self.state, row + 1, self.col_id
             )
-        )[row]
+        )
```
Marc-Andre-Rivet, line 136: Cleaned this up a bit. Thankfully this wasn't used anywhere so no tests need to be modified.

tests/selenium/test_tooltip.py
```diff
@@ -138,3 +139,77 @@ def test_ttip003_tooltip_disappears(test):
...
+    assert tooltip.get_text().strip() == "text1"
+    cell55.move_to()
+
+    target.cell(0, "2").move_to()
+    wait.until(lambda: target.tooltip.exists(), 3)
+    assert tooltip.get_text().strip() == "text2"
+    cell55.move_to()
+
+    target.cell(0, "3").move_to()
+    time.sleep(1)
```
Marc-Andre-Rivet, line 193: For the missing case, ensure the tooltip has had time to appear, just in case.

src/dash-table/dash/DataTable.js
alexcjohnson: Can we add an explicit `'both'` to go along with the implicit one, and just label that as the default? Even though it won't be set as such since it's nested, I think people would in some cases appreciate being able to set this explicitly.

src/dash-table/dash/DataTable.js
```diff
@@ -955,6 +958,43 @@ export const propTypes = {
         )
     ),
 
+    /**
+     * `tooltip_header` represents the tooltip shown
+     * for different header columns and cells.
+     * The `property` name refers to the column ID. Each property
+     * contains a list of tooltips mapped to the table's `header`
+     * row index.
```
alexcjohnson, line 966: "Each property contains a list of tooltips" implies to me that we're expecting a structure:
```
{'col-id': ['first row', 'second row']}
```
rather than what I think we're actually expecting:
```
[{'col-id': 'first row'}, {'col-id': 'second row'}]
```
But the former actually feels a bit more intuitive and compact to me, even if it doesn't match `tooltip_data` as closely. It simplifies some simple cases: single row headers, or the same tooltip for every row, you would just provide a string or object for the value, rather than an array of strings or objects. What do you think?

Also minor point: I'd just say "The property name" rather than "The `property` name" - as is I was looking for something called "property" below 😅

Marc-Andre-Rivet: Agreed. The current description is inaccurate and misleading.

For consistency, I don't foresee any major problem supporting both syntax (array of objects and object of arrays) for both props.

src/dash-table/dash/DataTable.js
Marc-Andre-Rivet: Sure https://github.com/plotly/dash-table/pull/831/commits/71e7ce9b7bbc2752fc121afba4e74a2641061724

src/dash-table/dash/DataTable.js
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/831/commits/a2011e0acdeb1ca1da51d5b255d6e1ef95d5eb2c & https://github.com/plotly/dash-table/pull/831/commits/e4433595a01388746905fee36f48d0e239ab050a

src/dash-table/derived/table/tooltip.ts
```diff
@@ -24,30 +29,42 @@ function getSelectedTooltip(
...
 
     if (id === undefined || row === undefined) {
         return undefined;
     }
 
+    const staticTooltip = tooltip_static?.[id];
+    const staticUseWith =
+        staticTooltip && typeof staticTooltip !== 'string'
+            ? staticTooltip.use_with
+            : TooltipUsage.Both;
```
Marc-Andre-Rivet, line 44: Apply `tooltips` to both data and headers by default. This is a change from the pre-existing behavior that would have shown the tooltip only on the data cells but it seems consistent in spirit with the new functionality.

src/dash-table/dash/DataTable.js
alexcjohnson: Hmm if either syntax can cover every use case, I don't think it's a good idea to support both, that'll just confuse people. 

Seems to me the `*_data` props should all match `data` itself, ie `arrayOf(objectOf(...))`.

For `tooltip_header`, I mostly wanted to see if we could simplify the cases of single-row headers and multi-row headers with the same tooltip for all rows, and to do that I was thinking (1) `tooltip_header` would be an `objectOf` the row IDs, and (2) you could choose not to use an array inside that. So if all you want is simple text that applies the same to every row (of if there's only one row) you could use:
```js
{'col1-id': 'all rows for col1', 'col2-id': 'all rows for col2'}
```
or eg markdown for all rows:
```js
{'col1-id': {'type': 'markdown', 'value': '#Wow\nwhat great data!'}}
```
But of course you'd still need the array form when you have multiple rows and you want them to behave differently:
```js
{
  'col1-id': ['first row', 'second row'],
  'col2-id': [{'type': 'markdown', 'value': '#Big'}, {'type': 'markdown', 'value': '####small'}]
}
```

Marc-Andre-Rivet: Sure.. can update to have:

`tooltip`: objectOf(string | tooltip) w/ **new** `use_with` for (both,data,header)
`tooltip_conditional`: **unchanged** arrayOf(conditionalTooltip)
`tooltip_data`: **unchanged** arrayOf(objectOf(string | tooltip))
`tooltip_header`: **new** objectOf(string | tooltip | arrayOf(null|string|tooltip))`

src/dash-table/derived/table/tooltip.ts
```diff
@@ -24,34 +31,55 @@ function getSelectedTooltip(
...
+                      ifFilter(
+                          tt.if,
+                          virtualized.data[row - virtualized.offset.rows]
+                      ))
+              );
+          }, tooltip_conditional);
+
+    if (conditionalTooltips.length) {
+        return conditionalTooltips.slice(-1)[0];
+    }
```
alexcjohnson, line 55: nonblocking, but perhaps https://ramdajs.com/docs/#findLast would be cleaner here?

Marc-Andre-Rivet: Sure https://github.com/plotly/dash-table/pull/831/commits/8eec8201d28a2560fd01e078e178ddfa88a1284b

alexcjohnson (approved):
💃  Beautiful. Lovely tests, very easy to read and cover the cases well.

PR #841, opened 2020-10-19
Issue 840 - Fix proptypes & syntax highlighting, and improve tests

### LINKED ISSUE #840: Bad prop-type definition causing console errors
There's no such thing as https://github.com/plotly/dash-table/blob/dev/src/dash-table/dash/DataTable.js#L999

The proper way of defining this would have been 
```
                PropTypes.oneOfType([
                    PropTypes.oneOf([null]),
                    PropTypes.string,
                    // ...
```

Marc-Andre-Rivet:
Closes #840 

- adds log checks for all table tests 
- standardizes `tests/integration` tests with `tests/selenium` tests using the newer fixture / pattern
- removes `python-3.6` CI job (to be removed on approval) and `percy/dash-table-python-v0` (those two visual tests have been flaky since forever and never uncovered a bug anyway)

As part of improving the tests, found an additional regression breaking `highlight.js` after https://github.com/plotly/dash-table/pull/814 -- this does not affect `dcc` as it uses `9.x` and the usage there is correct.


alexcjohnson:
Removing Py3.6 tests - interesting... I guess there's an argument that components only need to test with the latest Python, as they have no independent Python code, and in the main dash repo we create test components and test them against multiple Python versions, verifying that the generated code will all work across versions.

That said table has some Python helper functions for formatting, which it would be nice to test will work across versions. On the other hand how much longer are we even going to support Py2, which would be the main reason for such a test?

So in the end I'd say yeah, this is fine.

src/third-party/highlight.js
```diff
@@ -1,4 +1,4 @@
-import highlightjs from 'highlight.js/lib/highlight';
+import highlightjs from 'highlight.js/lib/core';
```
alexcjohnson, line 1: Want to upgrade highlight in DCC too while this is front of mind?

tests/selenium/test_basic_copy_paste.py
```diff
@@ -85,6 +85,7 @@ def test_tbcp001_copy_paste_callback(test):
 
     assert target.cell(1, 0).get_text() == "0"
     assert target.cell(1, 1).get_text() == "MODIFIED"
+    assert len(test.get_log_errors()) == 0
```
alexcjohnson, line 88:
```suggestion
    assert test.get_log_errors() == []
```
That way if it fails you see what the error was

tests/selenium/test_markdown_copy_paste.py
```diff
@@ -7,9 +8,7 @@
 rawDf = pd.read_csv(url)
 rawDf["Complaint ID"] = rawDf["Complaint ID"].map(lambda x: "**" + str(x) + "**")
 rawDf["Product"] = rawDf["Product"].map(lambda x: "[" + str(x) + "](plot.ly)")
-rawDf["Issue"] = rawDf["Issue"].map(
-    lambda x: "![" + str(x) + "](https://dash.plot.ly/assets/images/logo.png)"
-)
```
alexcjohnson, line 12: Ha good call - really shouldn't be depending on dash docs to be up and never change where or what the logo is!

alexcjohnson (approved):
Looks great! Just one nonblocking suggestion for cleaner error handling. 💃

src/third-party/highlight.js
Marc-Andre-Rivet: Sure. And if tacking that on, might as well add Julia support.

tests/selenium/test_basic_copy_paste.py
Marc-Andre-Rivet: Good point https://github.com/plotly/dash-table/pull/841/commits/9e8028981e433d91ecd3fe4aa7bea16b3a7934f7

Marc-Andre-Rivet:
For https://github.com/plotly/dash-table/pull/841#issuecomment-712549984 - the formatter tests are currently run with the unit tests, on py37 only. The py36 tests are a misnomer as all that job really did was load two apps with table/graph and take a screenshot of them.

src/third-party/highlight.js
Marc-Andre-Rivet: https://github.com/plotly/dash-core-components/pull/871

PR #844, opened 2020-10-28
Issue 680 - `row` and Bootstrap interactions

### LINKED ISSUE #680: [BUG] Left-most column clipped / overflow hidden
First of all thank you for the great work on Dash.

Since the 4.6.0 update all tables in my Dash app have their left-most column clipped. This doesn't seem to be an issue with the styling / sizing of the dash-table object. The image below describes it best:

image

Furthermore, if I inspect the element, the whole table does size correctly, yet only the left-most column cells overflow.

Thank you in advance!

### LINKED ISSUE #796: DataTable  Cannot Show Full Table at Left and Right Edge
I created a datatable an place it in a dbc.Col, but the leftest and rightest columns of the table can only display partially. I wonder it's something wrong with using dataTable with dbc.Col/dbc.Row. The table screenshot is as below. Can anyone help me with this?
TextAlign:Right
image
TextAlign: Left
image
Also if I fix the header and then change the columns, cell width of header row and other rows will distort as following:
image
PS: I only have this problem after upgrade dash from 1.1.1 to 1.13.0. Everything work fine with Dash 1.1.1. Here are the packages versions:
dash==1.13.0
dash-bootstrap-components==0.10.2
dash-table==4.8.0
Thanks much!

Marc-Andre-Rivet:
Fixes #680 
Fixes #796

- adds the `table-` prefix to the `row` and `row-{number}` classes
- unapplies Bootstrap styling for radio inputs
- unapplies `line-height` changes
- updates the visual tests to use bootstrap

Also adds visual tests to the `tests/selenium` test run and adds the new `percy/dash-table-e2e` snapshots bucket


package.json
```diff
@@ -76,6 +76,7 @@
     "@types/react-dom": "^16.9.8",
     "@types/react-select": "^3.0.16",
     "babel-loader": "^8.1.0",
+    "bootstrap": "^4.5.3",
```
Marc-Andre-Rivet, line 79: Both bootstrap and jquery are only used for testing purposes.

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -1010,7 +1010,7 @@ export default class ControlledTable extends PureComponent<
                             <div
                                 key={`r${rowIndex}`}
                                 ref={`r${rowIndex}`}
-                                className={`row row-${rowIndex}`}
+                                className={`table-row table-row-${rowIndex}`}
```
Marc-Andre-Rivet, line 1013: Fixing the class collision with Bootstrap

src/dash-table/components/Table/Table.less
```diff
@@ -344,13 +357,13 @@
...
 
-                .row-0 {
+                .table-row-0 {
                     display: flex;
                     flex: 0 0 auto;
                     flex-direction: row;
                 }
 
-                .row-1 {
+                .table-row-1 {
```
Marc-Andre-Rivet, line 366: Same class collision with Bootstrap

tests/visual/percy-storybook/Border.defaults.percy.tsx
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import random from 'core/math/random';
 import DataTable from 'dash-table/dash/DataTable';
+import './common';
```
Marc-Andre-Rivet, line 6: Could technically add this to only one test and it would be applied to all but it seemed better to duplicate the import just so it's clear that this is "expected".

tests/visual/percy-storybook/common/index.tsx
```diff
@@ -0,0 +1,2 @@
+import 'bootstrap';
+import 'bootstrap/dist/css/bootstrap.css';
```
Marc-Andre-Rivet, line 2: Import all the bootstrap things. This is applied to all tests (whether they all import this file or not). As currently set up, this means that we are no longer testing a "vanilla" version of the table in these tests and we would be required to duplicate this test run to do so (e.g. an additional ~1600 snapshots per build)

There are some differences between vanilla and the bootstrap styled table, as far as I can tell the main difference seems to be a slightly different font size. Nothing that strikes as compromising the vanilla test cases.

tests/selenium/test_bootstrap.py
```diff
@@ -0,0 +1,79 @@
...
+    ops,
+    fixed_rows_description,
+    fixed_columns_description,
+    ops_description,
+):
+    test.start_server(get_app(fixed_rows, fixed_columns, ops))
+
+    test.table("table").is_ready()
+
+    time.sleep(3)
```
Marc-Andre-Rivet, line 73: The table "jiggles" to find the right column width with certain prop values (fixed_*). Adding this delay makes sure it's done "jiggling".

tests/selenium/conftest.py
```diff
@@ -376,9 +376,7 @@ def test(request, dash_thread_server, tmpdir):
         remote_url=request.config.getoption("remote_url"),
         headless=request.config.getoption("headless"),
         options=request.config.hook.pytest_setup_options(),
-        download_path=tmpdir.mkdir("download").strpath,
-        percy_assets_root=request.config.getoption("percy_assets"),
+        download_path=tmpdir.mkdir("dt-download").strpath,
```
Marc-Andre-Rivet, line 379: Table tests sometimes use `dash_thread_server` and `dash_duo` -- `dash_duo` specifically also creates a temporary folder, both fixtures clash and the tests end up throwing an error because the directory already exists. Pytest uses a py27 compatible object that does not expose `exists_ok` and hence the error can't otherwise be prevented.

tests/selenium/test_bootstrap.py
```diff
@@ -0,0 +1,79 @@
...
+        style_table=dict(height="500px", maxHeight="500px", overflow="auto"),
+        editable=True,
+        sort_action="native",
+        include_headers_on_copy_paste=True,
+        **fixed_rows,
+        **fixed_columns,
+        **ops
+    )
+
+    app.layout = html.Div([DataTable(**props)])
```
Marc-Andre-Rivet, line 34: Initially wanted this test to show a unmodified table + a another table in an iframe with bootstrap styles applied to offer some comparison point during review (e.g. only the bootstrap version changed vs. both) but the iframe content is never visible.

tests/selenium/test_bootstrap.py
alexcjohnson: Would it be worthwhile including some signal in the DOM, like `data-table-is-iterating` or something, that gets removed when this jiggling is done? That would be useful for dash-docs too, where we see some percy flakiness on these tables - we could just wait until there are no iterating DOM elements.

Marc-Andre-Rivet: @alexcjohnson Actually I think I'm wrong about this. While at some point I had added a loop to resize the table columns for the fixed scenarios, I've eventually removed it entirely. Will try removing the sleep and see if there's any inconsistencies. I'd like to see the inconsistencies in table rendering you've been seeing. Some recent bug fixes have changed some aspects of rendering and would have caused a slew of `1 to few` pixels alignment changes.

alexcjohnson: Great! Maybe those issues have been resolved - but if I see them again I'll let you know :)

alexcjohnson (approved):
💃  LGTM!

PR #857, opened 2020-12-17
use dash loosen-testing-reqs branch and fix linting

alexcjohnson:
sanity for https://github.com/plotly/dash/pull/1506

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
                     python -m venv venv || virtualenv venv
                     . venv/bin/activate
                     pip install -r dev-requirements.txt --quiet
-                    git clone --depth 1 git@github.com:plotly/dash.git dash-main
+                    git clone --depth 1 -b loosen-testing-reqs git@github.com:plotly/dash.git dash-main
                     pip install -e ./dash-main[dev,testing] --quiet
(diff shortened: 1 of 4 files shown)


dash_table_base/__init__.py
```diff
@@ -18,6 +15,9 @@
     )
     _sys.exit(1)
 
+from ._imports_ import *  # noqa: E402, F401, F403
+from ._imports_ import __all__  # noqa: E402
```
alexcjohnson, line 19: The test above needs to happen before these imports or it's useless, because the components will try to use `dash.development`.

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
                     python -m venv venv || virtualenv venv
                     . venv/bin/activate
                     pip install -r dev-requirements.txt --quiet
-                    git clone --depth 1 git@github.com:plotly/dash.git dash-main
+                    git clone --depth 1 -b loosen-testing-reqs git@github.com:plotly/dash.git dash-main
```
alexcjohnson, line 35: we can revert these branches and merge this PR even before https://github.com/plotly/dash/pull/1506 merges, because the changes here are fine with `dash@dev`

Marc-Andre-Rivet (approved):

PR #861, opened 2021-01-04
Fix pandas orient arg (to_dict("rows") -> to_dict("records"))

wbrgss:
**Description of Changes**
Similar to https://github.com/plotly/dash-docs/pull/1058, changes the incorrect `"rows"` argument to the parameter `orient` to the equivalent `"records"`.

Fixes a warning; `to_dict('rows')` causes the following on pandas 1.1.5:

`Using short name for 'orient' is deprecated. Only the options: ('dict', list, 'series', 'split', 'records', 'index') will be used in a future version. Use one of the above to silence this warning.`

See https://github.com/plotly/dash-docs/commit/07ac6fcd8a711a1abfeef6d7bdb1f883f17684cc for a description of how this got mixed up:
> rows “happens” to work because pandas only looks at the first letter (“r”)

***

I also `black`ened some Selenium tests, as black `18.6b4` (the version in this repo's `requirements.txt`) complained on the `husky` pre-commit. I also added two directories to the black `--exclude` option, as I had a (non-hidden/dotfile) `venv` folder, as [suggested](https://github.com/plotly/dash-table/blob/dev/CONTRIBUTING.md#local-server-review-apps) in this repo's CONTRIBUTING.md, and I got 

```sh
error: cannot format /home/plotly/Workspace/dash-table/dash_table_base/__init__.py: 
All done! 💥 💔 💥
27 files would be left unchanged, 1 file would **fail** to reformat.
```
(`**` emphasis added)

when running `git commit` (i.e. `black` on the pre-commit) without excluding `dash_table_base`. cc @Marc-Andre-Rivet because these tests haven't been modified for ~10 months, so it might be a difference in my environment...


package.json
```diff
@@ -24,11 +24,11 @@
     "private::build:backends": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json --r-prefix 'dash' --r-suggests 'dash' --jl-prefix 'dash'",
     "private::format.ts": "npm run private::lint.ts -- --fix",
     "private::format.prettier": "prettier --config .prettierrc --write \"src/**/*.{js,ts,tsx}\"",
-    "private::format.black": "black --exclude dash_table .",
+    "private::format.black": "black --exclude '/(dash_table|venv|dash_table_base)/' .",
```
alexcjohnson, line 27: The exclusion of `dash_table` makes sense because it's mainly autogenerated files, but we should be able to blacken `dash_table_base`, shouldn't we?

In fact in the dash repo we just list the directories we DO want to run black on:
```
"private::format.black": "black dash tests --exclude metadata_test.py"
```
so I'd think here the equivalent would be `black dash_table_base tests`

wbrgss: @alexcjohnson Thanks for the suggestion; I agree that's more consistent. In 6408184

wbrgss: I excluded `dash_table_base` because I couldn't get black to format it, I got 

```
error: cannot format dash_table_base/__init__.py: 
All done! 💥 💔 💥
1 file failed to reformat.
```

without any indication of what was wrong. After removing a couple commas I managed to get

```
error: cannot format dash_table_base/__init__.py: Cannot parse: 12:12:         file=_sys.stderr
```

so I guessed that the `_` short name in `from __future__ import print_function` was confusing black when parsing `_`-prefixed variables and I removed it in 32b1ceb, which fixed the problem. I'm not sure why there was an alias there, as AFAIK this `print_function` is never explicitly called.

alexcjohnson: Annoying, I've never seen black behave like that, but I'm glad you sorted it out!

alexcjohnson (approved):
💃 Thanks for fixing this!

PR #878, opened 2021-03-18
Update toolchain

Marc-Andre-Rivet:
Depends on https://github.com/plotly/dash/pull/1563

Updates the table "build" to use Webpack 5 and newer versions of the toolchain packages (e.g. babel, typescript, loaders).

There's a chain of side-effects at play here and most (but not all) of this PR is to deal with the side-effects. Depending on which, they are more or less difficult to fix:

- `Storybook`, which we use here to generate the Percy snapshots hooks into the webpack configuration and expects Webpack 4.x -- some changes in tooling is done here to make it work with Webpack 5. Sadly there are "alpha" releases, which is a similar problem we've run into in the past. Storybook is notoriously slow at supporting new versions of key third-parties
 
- `Cypress` has been in a derelict state for our purpose for quite a while and this Webpack 5 update is causing problems that can't be fixed (as far as I can tell) in the unit tests -- this PR moves the unit tests from using Cypress to another stack that has been around for a lot of time and is very stable.. Karma/Mocha/Chai, since the cypress test syntax already was compatible with Mocha/Chai, it left only the runner to be updated. For context, mocha tests are normally run directly in Node which won't do for our purposes, Karma is used, amongst other things, to load the tests into the browser to be run there. Additional context, Karma has been around for a very long time (it's ancient by web development standards, ~2012), this will hopefully provide more solid footing for the unit tests

- More on `Cypress`, back in 2018 I introduced Cypress into the table stack to run integration tests as our Selenium setup was less than ideal at the time, this has proven problematic ever since Cypress 4.x came out. This PR includes a port of the remaining Cypress integration tests to Selenium. The tests are ported mostly "as-is" and while some improvements are made, tests that are sub-par are still basically just ported and left to be improved and re-organized at a later time.

- Update other dependencies. `react-select` can/t automatically be updated as the API changed and the visual style changed -- this would require deeper rework/investigation. Updating `@types/ramda` causes typing issues that have to do with the types definition and not with the underlying library -- as for react-select, fixing this would require a deeper investigation of the side-effects. Frequently the port will also involve running the test in more scenarios than it did before.

- Need to remove `ci/circleci: standalone-test` from the required build steps

All-in-all this PR proposes to update the table toolchain in depth as part of keeping up with our third party dependencies as the ever growing gap has now reached chasm proportions... one must not fear the reaper.


.circleci/config.yml
```diff
@@ -3,39 +3,28 @@ version: 2.1
 jobs:
     "server-test":
         docker:
-            - image: circleci/python:3.7.9-node-browsers
+            - image: circleci/python:3.9.2-buster-node-browsers
```
Marc-Andre-Rivet, line 6: Updating to use py3.9 as per https://github.com/plotly/dash-core/issues/203

Marc-Andre-Rivet: For now, can't use the new CircleCI `cimg` because `CHROME_BIN` is missing and expected by Karma

.circleci/config.yml
```diff
@@ -56,101 +44,39 @@ jobs:
                 command: npx percy finalize --all
                 when: always
 
-
-    "standalone-test":
```
Marc-Andre-Rivet, line 60: These tests have been ported to test-server / Selenium

.circleci/config.yml
```diff
@@ -56,101 +44,39 @@ jobs:
...
                     . venv/bin/activate
-                    npm run build
                     npm run test.unit
 
-
     "visual-test":
         docker:
-            - image: circleci/node:10-browsers
-
+            - image: circleci/node:14-browsers
```
Marc-Andre-Rivet, line 71: Node14 is the current active LTS and some npm packages are starting to drop v10

.config/webpack/base.js
```diff
@@ -40,15 +39,18 @@ module.exports = (options = {}) => {
...
+                    }
                 },
                 {
                     test: /\.csv$/,
                     loader: 'raw-loader'
                 },
                 {
                     test: /\.ts(x?)$/,
-                    include: /node_modules[\\\/](highlight[.]js)[\\\/]/,
+                    include: /node_modules[\\\/](highlight[.]js|d3-format)[\\\/]/,
```
Marc-Andre-Rivet, line 53: `d3-format` v2 is now bundled in ES6, we need to transpile it to ES5 for our IE11 support.

.config/webpack/base.js
```diff
@@ -108,7 +109,7 @@ module.exports = (options = {}) => {
         optimization: {
             splitChunks: {
                 chunks: 'async',
-                name: true,
+                name: '[name].js',
```
Marc-Andre-Rivet, line 112: Changed from Wepback4 -> 5

.config/webpack/base.js
```diff
@@ -25,7 +25,6 @@ module.exports = (options = {}) => {
         mode: mode,
         output: {
             path: path.resolve(__dirname, `./../../${dashLibraryName}`),
-            chunkFilename: '[name].js',
```
Marc-Andre-Rivet, line 28: Changed from Wepback4 -> 5 -- see `optimization` section

.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = C901, E203, E266, E501, E731, W503
+ignore = C901, E203, E231, E266, E501, E731, W503
```
Marc-Andre-Rivet, line 2: Flask / Black spacing and trailing `,`, deactivating in Flask and respecting Black formatting instead

karma.conf.js
```diff
@@ -0,0 +1,23 @@
+module.exports = config => {
+    config.set({
```
Marc-Andre-Rivet, line 2: Karma replaces Cypress as the unit test runner. Configuration to run in the desired env the desired tests correctly bundled.

package.json
```diff
@@ -30,98 +26,106 @@
...
-    "test.watch": "run-p --race \"private::build:js-test-watch\" --race private::host* private::opentests"
+    "test.visual-local": "build-storybook"
   },
   "author": "Chris Parmer <chris@plotly.com>",
   "maintainer": "Ryan Patrick Kyle <ryan@plotly.com>",
   "license": "MIT",
   "devDependencies": {
-    "@babel/cli": "^7.10.5",
-    "@babel/core": "^7.11.1",
+    "@babel/cli": "^7.13.10",
```
Marc-Andre-Rivet, line 45: Updating pretty much everything else too..

src/dash-table/LazyLoader.ts
```diff
@@ -7,7 +7,7 @@ export default class LazyLoader {
 
     public static get hljs() {
         return Promise.resolve(
-            window.hljs ||
+            (window as any).hljs ||
```
Marc-Andre-Rivet, line 10: Newer version of Typescript is picking that one up

src/dash-table/components/ControlledTable/index.tsx
```diff
@@ -55,9 +55,7 @@ const INNER_STYLE = {
     minWidth: '100%'
 };
 
-export default class ControlledTable extends PureComponent<
-    ControlledTableProps
-> {
+export default class ControlledTable extends PureComponent<ControlledTableProps> {
```
Marc-Andre-Rivet, line 58: Formatting changes with newer Prettier version

src/dash-table/utils/Markdown.ts
```diff
@@ -36,7 +36,7 @@ export default class Markdown {
     private static hljsResolve: () => any;
     private static _isReady: Promise<boolean> | true = new Promise<boolean>(
         resolve => {
-            Markdown.hljsResolve = resolve;
+            Markdown.hljsResolve = resolve as any;
```
Marc-Andre-Rivet, line 39: tsc picks up on this one too with the new version

tests/js-unit/clipboard_test.ts
```diff
@@ -1,3 +1,4 @@
+import { assert, expect } from 'chai';
```
Marc-Andre-Rivet, line 1: These imports are known for free in the Cypress stack but with Karma/Mocha/Chai they need to be imported

tests/js-unit/formatting_test.ts
```diff
@@ -25,7 +27,7 @@ describe('formatting', () => {
 
                 expect(formatter(0)).to.equal('$0.00');
                 expect(formatter(1)).to.equal('$1.00');
-                expect(formatter(-1)).to.equal('-$1.00');
+                expect(formatter(-1)).to.equal('−$1.00');
```
Marc-Andre-Rivet, line 30: `d3-format` v2 changed the default minus sign
https://github.com/d3/d3-format/releases/tag/v2.0.0

Since this is presentational/formatting only wouldn't call this breaking.

tests/js-unit/lexeme_test.ts
```diff
@@ -20,17 +22,17 @@ describe('expression', () => {
...
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: 'abc' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: '{abc' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: 'abc}' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: '}abc{' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: '{{abc}}' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: '"' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: '`' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: `'` } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: '{' } as ISyntaxTree)).to.throw();
+            expect(fieldExpression.resolve.bind(undefined, {}, { value: '}' } as ISyntaxTree)).to.throw();
```
Marc-Andre-Rivet, line 35: Small API difference here, tweaking to check for a throw instead of a specific type

tests/selenium/conftest.py
```diff
@@ -64,6 +63,29 @@ def double_click(self):
         ac.double_click()
         return ac.perform()
 
+    def exists(self):
```
Marc-Andre-Rivet, line 66: Lots of ported tests from this point forward and added capabilities in the table testing helper. Will not comment on individual test ports.

package.json
```diff
@@ -30,98 +26,106 @@
...
+    "@fortawesome/react-fontawesome": "^0.1.14",
+    "@percy/storybook": "^3.3.1",
     "@plotly/dash-component-plugins": "^1.2.0",
-    "@plotly/webpack-dash-dynamic-import": "^1.1.5",
-    "@storybook/cli": "^6.0.5",
-    "@storybook/react": "^6.0.5",
-    "@types/d3-format": "^1.3.1",
-    "@types/highlight.js": "^9.12.4",
-    "@types/papaparse": "^5.0.6",
+    "@plotly/webpack-dash-dynamic-import": "^1.2.0",
```
Marc-Andre-Rivet, line 58: The actual change from https://github.com/plotly/dash/pull/1563

.circleci/config.yml
```diff
@@ -3,39 +3,28 @@ version: 2.1
...
-
-            - restore_cache:
-                key: dep-{{ .Branch }}-{{ checksum "package-lock.json" }}-{{ checksum "package.json" }}-{{ checksum ".circleci/config.yml" }}
             - run:
                 name: Install npm packages
                 command: npm ci
-            - save_cache:
-                key: dep-{{ .Branch }}-{{ checksum "package-lock.json" }}-{{ checksum "package.json" }}-{{ checksum ".circleci/config.yml" }}
-                paths:
-                    - node_modules
```
alexcjohnson, line 26: Is it intentional to drop the cache? Or just temporary while testing?

.circleci/config.yml
```diff
@@ -3,39 +3,28 @@ version: 2.1
...
-
             - run:
                 name: Install requirements
                 command: |
-                    sudo pip install --upgrade virtualenv
+                    pip install --upgrade virtualenv
```
alexcjohnson, line 21: We don't have any Py2 tests here anymore, can we drop `virtualenv`?

.circleci/config.yml
Marc-Andre-Rivet: Temporary. Adding back. 🙈

.circleci/config.yml
alexcjohnson: `test-server` is awfully long now, nearly an hour. Can we parallelize it, shoot for <10min?

tests/js-unit/clipboard_test.ts
alexcjohnson: `assert` isn't used in this file - are we not linting the test scripts?

tests/js-unit/formatting_test.ts
alexcjohnson: I'm glad `d3-format` has caught up, we've had to patch this for years in plotly.js to get the correct minus sign 😎

.circleci/config.yml
Marc-Andre-Rivet: @alexcjohnson I've tried but for some reason one of the test consistently fails with parallelism turned on. I've got no idea why. Can try and figure it out before we consider this PR ready.

tests/selenium/test_header.py
```diff
@@ -0,0 +1,134 @@
...
+    title = [
+        target.column("rows").get_text(0),
+        target.column("rows").get_text(1),
+        target.column("rows").get_text(2),
+    ]
+
+    target.column("rows").edit(row)
+
+    alert = test.driver.switch_to.alert
+    time.sleep(2)
```
alexcjohnson, line 59: That's a pretty long wait, isn't there some condition we can wait for instead?

tests/js-unit/clipboard_test.ts
Marc-Andre-Rivet: Those folders were not linted before. Now covered (and more) with: https://github.com/plotly/dash-table/pull/878/commits/381e49de1d46e57bef864c221677b5e7f602e0ae, https://github.com/plotly/dash-table/pull/878/commits/c8985e37440e586c0a31b03fbd97c08aa8e23859,https://github.com/plotly/dash-table/pull/878/commits/6c086e393ff3fa9e0abff941b9c3618cfc40d7f1

tests/selenium/test_header.py
Marc-Andre-Rivet: 🔪 https://github.com/plotly/dash-table/pull/878/commits/4378ec00ae1d9d2520f9a2845c2d45479ae16037 - tests behave fine locally without it 🤞

.circleci/config.yml
Marc-Andre-Rivet: Related, there's one set of tests that is extremely long to run with Selenium (lots of DOM checks). The minimal run time with per-file slicing is probably not lower than 15-20minutes because of that test and its variations.

Out of scope for this PR: it should be possible to rewrite that test to run the checks directly from the browser (Karma) instead and improve its speed by at least an order of magnitude.

Marc-Andre-Rivet: @alexcjohnson Figured out what I did wrong. Also, broke down the large test with lots of permutations into many smaller tests in their own files. The last run took ~28 minutes but I expect the CI time splitting algorithm to be able to optimize this down to below 20 minutes soon enough.

alexcjohnson: > Figured out what I did wrong

Excellent, which change fixed this?

> I expect the CI time splitting algorithm to be able to optimize this down to below 20 minutes soon enough.

OK, that's reasonable for now, thanks!

alexcjohnson (commented):
Lovely 💃 🚀

alexcjohnson (approved):

.circleci/config.yml
Marc-Andre-Rivet: > which change fixed this?

When I updated the CI yaml to run tests in parallel, I copied it over from the DCC CI yaml which runs pytest with the `--headless` flag. For some reason `tbmu001`, specifically, doesn't pass when it's run headless. Not absolutely sure why as I didn't investigate it further but headless Chrome is known for sometimes causing different behaviors

Marc-Andre-Rivet:
🔪 `ci/circleci: standalone-test` step check as it has been removed from the CI run before merging.

PR #609, opened 2019-10-01
Issue 545 - Case-insensitive option for filters

dmt0:
Closes #545
Implements case-insensitive filtering.

TODO:
- [x] Tests
- [x] Incorporate changes from https://github.com/plotly/dash-table/pull/612
- [x] GUI for filters
- [x] GUI for global filter
- [x] tests for GUI + percy tests
- [ ] Update https://dash.plot.ly/datatable/filtering with new operators, etc.

PR #893, opened 2021-04-29
Issue 545 - Case (in)sensitive filtering

### LINKED ISSUE #545: case insensitive filtering
As discussed in https://community.plot.ly/t/case-insensitive-filter-search-for-dash-datatables/26258

Marc-Andre-Rivet:
Closes #545. Uses PR #609 as a starting point.

Still missing:
- e2e testing

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,6 @@
-{
+var path = require('path');
+
+module.exports = {
     "extends": [
         "plugin:@typescript-eslint/recommended",
         "prettier"
@@ -7,7 +9,7 @@
         "@typescript-eslint"
     ],
     "parserOptions": {
-        "project": "./tsconfig.lint.json"
+        "project": path.join(__dirname, "tsconfig.lint.json"),
     },
     "rules": {
         "arrow-parens": [
--- a/src/core/syntax-tree/lexer.ts
+++ b/src/core/syntax-tree/lexer.ts
@@ -10,6 +10,7 @@ export interface ILexerResult {
 
 export interface ILexemeResult {
     lexeme: ILexeme;
+    flags?: string;
     value?: string;
 }
 
@@ -37,8 +38,10 @@ export default function lexer(lexicon: Lexicon, query: string): ILexerResult {
             return {lexemes: result, valid: false, error: query};
         }
 
-        const value = (query.match(next.regexp) || [])[next.regexpMatch || 0];
-        result.push({lexeme: next, value});
+        const match = query.match(next.regexp) ?? [];
+        const value = match[next.regexpMatch || 0];
+        const flags = match[next.regexpFlags || -1];
+        result.push({lexeme: next, flags, value});
 
         query = query.substring(value.length);
     }
--- a/src/dash-table/components/Filter/Column.tsx
+++ b/src/dash-table/components/Filter/Column.tsx
@@ -2,17 +2,20 @@ import React, {CSSProperties, PureComponent} from 'react';
 
 import IsolatedInput from 'core/components/IsolatedInput';
 
-import {ColumnId} from 'dash-table/components/Table/props';
+import {ColumnId, FilterCase} from 'dash-table/components/Table/props';
 import TableClipboardHelper from 'dash-table/utils/TableClipboardHelper';
+import FilterOptions from 'dash-table/components/Filter/FilterOptions';
 
 type SetFilter = (ev: any) => void;
 
 interface IColumnFilterProps {
     className: string;
     columnId: ColumnId;
+    filterOptions: FilterCase;
     isValid: boolean;
     setFilter: SetFilter;
     style?: CSSProperties;
+    toggleFilterOptions: () => void;
     value?: string;
 }
 
@@ -41,7 +44,15 @@ export default class ColumnFilter extends PureComponent<
     };
 
     render() {
-        const {className, columnId, isValid, style, value} = this.props;
+        const {
+            className,
+            columnId,
+            filterOptions,
+            isValid,
+            style,
+            toggleFilterOptions,
+            value
+        } = this.props;
 
         return (
             <th
@@ -61,6 +72,10 @@ export default class ColumnFilter extends PureComponent<
                     placeholder={'filter data...'}
                     stopPropagation={true}
                     submit={this.submit}
+                />
+                <FilterOptions
+                    filterOptions={filterOptions}
+                    toggleFilterOptions={toggleFilterOptions}
                 />
             </th>
         );
--- a/src/dash-table/components/Filter/FilterOptions.tsx
+++ b/src/dash-table/components/Filter/FilterOptions.tsx
@@ -1,4 +1,4 @@
-import React, {memo} from 'react';
+import React from 'react';
 
 import {FilterCase} from 'dash-table/components/Table/props';
 
@@ -7,17 +7,15 @@ interface IFilterCaseButtonProps {
     toggleFilterOptions: () => void;
 }
 
-export default memo(
-    ({filterOptions, toggleFilterOptions}: IFilterCaseButtonProps) => (
-        <input
-            type='button'
-            className={`dash-filter--case ${
-                filterOptions === FilterCase.Insensitive
-                    ? 'dash-filter--case--sensitive'
-                    : 'dash-filter--case--insensitive'
-            }`}
-            onClick={toggleFilterOptions}
-            value='Aa'
-        />
-    )
-);
+export default ({
+    filterOptions,
+    toggleFilterOptions
+}: IFilterCaseButtonProps) => (<input
+    type='button'
+    className={`dash-filter--case ${filterOptions === FilterCase.Sensitive
+            ? 'dash-filter--case--sensitive'
+            : 'dash-filter--case--insensitive'
+        }`}
+    onClick={toggleFilterOptions}
+    value='Aa'
+/>);
--- a/src/dash-table/components/Filter/FilterOptions.tsx
+++ b/src/dash-table/components/Filter/FilterOptions.tsx
@@ -1,0 +1,23 @@
+import React, {memo} from 'react';
+
+import {FilterCase} from 'dash-table/components/Table/props';
+
+interface IFilterCaseButtonProps {
+    filterOptions: FilterCase;
+    toggleFilterOptions: () => void;
+}
+
+export default memo(
+    ({filterOptions, toggleFilterOptions}: IFilterCaseButtonProps) => (
+        <input
+            type='button'
+            className={`dash-filter--case ${
+                filterOptions === FilterCase.Insensitive
+                    ? 'dash-filter--case--sensitive'
+                    : 'dash-filter--case--insensitive'
+            }`}
+            onClick={toggleFilterOptions}
+            value='Aa'
+        />
+    )
+);
--- a/src/dash-table/components/FilterFactory.tsx
+++ b/src/dash-table/components/FilterFactory.tsx
@@ -13,8 +13,7 @@ import {
     TableAction,
     IFilterFactoryProps,
     SetFilter,
-    FilterLogicalOperator,
-    FilterCase
+    FilterLogicalOperator
 } from 'dash-table/components/Table/props';
 import derivedFilterStyles, {
     derivedFilterOpStyles
@@ -74,7 +73,6 @@ export default class FilterFactory {
     private filter = memoizerCache<[ColumnId, number]>()(
         (
             column: IColumn,
-            filterOptions: FilterCase,
             index: number,
             map: Map<string, SingleColumnSyntaxTree>,
             operator: FilterLogicalOperator,
@@ -88,7 +86,7 @@ export default class FilterFactory {
                     key={`column-${index}`}
                     className={`dash-filter column-${index}`}
                     columnId={column.id}
-                    filterOptions={filterOptions}
+                    filterOptions={column.filter_options}
                     isValid={!ast || ast.isValid}
                     setFilter={this.onChange.bind(
                         this,
@@ -121,7 +119,6 @@ export default class FilterFactory {
     ) {
         const {
             filter_action,
-            filter_options,
             map,
             row_deletable,
             row_selectable,
@@ -160,7 +157,6 @@ export default class FilterFactory {
             (column, index) => {
                 return this.filter.get(column.id, index)(
                     column,
-                    filter_options,
                     index,
                     map,
                     filter_action.operator,
--- a/src/dash-table/components/FilterFactory.tsx
+++ b/src/dash-table/components/FilterFactory.tsx
@@ -13,7 +13,8 @@ import {
     TableAction,
     IFilterFactoryProps,
     SetFilter,
-    FilterLogicalOperator
+    FilterLogicalOperator,
+    FilterCase
 } from 'dash-table/components/Table/props';
 import derivedFilterStyles, {
     derivedFilterOpStyles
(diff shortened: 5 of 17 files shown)


.eslintrc.js
```diff
@@ -7,7 +9,7 @@
         "@typescript-eslint"
     ],
     "parserOptions": {
-        "project": "./tsconfig.lint.json"
+        "project": path.join(__dirname, "tsconfig.lint.json"),
```
Marc-Andre-Rivet, line 12: Just picking up some rocks.. the configuration was valid but caused an error in VSCode, preventing the eslint extension from working correctly.

src/core/syntax-tree/lexer.ts
```diff
@@ -37,8 +38,10 @@ export default function lexer(lexicon: Lexicon, query: string): ILexerResult {
             return {lexemes: result, valid: false, error: query};
         }
 
-        const value = (query.match(next.regexp) || [])[next.regexpMatch || 0];
-        result.push({lexeme: next, value});
+        const match = query.match(next.regexp) ?? [];
+        const value = match[next.regexpMatch || 0];
+        const flags = match[next.regexpFlags || -1];
+        result.push({lexeme: next, flags, value});
```
Marc-Andre-Rivet, line 44: From a lexicon perspective treat `i|s` prefix to applicable relational operators as "flags" that can be used to determine the exact processing to do during evaluation. `value` is the "base" operator + flags (e.g. `icontains`)

src/dash-table/components/Filter/Column.tsx
```diff
@@ -62,6 +73,10 @@ export default class ColumnFilter extends PureComponent<
                     stopPropagation={true}
                     submit={this.submit}
                 />
+                <FilterOptions
+                    filterOptions={filterOptions}
+                    toggleFilterOptions={toggleFilterOptions}
+                />
```
Marc-Andre-Rivet, line 79: New component displayed in the filter cell to toggle filtering options

src/dash-table/components/FilterFactory.tsx
```diff
@@ -80,6 +95,11 @@ export default class FilterFactory {
                         operator,
                         setFilter
                     )}
+                    // Running into TypeScript binding issues with many parameters..
+                    // bind with no more than 4 params each time.. sigh..
+                    toggleFilterOptions={this.onToggleChange
+                        .bind(this, column, map, operator, setFilter)
+                        .bind(this, toggleFilterOptions, ast && ast.query)}
```
Marc-Andre-Rivet, line 102: So TypeScript defaults to a generic `...args: <type of first arg>[]` signature when there's more than 4 arguments apart from context. Breaking it up just so I don't get an error. Annoying limitation.

src/dash-table/components/Table/props.ts
```diff
@@ -193,6 +198,7 @@ export interface IBaseColumn {
     clearable?: boolean | boolean[] | 'first' | 'last';
     deletable?: boolean | boolean[] | 'first' | 'last';
     editable: boolean;
+    filter_options: FilterCase;
```
Marc-Andre-Rivet, line 201: FilterCase vs. `filter_options` may look inconsistent but is intended to be that way. In the future you may also want to add locale-aware comparisons and leaving some space to do so at a future time.

src/dash-table/components/FilterFactory.tsx
```diff
@@ -57,13 +57,27 @@ export default class FilterFactory {
...
+        column: IColumn,
+        map: Map<string, SingleColumnSyntaxTree>,
+        operator: FilterLogicalOperator,
+        setFilter: SetFilter,
+        toggleFilterOptions: (column: IColumn) => IColumn,
+        value: any
+    ) => {
+        const newColumn = toggleFilterOptions(column);
+
+        updateColumnFilter(map, newColumn, operator, value, setFilter);
```
Marc-Andre-Rivet, line 70: Trigger a `setProps` and return the new column configuration, use the new column configuration to update the column's query (and trigger a refresh)

src/dash-table/derived/table/index.tsx
```diff
@@ -26,25 +28,53 @@ const handleSetFilter = (
...
+    });
 }
 
 export default (propsFn: () => ControlledTableProps) => {
     const setFilter = memoizeOne((setProps: SetProps, setState: SetState) =>
         handleSetFilter.bind(undefined, setProps, setState)
     );
 
+    const toggleFilterOptions = memoizeOne(
+        (setProps: SetProps, columns: IColumn[]) => (column: IColumn) => {
```
Marc-Andre-Rivet, line 51: Called by https://github.com/plotly/dash-table/pull/893/files#r622661715

Defined at the factory level so that the individual components remain unaware of `columns` for memoization/cache busting / performance reasons and `setProps` so they remain unaware of Dash's idiosyncracies.

src/dash-table/derived/table/index.tsx
```diff
@@ -26,25 +28,53 @@ const handleSetFilter = (
...
+            newColumn.filter_options =
+                newColumn.filter_options === FilterCase.Insensitive
+                    ? FilterCase.Sensitive
+                    : FilterCase.Insensitive;
+
+            newColumns.splice(iColumn, 1, newColumn);
+
+            setProps({columns: newColumns});
+
+            return newColumn;
```
Marc-Andre-Rivet, line 65: update the table's columns in the shallowest way possible (minimize cache busting)

src/dash-table/syntax-tree/SingleColumnSyntaxTree.ts
```diff
@@ -15,13 +21,22 @@ import {
...
-function getImplicitLexeme(type: ColumnType = ColumnType.Any): ILexemeResult {
+function getImplicitLexeme(
+    filterOptions: FilterCase | undefined,
+    type: ColumnType = ColumnType.Any
+): ILexemeResult {
+    const flags = R.isNil(filterOptions)
+        ? ''
+        : filterOptions === FilterCase.Insensitive
+        ? 'i'
+        : 's';
```
Marc-Andre-Rivet, line 32: Only apply flags to implicit relational operators. If the user wrote it, we must assume that's what they want!

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -44,49 +44,80 @@ const LEXEME_BASE = {
...
+
+const fnEval = (
+    fn: (lhs: any, rhs: any) => boolean,
+    lhs: any,
+    rhs: any,
+    relOp: string
+): boolean =>
+    relOp[0] == 'i'
+        ? fn(lhs.toString().toUpperCase(), rhs.toString().toUpperCase())
+        : fn(lhs, rhs);
```
Marc-Andre-Rivet, line 68: Conditional evaluation of contains, =, !=, >=, >, <=, < based on the flags

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -44,49 +44,80 @@ const LEXEME_BASE = {
...
-                op.toString().indexOf(exp.toString()) !== -1
+            ([lhs, rhs, relOp]) =>
+                !R.isNil(rhs) &&
+                !R.isNil(lhs) &&
+                (R.type(rhs) === 'String' || R.type(lhs) === 'String') &&
+                containsEval(lhs, rhs, relOp)
         ),
         subType: RelationalOperator.Contains,
-        regexp: /^((contains)(?=\s|$))/i,
+        regexp: /^((i|s)?contains)(?=\s|$)/i,
```
Marc-Andre-Rivet, line 80: allow for optional flags

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -44,49 +44,80 @@ const LEXEME_BASE = {
...
+        regexpFlags: 2,
```
Marc-Andre-Rivet, line 81: regex match group for usage in https://github.com/plotly/dash-table/pull/893/files#r622660202

src/dash-table/syntax-tree/lexeme/relational.ts
```diff
@@ -44,49 +44,80 @@ const LEXEME_BASE = {
...
 export const equal: IUnboundedLexeme = R.merge(
     {
-        evaluate: relationalEvaluator(([op, exp]) =>
-            isNumeric(op) && isNumeric(exp) ? +op === +exp : op === exp
+        evaluate: relationalEvaluator(([lhs, rhs, relOp]) =>
+            equalEval(lhs, rhs, relOp)
         ),
         subType: RelationalOperator.Equal,
-        regexp: /^(=|(eq)(?=\s|$))/i,
+        regexp: /^((i|s)?(=|(eq)(?=\s|$)))/i,
```
Marc-Andre-Rivet, line 93: same thing for all applicable relational operators

tests/js-unit/dash_table_queries_test.ts
```diff
@@ -265,6 +265,20 @@ describe('Dash Table Queries', () => {
...
+                                target: {a: 'abc'},
+                                valid: true,
+                                evaluate: true
+                            },
+                            {
+                                name: "compare 'abc' to 'ABC' (sensitive)",
+                                query: `${c.hideOperand ? '' : '{a} '}seq ABC`,
+                                target: {a: 'abc'},
+                                valid: true,
+                                evaluate: false
```
Marc-Andre-Rivet, line 281: Multiple new tests for the (in)sensitive versions of the operators -- as used by the table itself

tests/js-unit/query_syntactic_tree_test.ts
```diff
@@ -646,6 +646,22 @@ describe('Query Syntax Tree', () => {
...
+            expect(tree.evaluate({a: 'ABC'})).to.equal(true);
+        });
+
+        it('can do equality (seq) test', () => {
+            const tree = new QuerySyntaxTree('{a} seq "abc"');
+
+            expect(tree.isValid).to.equal(true);
+            expect(tree.evaluate({a: 'abc'})).to.equal(true);
+            expect(tree.evaluate({a: 'ABC'})).to.equal(false);
+        });
```
Marc-Andre-Rivet, line 663: Test all new operators with the various query flavors

src/dash-table/components/Filter/FilterOptions.tsx
```diff
@@ -0,0 +1,24 @@
...
+interface IFilterCaseButtonProps {
+    filterOptions: FilterCase;
+    toggleFilterOptions: () => void;
+}
+
+export default ({
+    filterOptions,
+    toggleFilterOptions
+}: IFilterCaseButtonProps) => (
+    <input
```
wbrgss, line 14: Can we add a `title` attribute here as well? Something like 'Toggle filter case sensitivity'

wbrgss (commented):
Looking nice so far! I'll probably do another more fine-grained pass of the code later. The functionality looks good.

If you have time @alexcjohnson I'd welcome a review from you as well

tests/selenium/test_filter.py
```diff
@@ -78,3 +78,98 @@ def test_filt001_basic(test, props, expect):
...
+    )
+
+    sensitivity = (
+        filter_options if column_filter_options is None else column_filter_options
+    )
+
+    test.start_server(get_app(props))
+
+    target = test.table("table")
+    # time.sleep(5000)
```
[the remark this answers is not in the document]
wbrgss: (never mind, in 551cd3a)

src/dash-table/syntax-tree/SingleColumnSyntaxTree.ts
Marc-Andre-Rivet: I'm of the opinion that's it's very readable. @alexcjohnson did agree when we last discussed it. It is somewhat controversial 😆

src/dash-table/syntax-tree/SingleColumnSyntaxTree.ts
```diff
@@ -1,9 +1,15 @@
+import * as R from 'ramda';
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: That's a fair question. For this project both are equivalent and the tree-shaking will happen because of https://github.com/plotly/dash-table/blob/dev/tsconfig.base.json#L4 - otherwise this would actually be invalid usage in more recent version of ramda.

src/dash-table/components/FilterFactory.tsx
Marc-Andre-Rivet: Only on `bind`. The context is part of the bind signature so yes. R.partial suffers from the same limitations, see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ramda/index.d.ts#L1369-L1378. The problem isn't bind or R.partial per say but the type definitions.

src/dash-table/syntax-tree/SingleColumnSyntaxTree.ts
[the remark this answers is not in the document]
Marc-Andre-Rivet: @wbrgss This is prettier's formatting. I don't believe this is configurable.

src/dash-table/components/FilterFactory.tsx
Marc-Andre-Rivet: For what it's worth here is the typedef for bind

```
    /**
     * For a given function, creates a bound function that has the same body as the original function.
     * The this object of the bound function is associated with the specified object, and has the specified initial parameters.
     * @param thisArg The object to be used as the this object.
     * @param args Arguments to bind to the parameters of the function.
     */
    bind<T>(this: T, thisArg: ThisParameterType<T>): OmitThisParameter<T>;
    bind<T, A0, A extends any[], R>(this: (this: T, arg0: A0, ...args: A) => R, thisArg: T, arg0: A0): (...args: A) => R;
    bind<T, A0, A1, A extends any[], R>(this: (this: T, arg0: A0, arg1: A1, ...args: A) => R, thisArg: T, arg0: A0, arg1: A1): (...args: A) => R;
    bind<T, A0, A1, A2, A extends any[], R>(this: (this: T, arg0: A0, arg1: A1, arg2: A2, ...args: A) => R, thisArg: T, arg0: A0, arg1: A1, arg2: A2): (...args: A) => R;
    bind<T, A0, A1, A2, A3, A extends any[], R>(this: (this: T, arg0: A0, arg1: A1, arg2: A2, arg3: A3, ...args: A) => R, thisArg: T, arg0: A0, arg1: A1, arg2: A2, arg3: A3): (...args: A) => R;
    bind<T, AX, R>(this: (this: T, ...args: AX[]) => R, thisArg: T, ...args: AX[]): (...args: AX[]) => R;

```

src/dash-table/components/Filter/FilterOptions.tsx
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/893/commits/d76ae04ef2badd066ef37437b8a6513f69927642

wbrgss (commented):
@Marc-Andre-Rivet Let's say I set a column to be case-insensitive using the toggle, or with column-level `'filter_options': 'insensitive'`. I know I can force case-insensitive comparison using e.g. `ilt`, but I might also expect `lt` to behave the same as `ilt`, considering the case-insensitive toggle is on, and right now it doesn't on my end. Does that make sense?

filtering

Marc-Andre-Rivet:
@wbrgss Good point. Yes it makes total sense to have `lt` like `ilt` if the toggle is turned on. I'll see about adding that in.

wbrgss (approved):
Very nice @Marc-Andre-Rivet

:woman_dancing:

src/dash-table/dash/DataTable.js
```diff
@@ -1142,6 +1155,17 @@ export const propTypes = {
...
+    /**
+     * There are two `filter_options` props in the table.
+     * This is the table-level filter_options prop and there is
+     * also the column-level `filter_options` prop.
+     * These props determine whether the applicable filter relational
+     * operators will default to `sensitive` or `insensitive` comparison.
+     * If the column-level `filter_options` prop is set it overrides
+     * the table-level `filter_options` prop for that column.
+     */
+    filter_options: PropTypes.oneOf(['sensitive', 'insensitive']),
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: https://github.com/plotly/dash-table/pull/893/commits/ada845620586f3eecf77d66a6be9b5b04a543676

alexcjohnson (approved):
💃

AnnMarieW:
This is so awesome! :confetti_ball:

PR #906, opened 2021-05-31
Attempt to fix incorrect tooltip placement

### LINKED ISSUE #872: Tooltip position is incorrect after sorting/filtering data table
Tooltips get messed up after sorting or filtering. Simple reproducible example below.

```
import dash
import dash_table
import pandas as pd

df = pd.read_csv('https://raw.githubusercontent.com/curran/data/gh-pages/Rdatasets/csv/MASS/Animals.csv')
df.columns = ["Animal", "body", "brain"]
def create_tooltip(cell):
	num = len(cell)
	if num > 7:
		return(cell)
	else:
		return("")
tooltip_data = [{
            column: 
            {'value': create_tooltip(df.loc[i, column]), 'type': 'markdown'} for column in df.columns if column == "Animal"} for i in range(len(df)-1)]

app = dash.Dash(__name__)

    id='table',
    data=df.to_dict('records'),
    sort_action = "native",
    filter_action  = 'native',
    style_header={'fontWeight': 'bold','border': 'thin lightgrey solid','backgroundColor': 'rgb(100, 100, 100)','color': 'white'},
    style_cell={'fontFamily': 'Open Sans','textAlign': 'left','maxWidth': '0px','whiteSpace': 'no-wrap','overflow': 'hidden','textOverflow': 'ellipsis'},
    style_table = {'overflowX':'scroll', 'overflowY':'scroll', 'maxHeight':'400px', 'maxWidth':'200px', 'marginLeft':'100px'},
    tooltip_data=tooltip_data,
    tooltip_delay=0,
    tooltip_duration=None
)


if __name__ == '__main__':
    app.run_server(debug=True)
```

adament:
This is an attempt to fix #872, the interpretation of the row value of currentTooltip was inconsistent across the codebase. In this patch I try to fix the row value to mean the row in the current displayed data which means that getSelectedTooltip needs to convert to the row number of the full dataset. It works for filtering and pagination in the demo application. However I have not tested with a proper virtualized dataset.

I have little to no experience with javascript, typescript or developing dash components so there is a great risk that I have overlooked something trivial.

Thank you very much for an amazing tool in dash table.

### DIFF
--- a/src/dash-table/derived/table/tooltip.ts
+++ b/src/dash-table/derived/table/tooltip.ts
@@ -42,7 +42,7 @@ function getSelectedTooltip(
               return (
                   !tt.if ||
                   (ifColumnId(tt.if, id) &&
-                      ifRowIndex(tt.if, row) &&
+                      ifRowIndex(tt.if, virtualized.indices[row - virtualized.offset.rows]) &&
                       ifFilter(
                           tt.if,
                           virtualized.data[row - virtualized.offset.rows]
--- a/src/dash-table/handlers/cellEvents.ts
+++ b/src/dash-table/handlers/cellEvents.ts
@@ -165,13 +165,13 @@ export const handleEnter = (
     idx: number,
     i: number
 ) => {
-    const {setState, virtualized, visibleColumns} = propsFn();
+    const {setState, visibleColumns} = propsFn();
 
     setState({
         currentTooltip: {
             header: false,
             id: visibleColumns[i].id,
-            row: virtualized.indices[idx - virtualized.offset.rows]
+            row: idx
         }
     });
 };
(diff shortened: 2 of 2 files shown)


adament:
I am not really sure what is wrong with the lint or how to fix it, so any help would be greatly appreciated.

alexcjohnson:
Thanks @adament - we'll need to dig into the virtualized case a bit to see it this still works. In order to properly lock down the fix as well as these other cases we'll want a couple of tests - probably [test_tooltip.py](https://github.com/plotly/dash-table/blob/dev/tests/selenium/test_tooltip.py) is the right place as well as a good starting point to create the tests.

Re: linting - `npm run format` should fix it for you.

tlinnet:
Any update on this fix?

akullenberg-cbp:
Can someone pick this up again?

kenshima4:
Hi can I have a crack at this?

Issue #834, opened 2020-09-29
Pagination doesn't work when more than one table is on the page


chriddyp:
```python
import dash
import dash_table
import dash_html_components as html
import pandas as pd

df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/solar.csv')

app = dash.Dash(__name__)

table1 = dash_table.DataTable(
    id='table1',
    columns=[{"name": i, "id": i} for i in df.columns],
    data=df.to_dict('records'),
    page_action="native",
    page_size=5,
)

table2 = dash_table.DataTable(
    id='table2',
    columns=[{"name": i, "id": i} for i in df.columns],
    data=df.to_dict('records'),
    page_action="native",
    page_size=7,
)

app.layout = html.Div([
    table1,
    table2
])

if __name__ == '__main__':
    app.run_server(debug=True)
```

As reported in https://community.plotly.com/t/potential-bug-dash-datatable-paging-for-multiple-datatables-on-the-same-page/45431

abondrn:
I am encountering this as well. As in the original post, the first table has page numbers but is unresponsive when you click or try to type in a page number, however this works fine on the second table.

Interestingly, other native interactivity such as sorting and filtering work just fine, so my guess is the culprit lies in how the pager is linked to the table, rather than the other way around. I'd love to take a crack at this, but I'm not at all familiar with the codebase, would greatly appreciate pointers.

mrspieb:
I just had the same issue. It seems to be an error with the spacing underneath the table. The following workaround works for me:

```
import dash
import dash_table
import dash_html_components as html
import pandas as pd

df = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/solar.csv")

app = dash.Dash(__name__)

table1 = dash_table.DataTable(
    id="table1",
    columns=[{"name": i, "id": i} for i in df.columns],
    data=df.to_dict("records"),
    page_action="native",
    page_size=5,
)

table2 = dash_table.DataTable(
    id="table2",
    columns=[{"name": i, "id": i} for i in df.columns],
    data=df.to_dict("records"),
    page_action="native",
    page_size=7,
)

app.layout = html.Div(
    [
        html.Div(table1, style={"padding-bottom": "50px"}),
        table2
    ]
)

if __name__ == "__main__":
    app.run_server(debug=True)

```

PR #907, opened 2021-06-01
issue #834 fix when pagination doesn't work when more than one table …

### LINKED ISSUE #635: Table does not default to page 1 after filtering
After filtering a column in the table, page number is not reset to 1. This causes no problem when filtering is applied whilst page 1 is being inspected, but when looking at a higher page, if the filter applied reduces the size of the table to below the current page number, the forward and previous buttons are removed and navigation becomes impossible.

The buttons return upon removal of the filter, but it can be bothersome to have to return to page 1 before applying any filters.

<pre>
import numpy as np
import pandas as pd
import dash
import dash_core_components as dcc
import dash_table

data = {'numbers': np.arange(0,100)}
df = pd.DataFrame(data, columns=['numbers'])

app = dash.Dash(__name__)

app.layout = html.Div(
			style = {'width':'48%'},
			className = 'table-container',
			children=dash_table.DataTable(
					columns = [{'name' : i, 'id' : i} for i in df.columns],
					data=df.to_dict('records'),
					editable=False,
					filter_action="native",
					filter_query='',
					sort_action="native",
					row_deletable=False,
					selected_columns=[],
					selected_rows=[],
					page_action="native",
					page_current= 0,
					page_size= 12,
				),
			)

if __name__ == '__main__':
	app.run_server(debug=False)
</pre>

### LINKED ISSUE #698: go back to page 1 when data is updated and you are on a page that no longer exists
see https://community.plot.ly/t/unavoidable-circular-dependency-when-using-backend-paging-and-filtering-with-datatables-feature-request/35161/2

if page_current > page_count , force page_current = page_count by changing page_current

 or perhaps we go back to page 1?

### LINKED ISSUE #727: I have an issue related to pagination in dash
Hi Team,

I created a dashboard  and gave pagination fields but my output is not showing the pagination fields in the o/p. Though I have 150+ rows in the data, it is not showing pagination.

code is also attached.

[Corona_Dashboard.zip](https://github.com/plotly/dash-table/files/4430386/Corona_Dashboard.zip)

pagination

### LINKED ISSUE #739: pagination UI doesn't have 100% width
Content that is displayed beneath the table can get rendered _next to_ the pagination UI instead of _below_ the pagination UI.

I believe that if the pagination UI had a 100% width and float right, this issue would be fixed.

image

### LINKED ISSUE #782: Native filter does not work with pagination after page 0
Greetings. It seems like there is a bug with native filtering when it is used on the datatable with pagination after page 1. The datatable documentation version works but it could be due to custom filtering?

Please see this gif below:
ezgif-2-2114c5eb4d6a

It is rather small to see but basically when I am in page 2 and type "Cal" or "Texas", the datatable returns empty (even when Texas is in page 2). This has been tested on a larger dataset and the same issue occurs.

Below is the full sample code that replicate this issue:

```
import dash
import dash_table
import pandas as pd

app = dash.Dash(__name__)

    id='table',
    data=df.to_dict('records'),
	filter_action="native", 
	row_selectable="multi",
	page_action="native", 
	page_current=0,
	page_size=5,
)

if __name__ == '__main__':
    app.run_server(debug=True)
```

### LINKED ISSUE #892: Page selection disappears when number of pages decreases to one
dash-table version: 4.11.3
Browsers tested: Chromium 78.0.3904.70, Edge 83.0.478.58

On tables where the number of entries (and in consqeuence the number of pages) may change dynamically, dash-table will remain on the currently selected page. If I am on page 4/8 and the number of pages reduces to 2, I will then be on page 4/2.

When this happens, I can use the page selectors to navigate back to previous pages. However, when the number of pages decreases to 1, I am stuck on another page without any apparent way to navigate to the actual content.

Here is a simple example illustrating the issue:
pagechange

And here is the code used to produce the example above:
```
import dash
import dash_table
import dash_core_components as dcc
from dash.dependencies import Input, Output

app = dash.Dash(__name__)

app.layout = html.Div([
    dcc.Input(
        id='input', type='number',
    ),
    dash_table.DataTable(
        id='table', page_size= 10,
        columns=[{'name':'i', 'id':'i'}, {'name':'square', 'id':'square'}],
    )
])

@app.callback(Output('table', 'data'),
                    Input('input', 'value'))
def update_table_data(val):
    if not val: val = 9
    return [{'i':i, 'square':i**2} for i in range(val+1)]

if __name__ == '__main__':
    app.run_server(debug=True)
```

This is especially bothersome if the number of table entries cannot be controlled by the user, because the user is then stuck on the non-existing page.

Possible solutions:
- Automatically change pages when the currently selected page exceeds the new total number of pages: User is on page 4/8, number of pages reduces to 3 -> User is on page 3/3 or 1/3.
- Do not hide the page selectors if the user is on another page than page 1, even when there is only 1 page in total. This would allow the user to navigate back to the content even in that case.

I am currently handling this by checking myself and updating ```page_current``` as needed, but a solution on the side of Dash would be preferred, as I don't think the current behaviour is intended.

AnnMarieW:
This PR fixes two pagination issues:
1) Where pagination buttons are not visible or do not work because another table or figure overlaps the pagination container.
    - Closes:  https://github.com/plotly/dash-docs/pull/1150
    - Fixes https://github.com/plotly/dash-table/issues/834
    - Fixes #739 
    - Fixes #727 

2) If you are on a page that no longer exists after data is updated, then no data shows in the table.  This is fixed by resetting the `page_current` to 0
    - Fixes #892 
    - Fixes #698 
    - Fixes #782
    - Fixes #635 

---------------------
-------------------

__Pagination buttons now work in the example from 834:__

pagination

--------
-----------

__Pagination buttons are are now visible here: https://dash.plotly.com/datatable/interactivity__
image

### DIFF
--- a/tests/selenium/test_pagination.py
+++ b/tests/selenium/test_pagination.py
@@ -2,6 +2,7 @@ import dash
 from dash.dependencies import Input, Output
 from dash.exceptions import PreventUpdate
 import dash_html_components as html
+import dash_core_components as dcc
 
 from dash_table import DataTable
 
@@ -221,6 +222,7 @@ def get_app2():
                 data=[{"i": i, "square": i ** 2} for i in range(50 + 1)],
                 page_current=5,
             ),
+            dcc.Graph(),
         ]
     )
 
@@ -243,3 +245,11 @@ def test_tpag011_valid_page(test):
 
     assert target.paging.current.get_value() == "1"
     assert test.get_log_errors() == []
+
+
+def test_tpage012_pagination_row_visible(test):
+    test.start_server((get_app2()))
+
+    test.table("table").is_ready()
+
+    test.percy_snapshot("test_tpage012 Pagination row visible")
--- a/tests/selenium/test_pagination.py
+++ b/tests/selenium/test_pagination.py
@@ -246,10 +246,5 @@ def test_tpag011_valid_page(test):
     assert target.paging.current.get_value() == "1"
     assert test.get_log_errors() == []
 
-
-def test_tpage012_pagination_row_visible(test):
-    test.start_server((get_app2()))
-
     test.table("table").is_ready()
-
-    test.percy_snapshot("test_tpage012 Pagination row visible")
+    test.percy_snapshot("test_tpage011 Pagination row visible")
--- a/tests/selenium/test_pagination.py
+++ b/tests/selenium/test_pagination.py
@@ -247,4 +247,4 @@ def test_tpag011_valid_page(test):
     assert test.get_log_errors() == []
 
     test.table("table").is_ready()
-    test.percy_snapshot("test_tpage011 Pagination row visible")
+    test.percy_snapshot("test_tpag011 Pagination row visible")
--- a/tests/selenium/test_pagination.py
+++ b/tests/selenium/test_pagination.py
@@ -1,6 +1,7 @@
 import dash
 from dash.dependencies import Input, Output
 from dash.exceptions import PreventUpdate
+import dash_html_components as html
 
 from dash_table import DataTable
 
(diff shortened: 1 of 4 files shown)


tests/selenium/test_pagination.py
```diff
@@ -205,3 +207,49 @@ def test_tpag010_limits_page(test):
...
+    assert target.paging.current.get_value() == "1"
+    assert test.get_log_errors() == []
+
+
+def test_tpage012_pagination_row_visible(test):
+    test.start_server((get_app2()))
+
+    test.table("table").is_ready()
+
+    test.percy_snapshot("test_tpage012 Pagination row visible")
```
alexcjohnson, line 255: I'd just add these two lines to the previous test. Yes they're testing different things, but you have exactly this state in `tpag011`, separating these concerns isn't really worth the overhead of tearing down and recreating the app IMO.

AnnMarieW: Right - I thought about doing it way, but didn't know if it was better to have a separate test.  I'll make the change.  Thanks!

tests/selenium/test_pagination.py
```diff
@@ -205,3 +207,44 @@ def test_tpag010_limits_page(test):
...
+    test.start_server(get_app2())
+
+    target = test.table("table")
+    test.find_element("#button").click()
+
+    assert target.paging.current.get_value() == "1"
+    assert test.get_log_errors() == []
+
+    test.table("table").is_ready()
+    test.percy_snapshot("test_tpage011 Pagination row visible")
```
alexcjohnson, line 250:
```suggestion
    test.percy_snapshot("test_tpag011 Pagination row visible")
```
🤓

alexcjohnson (approved):
Fantastic! 💃 

If you merge `dev` back in here the CI tests should run correctly again.

PR #918, opened 2021-07-07
Deps update 20210702

alexcjohnson:


tests/selenium/test_scrolling.py
```diff
@@ -2,11 +2,12 @@
 import dash.testing.wait as wait
 from dash_table import DataTable
 
+import os
```
eff-kay, line 5: its probably better not to import entire modules, they pollute the namespace. Import the relevant stuff. i.e. 

`from os.path import join, dirname`

eff-kay:
LGTM.

tests/selenium/test_scrolling.py
alexcjohnson: Not critical, as you say, but I think I disagree about this actually: `import os` adds one thing to the namespace, and the same every time, whereas `from os.path import …` adds N. Plus in usage `os.path.join` is very clear, whereas `join` by itself could be confused for string join or at least make you look back to the top to see which join it is.

eff-kay: yea makes sense.

Issue #934, opened 2021-08-11
Case-insensitive filter causes "Cannot read property 'toString' of null" exception when the column contains null value


xwk:
If a column in the table contains null value at some rows and you try to do a case insensitive filtering like "ine foo',  a javascript exception "Cannot read property 'toString' of null" exception" will occur. Apparently it is caused by the line "lhs.toString().toUpperCase()" of fnEval() method in relational.ts failed to check whether lhs (i.e. the cell value) is null or not.

A sample app to reproduce the problem.

```python
import dash_html_components as html
import dash_table
import pandas as pd
from collections import OrderedDict

import dash

app = dash.Dash(__name__)

df = pd.DataFrame(OrderedDict([
    ('climate', [None, 'Snowy', 'Sunny', 'Rainy']),
    ('temperature', [13, 43, 50, 30]),
    ('city', ['NYC', None, 'Miami', 'NYC'])
]))


app.layout = html.Div([
    dash_table.DataTable(
        id='table',
        data=df.to_dict('records'),
        columns=[
            {'id': 'climate', 'name': 'climate'},
            {'id': 'temperature', 'name': 'temperature'},
            {'id': 'city', 'name': 'city'},
        ],
        filter_action="native",
    ),
    html.Div(id='table-dropdown-container')
])


if __name__ == '__main__':
    app.run_server(debug=True, port=8051)

```

Run the app, in the 'city' column header of the table, type in 'ine foo' and hit enter, which should reproduce the problem.

Environment:
* Python 3.8
* Dash-table 4.12.1

xwk:
Created a PR https://github.com/plotly/dash-table/pull/935/files to fix this issue

PR #935, opened 2021-08-11
Handle null value safely in case-insensitive filter

xwk:
Fix the issue described in  https://github.com/plotly/dash-table/issues/934

Basically if the value is null, return empty string directly instead of call .toString()

### DIFF
--- a/src/dash-table/syntax-tree/lexeme/relational.ts
+++ b/src/dash-table/syntax-tree/lexeme/relational.ts
@@ -64,8 +64,11 @@ const fnEval = (
     relOp: string
 ): boolean =>
     relOp[0] == 'i'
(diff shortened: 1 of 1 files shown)


zekeshearer:
I have this same issue on dash 2.8.1. @alexcjohnson or other maintainer, could you review this?

alexcjohnson:
Oh thanks @zekeshearer - I had missed this PR as this repo is obsolete. I think perhaps it was created right around when we merged Dash Table into the main Dash repo.

So we can recreate it in the Dash repo. Definitely we should fix this, but I’m not sure if empty string is the right fallback, would we be better off leaving it as its non-string value so it doesn’t match an explicit empty string?

Lastly it deserves a test case.

bparzonz:
Hi, this appears to still be an issue. Is it being addressed elsewhere?

PR #937, opened 2021-08-23
Dash 2 stub package

jonmmease:
Stub package that Dash 2 will depend on to provide a `dash_table` import that is backward compatible with Dash 1. Raises a warning on import.  Stub package has version 5.0.0rc1.

The diff is not very helpful. Probably easiest to check out the branch and look at what's left


dash_table/__init__.py
```diff
@@ -0,0 +1,9 @@
+from dash.dash_table import *  # noqa: F401, F403, E402
+import warnings
+
+warnings.warn(
+    """
+The dash_table package is deprecated. Please replace
+`import dash_table` with `from dash import dash_table`""",
```
alexcjohnson: Should we also mention `FormatTemplate` and `Format`? In https://dash.plotly.com/datatable/data-formatting we have (combining them all):
```py
from dash_table import DataTable, FormatTemplate
from dash_table.Format import Format, Align, Group, Padding, Prefix, Scheme, Sign, Symbol, Trim
```
I guess ATM these won't automatically come along inside `from dash import dash_table`, they'll need to be explicitly imported as:
```py
from dash.dash_table import DataTable, FormatTemplate
from dash.dash_table.Format import Format, Align, Group, Padding, Prefix, Scheme, Sign, Symbol, Trim
```
Is that what we want? Or do we want to promote them to:
```py
from dash.dash_table import Format, Align, Group, Padding, Prefix, Scheme, Sign, Symbol, Trim
```
and include them all in `dash_table/__init__.py` so that you can also access them as:
```py
from dash import dash_table
dash_table.FormatTemplate
dash_table.Align
```
etc?

jonmmease: Updated in faac7b9

alexcjohnson (approved):
💃

PR #939, opened 2021-09-03
stub Format and FormatTemplate

alexcjohnson:
in case someone does eg:
`from dash_table.Format import Format, Sign`
`from dash_table.FormatTemplate import money`


jonmmease (approved):

jonmmease:
:woman_dancing:

